PR #1, opened 2014-07-02
Order + order type initial definition

jsacksick:
Stable enough to be merged.


modules/order/src/Entity/CommerceOrder.php
```diff
@@ -0,0 +1,292 @@
...
+      ->setRevisionable(TRUE)
+      ->setTranslatable(TRUE);
+
+    $fields['changed'] = FieldDefinition::create('changed')
+      ->setLabel(t('Changed'))
+      ->setDescription(t('The time that the order was last edited.'))
+      ->setRevisionable(TRUE)
+      ->setTranslatable(TRUE);
+
+    $fields['line_items'] = FieldDefinition::create('map')
```
joshmiller83, line 400: Shouldn't we just leave off the line_items definition for now? Just a thought ...

jsacksick, line 400: I only added a serialized field for now, nothing more, if it's not ok, I'll remove it :).

modules/order/src/CommerceOrderInterface.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Drupal\commerce_order;
+
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Provides an interface defining a Commerce Order entity.
+ */
+interface CommerceOrderInterface extends EntityInterface {
+
```
bojanz, line 197: Can we fill-in this interface before merging?

modules/order/src/Entity/CommerceOrder.php
```diff
@@ -0,0 +1,306 @@
...
+ *   controllers = {
+ *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
+ *     "list_builder" = "Drupal\commerce_order\Controller\CommerceOrderListBuilder",
+ *     "form" = {
+ *       "add" = "Drupal\commerce_order\Form\CommerceOrderForm",
+ *       "edit" = "Drupal\commerce_order\Form\CommerceOrderForm",
+ *       "delete" = "Drupal\commerce_order\Form\CommerceOrderDeleteForm"
+ *     }
+ *   },
+ *   base_table = "commerce_order",
```
bojanz, line 33: Since we're implementing the revision methods, we should go ahead and declare a revision_table as well.

modules/order/src/CommerceOrderInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_order\CommerceOrderInterface.
+ */
+
+namespace Drupal\commerce_order;
```
ameenross, line 8: Hmm, what is the reason that this namespace is `snake_case`?

bojanz, line 8: That's the Drupal convention, commerce_order is the modue name.

modules/order/commerce_order.module
```diff
@@ -0,0 +1,14 @@
+<?php
+
```
rszrama, line 2: Missing @file doc block.

modules/order/src/Controller/CommerceOrderListBuilder.php
```diff
@@ -0,0 +1,100 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildHeader() {
+    $header = array(
+      'order_id' => array(
+        'data' => $this->t('Order id'),
```
rszrama: Make this "Order ID" anywhere it's used. We've typically allowed ID to ignore sentence capitalization.

modules/order/src/Entity/CommerceOrder.php
```diff
@@ -0,0 +1,306 @@
...
+ *   fieldable = TRUE,
+ *   entity_keys = {
+ *     "id" = "order_id",
+ *     "uuid" = "uuid",
+ *     "revision" = "revision_id",
+ *     "bundle" = "type"
+ *   },
+ *   links = {
+ *     "admin-form" = "commerce_order.type_edit",
+ *     "canonical" = "commerce_order.view",
```
rszrama: Do we know what Drupal 8 wants to do for entities that have two "view" URLs? We have the customer order view and the admin order view, and I'm not sure which one would be considered canonical. : ?

modules/order/src/Form/CommerceOrderDeleteForm.php
```diff
@@ -0,0 +1,73 @@
...
+    return new static(
+      $container->get('entity.manager')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getQuestion() {
+    return t('Are you sure you want to delete the order %order_id?', array('%order' => $this->entity->id()));
```
rszrama: Interface text should use the order number, not the raw ID... which I believe you're adding in now while I leave this comment. : )

modules/order/src/Entity/CommerceOrder.php
```diff
@@ -0,0 +1,306 @@
...
+ *     "form" = {
+ *       "add" = "Drupal\commerce_order\Form\CommerceOrderForm",
+ *       "edit" = "Drupal\commerce_order\Form\CommerceOrderForm",
+ *       "delete" = "Drupal\commerce_order\Form\CommerceOrderDeleteForm"
+ *     }
+ *   },
+ *   base_table = "commerce_order",
+ *   admin_permission = "administer commerce_order entities",
+ *   fieldable = TRUE,
+ *   entity_keys = {
```
rszrama, line 36: When we have the order number in here, we can add a label entity key for it.

modules/order/src/CommerceOrderInterface.php
ameenross, line 8: Is it really Drupal convention that the namespace corresponds to the module name or is it due to PSR autoloading? AFAICT it's the latter. If so, then my next question is: what is the reason for the module name being `snake_case`?

Sweetchuck:
I just asking: Why the namespace is Drupal\commerce_order? 
Why not Drupal\commerce\order?

(Good to see the 8.x development has started :-)

modules/order/src/Entity/CommerceOrder.php
jsacksick, line 33: as discussed, not necessary.

modules/order/src/Entity/CommerceOrder.php
jsacksick, line 36: Done.

modules/order/src/CommerceOrderInterface.php
ameenross, line 8: Answering my own question, this is carried over from the ol' hook system. It's not very clear in the docs though, I'll edit that.

PR #2, opened 2014-07-02
Define the commerce payment transaction entity type.

jkuma:
Here is a first draft of the commerce_payment_transaction entity type shipped with the commerce_payment module. No forms are defined so far, only database model has been implemented.

Please feel free to blame me :)

### DIFF
--- a/modules/payment/src/CommercePaymentTransactionInterface.php
+++ b/modules/payment/src/CommercePaymentTransactionInterface.php
@@ -7,66 +7,90 @@
 
 namespace Drupal\commerce_payment;
 
-use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 
 /**
  * Provides an interface defining a Commerce Transaction entity.
  */
-interface CommercePaymentTransactionInterface extends EntityInterface {
+interface CommercePaymentTransactionInterface extends ContentEntityInterface {
   /**
-   * Returns the identifier.
+   * Sets the instance identifier for a transaction.
    *
-   * @return int
-   *   The entity identifier.
-   */
-  public function id();
-
-  /**
-   * Returns the revision identifier
-   *
-   * @return int
-   *   The entity revision identifier.
-   */
-  public function revisionId();
-
-  /**
-   * Returns the entity UUID (Universally Unique Identifier).
-   *
-   * The UUID is guaranteed to be unique and can be used to identify an entity
-   * across multiple systems.
-   *
-   * @return string
-   *   The UUID of the entity.
-   */
-  public function uuid();
-
-  /**
-   * Return the label of the transaction.
-   *
-   * @return string
-   *   The content of the field.
-   */
-  public function getLabel();
-
-  /**
-   * Sets the label of the transaction.
-   *
-   * @param string $label
-   *   The new name of the transaction.
+   * @param string $instance_id
+   *   The instance identifier of the transaction.
    *
    * @return \Drupal\commerce_payment\CommercePaymentTransactionInterface
    *   The class instance that this method is called on.
    */
-  public function setLabel($label);
+  public function setInstanceId($instance_id);
+
 
   /**
-   * Return the currency code for a transaction.
+   * Returns the instance identifier for a transaction.
    *
    * @return string
-   *   The content of the field.
+   *   The instance id.
    */
-  public function getCurrencyCode();
+  public function getInstanceId();
+
+  /**
+   * Sets the remote identifier for a transaction.
+   *
+   * @param string $remote_id
+   *   The remote transaction identifier.
+   *
+   * @return \Drupal\commerce_payment\CommercePaymentTransactionInterface
+   *   The class instance that this method is called on.
+   */
+  public function setRemoteId($remote_id);
+
+  /**
+   * Returns the remote identifier for a transaction.
+   *
+   * @return string
+   *   The remote id.
+   */
+  public function getRemoteId();
+
+  /**
+   * Sets the human-readable message associated to this transaction.
+   *
+   * @param string|array $message
+   *   A human-readable message that is later serialized into the message
+   *   column.
+   *
+   * @return \Drupal\commerce_payment\CommercePaymentTransactionInterface
+   *   The class instance that this method is called on.
+   */
+  public function setMessage($message);
+
+  /**
+   * Returns human-readable message associated to this transaction.
+   *
+   * @return string|array
+   *   A human-readable message.
+   */
+  public function getMessage();
+
+  /**
+   * Sets the amount of this transaction.
+   *
+   * @param int $amount
+   *   An integer amount.
+   *
+   * @return \Drupal\commerce_payment\CommercePaymentTransactionInterface
+   *   The class instance that this method is called on.
+   */
+  public function setAmount($amount);
+
+  /**
+   * Returns the amount of this transaction.
+   *
+   * @return int
+   *   The amount of this transaction.
+   */
+  public function getAmount();
 
   /**
    * Sets the currency code for a transaction.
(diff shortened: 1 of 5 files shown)


modules/payment/src/CommercePaymentTransactionInterface.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Drupal\commerce_payment;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+
+/**
+ * Provides an interface defining a Commerce Transaction entity.
+ */
+interface CommercePaymentTransactionInterface extends EntityInterface {
```
bojanz: Shouldn't this extend ContentEntityInterface?

modules/payment/src/CommercePaymentTransactionInterface.php
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Returns the identifier.
+   *
+   * @return int
+   *   The entity identifier.
+   */
+  public function id();
```
bojanz: id(), revision() and friends exist on the parent interface(s). We only need to include what's specific to this entity.

modules/payment/src/Entity/CommercePaymentTransaction.php
```diff
@@ -0,0 +1,270 @@
...
+use Drupal\Core\Entity\ContentEntityBase;
+use Drupal\commerce_payment\CommercePaymentTransactionInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+
+/**
+ * Defines the Commerce Transaction entity.
+ *
+ * @ContentEntityType(
+ *   id = "commerce_payment_transaction",
+ *   label = @Translation("Transaction"),
```
bojanz: "Payment Transaction" instead of "Transaction" as the label?

modules/payment/src/Entity/CommercePaymentTransaction.php
```diff
@@ -0,0 +1,270 @@
...
+ *   controllers = {
+ *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
+ *     "list_builder" = "Drupal\commerce\CommercePaymentTransactionListBuilder",
+ *     "form" = {
+ *       "add" = "Drupal\commerce\CommercePaymentTransactionForm",
+ *       "edit" = "Drupal\commerce\CommercePaymentTransactionForm",
+ *       "delete" = "Drupal\commerce\Form\CommercePaymentTransactionDeleteForm"
+ *     },
+ *     "translation" = "Drupal\content_translation\ContentTranslationController"
```
bojanz: The class is now called ContentTranslationHandler

modules/payment/src/Entity/CommercePaymentTransaction.php
```diff
@@ -0,0 +1,270 @@
...
+ *   },
+ *   base_table = "commerce_payment_transaction",
+ *   revision_table = "commerce_payment_transaction_revision",
```
bojanz: This entity type is not revisionable.

modules/payment/src/Entity/CommercePaymentTransaction.php
rszrama: Agreed.

modules/payment/src/CommercePaymentTransactionInterface.php
rszrama: Yep. You can reference the Store interfaces / classes if need be.

modules/payment/src/Entity/CommercePaymentTransaction.php
```diff
@@ -0,0 +1,270 @@
...
+ *     "edit-form" = "commerce_payment_transaction.edit",
+ *     "delete-form" = "commerce_payment_transaction.delete"
+ *   },
+ * )
+ */
+class CommercePaymentTransaction extends ContentEntityBase implements CommercePaymentTransactionInterface {
+  /**
+   * {@inheritdoc}
+   */
+  public function id() {
```
bojanz: Already exists on ContentEntityBase, no need to redeclare it. 
(Other core content entities don't seem to do so)

modules/payment/src/Entity/CommercePaymentTransaction.php
```diff
@@ -0,0 +1,270 @@
...
+    return $this->get('transaction_id')->value;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function revisionId() {
```
bojanz: This entity type is not revisionable, and ContentEntityBase has a getRevisionId() that's supposed to be used for fetching the revision id.

bojanz:
The entity in this PR has a label, but the D7 entity doesn't have a label. I'm guessing we don't want it?

modules/payment/src/CommercePaymentTransactionInterface.php
jkuma: done.

modules/payment/src/CommercePaymentTransactionInterface.php
jkuma: Ok, got it.

modules/payment/src/Entity/CommercePaymentTransaction.php
jkuma: done.

modules/payment/src/Entity/CommercePaymentTransaction.php
jkuma: done.

modules/payment/src/Entity/CommercePaymentTransaction.php
jkuma: yeah, Like you mentioned in your replies above, I'll only declare the methods that are not already implemented in parent classes. Sorry for the mistake.

jkuma:
Dear DC maintainers,

So I've dropped the revision table and I fixed up the errors you have pointed me. The commerce_payment_transaction interface is now complete. And I agree with you guys, the label property doesn't make sense..... So I dropped it

modules/payment/src/CommercePaymentTransactionInterface.php
```diff
@@ -0,0 +1,228 @@
...
+   *
+   * @return \Drupal\commerce_payment\CommercePaymentTransactionInterface
+   *   The class instance that this method is called on.
+   */
+  public function setInstanceId($instance_id);
+
+
+  /**
+   * Returns the instance identifier for a transaction.
+   *
```
rszrama, line 31: I don't believe we need an instance ID any more, or at least we won't call it the same. This property in Commerce 1.x was just a hash to locate the rule that enabled a payment method so the payment method configuration could be extracted from its action. In Commerce 2.x, we'll be storing payment method configurations via some entity type (it isn't clear to me if it will be a content entity type or configuration entity type). We can change this from instance_id to payment_method_id at the very least; but we can always make this change after we've defined the payment method entity type. : )

modules/payment/src/CommercePaymentTransactionInterface.php
```diff
@@ -0,0 +1,228 @@
...
+  /**
+   * Sets the amount of this transaction.
+   *
+   * @param int $amount
+   *   An integer amount.
+   *
+   * @return \Drupal\commerce_payment\CommercePaymentTransactionInterface
+   *   The class instance that this method is called on.
+   */
+  public function setAmount($amount);
```
rszrama: I'm pretty sure we should just make payment transactions fieldable and use a price field this time around. I'm not sure it was a very good idea to include the amount / currency_code as payment transaction properties in 1.x.

jkuma: Yeah, it definitely makes sense to add price field instead of static fields such as "currency_code" and "amount". So I'm gonna drop theses properties and I'll add a TODO for the price field creation.

modules/payment/src/CommercePaymentTransactionInterface.php
jkuma, line 31: I don't think it's a blocker point so far, like you said, we can re-adjust the base fields of this entity type afterwards.

jkuma:
So, Let me summarize things which are included in the commerce_payment module so far:
• The definition of CommercePaymentTransaction entity type. (storing data for payment transactions)
• The definition of CommercePaymentInformation entity type. (storing data around credit cards and payment token, but mostly for credit cards ;) )

bojanz:
Merged the module without the payment information interface and entity class, since we agreed to remake them dramatically.

PR #4, opened 2014-07-03
Product entities

dpolant:
Product/product type entity UI and field definitions

### DIFF
--- a/modules/product/commerce_product.local_actions.yml
+++ b/modules/product/commerce_product.local_actions.yml
@@ -3,3 +3,9 @@ commerce_product.add:
   title: 'Add a new product'
   appears_on:
     - commerce_product.list
+
+commerce_product.product_type_add:
+  route_name: commerce_product.product_type_add
+  title: 'Add a new product type'
+  appears_on:
+    - commerce_product.product_type_list
--- a/modules/product/commerce_product.local_actions.yml
+++ b/modules/product/commerce_product.local_actions.yml
@@ -1,0 +1,5 @@
+commerce_product.add:
+  route_name: commerce_product.add
+  title: 'Add a new product'
+  appears_on:
+    - commerce_product.list
--- a/modules/product/commerce_product.local_actions.yml
+++ b/modules/product/commerce_product.local_actions.yml
@@ -9,3 +9,4 @@ commerce_product.product_type_add:
   title: 'Add a new product type'
   appears_on:
     - commerce_product.product_type_list
+    
--- a/modules/product/commerce_product.local_actions.yml
+++ b/modules/product/commerce_product.local_actions.yml
@@ -1,5 +1,5 @@
 commerce_product.add:
-  route_name: commerce_product.add
+  route_name: commerce_product.product_add_page
   title: 'Add a new product'
   appears_on:
     - commerce_product.list
--- a/modules/product/commerce_product.module
+++ b/modules/product/commerce_product.module
@@ -1,0 +1,8 @@
+<?php
+
+/**
+ * @file
+ * Defines the core Commerce product entity, including the entity itself, the
+ * bundle definitions (product types), and various API functions to manage
+ * products and interact with them through forms and autocompletes.
+ */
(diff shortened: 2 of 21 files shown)


modules/product/commerce_product.local_actions.yml
```diff
@@ -0,0 +1,11 @@
...
+  route_name: commerce_product.product_add_page
+  title: 'Add a new product'
+  appears_on:
+    - commerce_product.list
+
+commerce_product.product_type_add:
+  route_name: commerce_product.product_type_add
+  title: 'Add a new product type'
+  appears_on:
+    - commerce_product.product_type_list
```
bojanz: Missing new line.

modules/product/commerce_product.module
```diff
@@ -0,0 +1,35 @@
+<?php
+
+use Drupal\commerce_product\Entity\CommerceProductType;
+
+/**
+ * @file
+ * Defines the core Commerce product entity, including the entity itself, the
+ * bundle definitions (product types), and various API functions to manage
+ * products and interact with them through forms and autocompletes.
+ */
```
bojanz, line 10: Missing empty line between two docblocks.

modules/product/src/CommerceProductInterface.php
```diff
@@ -0,0 +1,46 @@
...
+ * Provides an interface defining a Commerce Product entity.
+ */
+interface CommerceProductInterface extends EntityInterface {
+  /**
+   * Returns the identifier.
+   *
+   * @return int
+   *   The entity identifier.
+   */
+  public function id();
```
bojanz: The interface should not contain methods that already exist on the parent interface.
On the other hand, it must contain methods specific to this entity class (like the sku getter and setter, etc)

modules/product/src/Entity/CommerceProduct.php
```diff
@@ -0,0 +1,261 @@
...
+ *   label = @Translation("Product"),
+ *   controllers = {
+ *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
+ *     "list_builder" = "Drupal\commerce_product\CommerceProductListBuilder",
+ *     "form" = {
+ *       "add" = "Drupal\commerce_product\Form\CommerceProductForm",
+ *       "edit" = "Drupal\commerce_product\Form\CommerceProductForm",
+ *       "delete" = "Drupal\commerce_product\Form\CommerceProductDeleteForm"
+ *     },
+ *     "translation" = "Drupal\content_translation\ContentTranslationController"
```
bojanz: according to latest core code, this is supposed to be
"translation" = "Drupal\content_translation\ContentTranslationHandler"

modules/product/src/Entity/CommerceProduct.php
```diff
@@ -0,0 +1,261 @@
...
+  public function setSku($sku) {
+    $this->set('sku', $sku);
+    return $this;
+  }
+  
+  /**
+   * {@inheritdoc}
+   */
+  public function getRevisionId() {
+    return $this->get('revision_id')->value;
```
bojanz: Remove id() and getRevisionId(), they already exist on the parent and work the same.
And there is no point in having setRevisionId(), that id is not changeable.

modules/product/src/Form/CommerceProductForm.php
```diff
@@ -0,0 +1,52 @@
...
+      $form['status']['widget']['#default_value'] = 1;
+    }
+    
+    return $form;
+  }
+
+  /**
+   * Overrides \Drupal\Core\Entity\EntityFormController::submit().
+   */
+
```
bojanz: Extra newline.

modules/product/src/CommerceProductInterface.php
rszrama: Guess I should fix this in Store, too; I didn't double check it, but the tutorial was wrong on this point.

modules/product/src/Entity/CommerceProduct.php
rszrama: D'oh, this is wrong in Store as well. :-/

PR #5, opened 2014-07-03
Issue #2297273 by fgm: Added devel module support for Store/StoreType.

fgm:


bojanz:
Can we do the same for orders while we're here?

src/Controller/CommerceDevelController.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * @file
+ * Contains CommerceDevelController.php
+ */
+
+namespace Drupal\commerce\Controller;
+
+
```
bojanz: Double newline, cool in Python, not cool in Drupal :)

src/Controller/CommerceDevelController.php
```diff
@@ -0,0 +1,23 @@
...
+use Drupal\commerce\CommerceStoreInterface;
+use Drupal\commerce\CommerceStoreTypeInterface;
+use Drupal\devel\Controller\DevelController;
+
+class CommerceDevelController extends DevelController {
+
+  public function storeTypeLoad(CommerceStoreTypeInterface $commerce_store_type) {
```
bojanz, line 26: No docblocks anywhere?

PR #9, opened 2014-07-03
Added composer to the module to download external libraries.

JulienD:
First draft to use composer to download external libraries.

Currently there is not a good/official way to include external lib with composer

Problems found : 
- This download libraries in a vendor dir inside the Commerce module and doesn't use Drupal vendor dir. This could cause duplicates libraries and conflicts for Drupal. We should have a composer manager to handle that.
- The "auto-install" is run using Symfony process but this seems to be the simple solution right now


bojanz:
We need to use the composer_manager module instead.

PR #15, opened 2014-07-04
Added product entity display; fixed some router paths

dpolant:


modules/product/commerce_product.routing.yml
```diff
@@ -21,17 +21,27 @@ commerce_product.add:
...
-    _title: 'Edit a product'
+    _title_callback: '\Drupal\commerce_product\Controller\CommerceProductController::editPageTitle'
   requirements:
     _permission: 'update commerce_product entities'
+  options:
+    _admin_route: TRUE    
 
 commerce_product.delete:
-  path: '/admin/commerce/products/{commerce_product}/delete'
+  path: '/product/{commerce_product}/delete'
```
joshmiller83, line 44: The edit path is "/commerce_product/{id}/edit" ... so maybe delete should be "/commerce_product/..." instead of "/product/.../delete"

PR #18, opened 2014-07-04
Issue #2297989 by fgm: prevent deletion of Commerce Types with existing content

fgm:


src/CommerceConfigEntityBundleBase.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/**
+ * @file
+ * CommerceConfigBundleBase.php
+ *
+ * @author: Frédéric G. MARAND <fgm@osinet.fr>
```
bojanz: I don't think we want to add these fancy headers with the individual authors and copyrights to Commerce, since Drupal never does that, and we didn't do it in Commerce 1.x.

(We might need to do it for the libraries though)

src/CommerceConfigEntityBundleBase.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @license General Public License version 2 or later
+ */
+
+namespace Drupal\commerce;
+
+
+use Drupal\Core\Config\Entity\ConfigEntityBundleBase;
+
+class CommerceConfigEntityBundleBase extends ConfigEntityBundleBase implements CommerceEntityTypeInterface {
```
bojanz, line 15: Docblock?

src/CommerceConfigEntityBundleBase.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @author: Frédéric G. MARAND <fgm@osinet.fr>
+ *
+ * @copyright (c) 2014 Ouest Systèmes Informatiques (OSInet).
+ *
+ * @license General Public License version 2 or later
+ */
+
+namespace Drupal\commerce;
+
```
bojanz, line 9: And there's that pythoning double-newline again :)

src/CommerceConfigEntityBundleBase.php
fgm: Indeed. I hadn't made sure to remove them (they're part of my ide setup). Fixing.

fgm:
Seems like this no longer works anyway although it did last friday : yesterday admin/commerce/config/store/add/online wednesday was failing a class autoload, this morning after core update, the commerce_store_type is no longer upcasted from a string to a CommerceStoreType config entity.

joshuataylor:
I am looking into getting this to work with the latest changes.

bojanz:
Merged a simpler solution in #36 . I still want to look into an access based solution, so I will be coming back to this.

PR #27, opened 2014-08-08
Use newly introduced Path extension for generating urls, date service renamed to date formatter

Jbekker:
- Use newly introduced Path extension for generating urls https://www.drupal.org/node/2073811
- Date service renamed to date formatter (Was already done for the product module)
- Missing semicolon in CommerceOrderTypeForm


bojanz:
Woohoo, thanks!

PR #28, opened 2014-08-10
Created payment information entity

pcambra:
Some items to discuss:
- Not sure if this should go to an independent module instead of commerce payment
- The payment information type bundle needs to have some kind of relationship towars the payment method, but we don't have that part (at least me) clear or developed.
- Some renaming probably needs to happen, commerce_payment_information_type was too long so I renamed to info everywhere.
- Some fields needs to be reviewed, month and year could leverage on Date somehow for example.

I'll ping on IRC for some live discussion :)

### DIFF
--- a/modules/payment/src/CommercePaymentInfoInterface.php
+++ b/modules/payment/src/CommercePaymentInfoInterface.php
@@ -14,6 +14,7 @@ use Drupal\Core\Entity\EntityTypeInterface;
  * Provides an interface defining a Commerce Payment Information entity.
  */
 interface CommercePaymentInfoInterface extends ContentEntityInterface {
+
   /**
    * Sets the method_id of the payment method that stored the card.
    *
@@ -155,14 +156,4 @@ interface CommercePaymentInfoInterface extends ContentEntityInterface {
    */
   public function getData();
 
-  /**
-   * Defines the base fields of the entity type.
-   *
-   * @param EntityTypeInterface $entity_type
-   *   Name of the entity type
-   *
-   * @return \Drupal\Core\Field\FieldDefinitionInterface
-   *   An array of entity field definitions, keyed by field name.
-   */
-  public static function baseFieldDefinitions(EntityTypeInterface $entity_type);
 }
--- a/modules/payment/src/CommercePaymentInfoInterface.php
+++ b/modules/payment/src/CommercePaymentInfoInterface.php
@@ -1,0 +1,168 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_payment\CommercePaymentInfoInterface.
+ */
+
+namespace Drupal\commerce_payment;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+
+/**
+ * Provides an interface defining a Commerce Payment Information entity.
+ */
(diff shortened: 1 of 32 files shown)


modules/payment/src/Entity/CommercePaymentInfo.php
```diff
@@ -0,0 +1,336 @@
...
+        'text_processing' => 0,
+      ))
+      ->setDisplayOptions('form', array(
+        'type' => 'text_textfield',
+        'weight' => 2,
+      ))
+      ->setDisplayConfigurable('form', TRUE);
+
+    /*$fields['card_type'] = FieldDefinition::create('string')
+      ->setLabel(t('Card Type'))
```
bojanz: This commented-out block should be removed.

modules/payment/src/Entity/CommercePaymentInfo.php
```diff
@@ -0,0 +1,336 @@
...
+        'weight' => 3,
+        'settings' => array(
+          'display_label' => TRUE
+        )
+      ))
+      ->setDisplayConfigurable('form', TRUE);
+
+    $fields['created'] = FieldDefinition::create('created')
+      ->setLabel('Created')
+      ->setDescription(t('The Unix timestamp when the card was first stored.'))
```
bojanz: Some properties still reference "cards".

modules/payment/src/Entity/CommercePaymentInfo.php
```diff
@@ -0,0 +1,336 @@
...
+        'weight' => 3,
+        'settings' => array(
+          'display_label' => TRUE
+        )
+      ))
+      ->setDisplayConfigurable('form', TRUE);
+
+    $fields['status'] = FieldDefinition::create('boolean')
+      ->setLabel(t('Status'))
+      ->setDescription(t('Card status: inactive (0), active (1), not deletable (2), declined (3).'))
```
bojanz: Remove "not deletable", it makes no sense in D7 either.

modules/payment/src/CommercePaymentInfoTypeInterface.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Drupal\commerce_payment;
+
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
+
+/**
+ * Provides an interface defining a Commerce Payment Information type entity.
+ */
+interface CommercePaymentInfoTypeInterface extends ConfigEntityInterface {
+
```
bojanz, line 16: This interface should contain getDescription / setDescription methods.

modules/payment/src/CommercePaymentInfoInterface.php
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * Defines the base fields of the entity type.
+   *
+   * @param EntityTypeInterface $entity_type
+   *   Name of the entity type
+   *
+   * @return \Drupal\Core\Field\FieldDefinitionInterface
+   *   An array of entity field definitions, keyed by field name.
+   */
+  public static function baseFieldDefinitions(EntityTypeInterface $entity_type);
```
bojanz: This doesn't belong on our interface. Remove it.

bojanz:
We talked about moving the UI to user/%user/payment-info, but since this is non-trivial, leaving it for a followup. Thank you, Pedro.

PR #33, opened 2014-08-11
Add Commerce Product SKU validation.

joshuataylor:
This will require the patch at https://www.drupal.org/node/2319719 for this to work.

### DIFF
--- a/modules/product/src/Plugin/Validation/Constraint/ProductSkuConstraintValidator.php
+++ b/modules/product/src/Plugin/Validation/Constraint/ProductSkuConstraintValidator.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\comment\Plugin\Validation\Constraint\ProductSkuConstraintValidator.
+ */
+
+namespace Drupal\commerce_product\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Validates the ProductSku constraint.
+ */
+class ProductSkuConstraintValidator extends ConstraintValidator {
+
+  /**
+   * {@inheritdoc}
+   */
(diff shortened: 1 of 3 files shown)


modules/product/src/Plugin/Validation/Constraint/ProductSkuConstraintValidator.php
```diff
@@ -0,0 +1,42 @@
...
+  public function validate($field_item, Constraint $constraint) {
+    $sku = $field_item->value;
+    if (isset($sku) && $sku !== '') {
+      $product_id = $field_item->getParent()->id();
+      $query = \Drupal::entityQuery('commerce_product')
+        ->condition("sku", $sku)
+        ->range(0, 1);
+      if ($product_id) {
+        $query->condition("product_id", $product_id, "!=");
+        $query->condition("sku", $sku, "!=");
```
bojanz: We don't need $query->condition("sku", $sku, "!="), it makes no sense.

modules/product/src/Plugin/Validation/Constraint/ProductSkuConstraintValidator.php
```diff
@@ -0,0 +1,42 @@
...
+   */
+  public function validate($field_item, Constraint $constraint) {
+    $sku = $field_item->value;
+    if (isset($sku) && $sku !== '') {
+      $product_id = $field_item->getParent()->id();
+      $query = \Drupal::entityQuery('commerce_product')
+        ->condition("sku", $sku)
+        ->range(0, 1);
+      if ($product_id) {
+        $query->condition("product_id", $product_id, "!=");
```
bojanz: By casting product_id to an integer (like the user constraint does), we could have avoided the if ($product_id).
See that code:
    $value_taken = (bool) \Drupal::entityQuery('user')
      // The UID could be NULL, so we cast it to 0 in that case.
      ->condition('uid', (int) $uid, '<>')

PR #35, opened 2014-08-13
Issue #2297989 by fgm: prevent deletion of Commerce Types with existing content

joshuataylor:
Brought across and ready to be committed .

### DIFF
--- a/modules/order/commerce_order.routing.yml
+++ b/modules/order/commerce_order.routing.yml
@@ -119,3 +119,4 @@ entity.commerce_order_type.devel:
   requirements:
     _module_dependencies: 'devel'
     _permission: 'access devel information'
+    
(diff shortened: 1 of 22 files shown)


joshuataylor:
Thanks so much for @Jbekker  for solving this!

Jbekker:
The delete links are removed from the Order & Product type list pages when there are entities attached to them. Is this the correct behaviour or should the delete link exist but give a error message when trying to delete them?

When a Node Content Type is being deleted with content still attached the following message is show: "Test" is used by 1 piece of content on your site. You can not remove this content type until you have removed all of the Test content.

joshuataylor:
As mentioned on IRC, showing the Delete button how core does it (https://github.com/drupal/drupal/blob/8.0.x/core/modules/node/src/Form/NodeTypeDeleteConfirm.php) is a better User Experience.

The patch has been updated for this.

PR #36, opened 2014-08-14
Prevent deletion of Commerce Types with existing content

Jbekker:
This is a redo of the other pull requests.
- Small cleanups
- Dont allow removing entity types while they are still referenced


Jbekker:
If this pull request gets merged, the other two can be closed.

joshuataylor:
Should also change to the database query to https://www.drupal.org/node/2319977 .

Jbekker:
Nice find @joshuataylor will work on this.

PR #40, opened 2014-08-18
Updates Commerce entity annotation per #2309187

mglaman:
Entity annotation changed in core to clear up "Fix double-link-entry between Entity and Entity Type classes." PR updates this.


bojanz:
The routes you declare as field_ui_base_route need to exist in the module's routing.yml files. Most of these don't

mglaman:
Fixed PR to use proper routes

PR #42, opened 2014-08-19
As commerce relies on devel, add it to the dependencies.

joshuataylor:


joshuataylor:
Agreed, let's wait until that patch lands.

PR #46, opened 2014-08-21
Enable revision for Commerce product entity

mglaman:
You'll see the Form API checks user access to administer commerce_product entities. This is how Core checks if a user can do a revision or not, following same pattern.

Note: Couldn't sleep so I squeezed this in before vacation, which means I may not be able to reply/fix any comments until Monday.


mglaman:
Updated to fix conflicts.

PR #47, opened 2014-08-21
Initial tests for the commerce_store and commerce_store_type entities

Jbekker:
- Initial tests
- Change route permissions to get these tests green untill we setup the entities permissions correctly

### DIFF
--- a/src/Tests/CommerceStoreCRUDTest.php
+++ b/src/Tests/CommerceStoreCRUDTest.php
@@ -1,0 +1,103 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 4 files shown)


src/Tests/CommerceStoreCRUDTest.php
```diff
@@ -0,0 +1,103 @@
...
+      'default_currency' => 'EUR'
+    );
+    $this->drupalPostForm(NULL, $edit, t('Save'));
+  }
+
+  /**
+   * Tests updating a store through the edit form.
+   */
+  public function testUpdateStore() {
+    //Create a new store
```
bojanz: Can we make sure each comment has a space after the // and ends with a dot, please?

PR #50, opened 2014-08-24
Add permissions for product types

joshuataylor:


modules/product/commerce_product.module
```diff
@@ -69,6 +69,11 @@ function commerce_product_permission() {
       'description' => t('Allows users to delete commerce product entities.'),
       'restrict access' => TRUE,
     ),
+    'administer commerce_product_type entities' => array(
+      'title' => t('Administer commerce_product_type entities'),
+      'description' => t('Allows users to administer commerce product entities.'),
```
bojanz: The description is not correct.

PR #56, opened 2014-08-27
Rename variables to $product_type or $product.

joshuataylor:


Jbekker:
What about our other entities?

joshuataylor:
@Jbekker This is only for variables, not entity names, unless I missed something?

Jbekker:
I meant, there are also $commerce_store_type $commerce_store variables etc. that should be renamed.

bojanz:
I agree with jbekker, let's do that in the next PR.

PR #57, opened 2014-08-27
Issue #2323209 - Created a basic price field

pcambra:
I haven't add a price field instance to the commerce payment transaction because we don't have bundles yet, the price field storage is provided.
See https://www.drupal.org/node/2323209

### DIFF
--- a/modules/price/src/Plugin/Field/FieldType/Price.php
+++ b/modules/price/src/Plugin/Field/FieldType/Price.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_price\Plugin\Field\FieldType\Price.
+ */
+
+namespace Drupal\commerce_price\Plugin\Field\FieldType;
+
+use Drupal\Core\Field\FieldItemBase;
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\TypedData\DataDefinition;
+
+/**
+ * Plugin implementation of the commerce price field type.
+ *
+ * @FieldType(
+ *   id = "price",
+ *   label = @Translation("Price"),
+ *   description = @Translation("Stores a decimal amount and a three letter currency code."),
+ *   default_widget = "price_simple",
+ *   default_formatter = "price",
+ * )
+ */
+class Price extends FieldItemBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function propertyDefinitions(FieldStorageDefinitionInterface $field_definition) {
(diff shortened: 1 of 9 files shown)


modules/price/src/Plugin/Field/FieldType/Price.php
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function schema(FieldStorageDefinitionInterface $field_definition) {
+    return array(
+      'columns' => array(
+        'amount' => array(
+          'description' => 'The price amount.',
+          'type' => 'float',
```
bojanz: We need to store amounts in a decimal(19,6) field, which maps to "type" => "numeric", "precision" => 19, "scale" => 6, in schema api.

modules/price/src/Plugin/Field/FieldWidget/PriceSimpleWidget.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+namespace Drupal\commerce_price\Plugin\Field\FieldWidget;
+
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Field\WidgetBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Plugin implementation of the 'link' widget.
```
bojanz: 'link' widget? :)

modules/price/src/Plugin/Field/FieldWidget/PriceSimpleWidget.php
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @FieldWidget(
+ *   id = "price_simple",
```
bojanz, line 18: Why is the widget called price_simple? we won't have a price_complex widget ;)

modules/price/src/Plugin/Field/FieldWidget/PriceSimpleWidget.php
```diff
@@ -0,0 +1,51 @@
...
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+    $element['amount'] = array(
+      '#type' => 'textfield',
+      '#title' => $this->t('Amount'),
+      '#default_value' => isset($items[$delta]->amount) ? $items[$delta]->amount : NULL,
+      '#required' => $element['#required'],
+      '#size' => 10,
+      '#maxlength' => 255,
+    );
+    $element['currency_code'] = array(
```
bojanz, line 43: Can this be a dropdown listing active currencies instead?

PR #66, opened 2014-09-04
Initial product attributes

joshuataylor:
This adds initial product attributes.

The reason why we don't use third_party_settings for the form instance is because otherwise the form will double save.

If I could get this checked over by @Jbekker before Bojanz that would be great.

### DIFF
--- a/modules/product/commerce_product.module
+++ b/modules/product/commerce_product.module
@@ -7,13 +7,12 @@
  * products and interact with them through forms and autocompletes.
  */
 
-use Drupal\commerce_product\Entity\CommerceProductType;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\commerce_product\CommerceProductInterface;
 use Drupal\Core\Url;
-use Drupal\Core\Language\LanguageInterface;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\Core\Form\FormStateInterface;
 
 /**
  * Implements hook_entity_operation().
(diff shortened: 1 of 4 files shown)


joshuataylor:
This has had a quick look over by @Jbekker , @bojanz can you have a look :)?

modules/product/config/schema/commerce_product.schema.yml
```diff
@@ -14,3 +14,20 @@ commerce_product.commerce_product_type.*:
...
+    attributes:
+      type: sequence
+      label: 'Field properties for which to synchronize product attributes'
+      attribute_field:
+        - type: boolean
+          label: 'Whether or not this field should be used as a product attribute'
+      attribute_widget:
+        - type: string
+          label: 'The widget to use for the product attribute'
+      attribute_widget_title:
```
bojanz, line 31: We should mark this title as translatable, and confirm that it can be translated with config_translation.

modules/product/commerce_product.module
```diff
@@ -122,3 +121,79 @@ function commerce_product_add_body_field($product_type_id, $label = 'Body') {
...
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter() for 'field_ui_instance_edit_form'.
+ */
+function commerce_product_form_field_ui_field_instance_edit_form_alter(array &$form, FormStateInterface $form_state) {
+  $instance = $form_state['instance'];
+  // If the current field instance is attached to a product type,
+  // and of a field type that defines an options list.
+  if ($instance->entity_type == 'commerce_product' && $form['#field']->module == "options") {
```
bojanz: Let's confirm that the module check doesn't leave out taxonomy terms, or entityreference fields. They're both valid candidates.

modules/product/commerce_product.module
```diff
@@ -122,3 +121,79 @@ function commerce_product_add_body_field($product_type_id, $label = 'Body') {
...
+          ':input[name="instance[commerce_product][attribute_field]"]' => array('checked' => TRUE),
+        ),
+      ),
+      '#default_value' => $default_attribute_widget_title
+    );
+    $form['actions']['submit']['#submit'][] = 'commerce_product_form_field_ui_field_instance_edit_form_submit';
+  }
+}
+
+function commerce_product_form_field_ui_field_instance_edit_form_submit($form, FormStateInterface $form_state) {
```
bojanz: Missing a docblock for this function.

modules/product/commerce_product.module
```diff
@@ -122,3 +121,87 @@ function commerce_product_add_body_field($product_type_id, $label = 'Body') {
...
 }
+
+/**
+ * Implements hook_form_FORM_ID_alter() for 'field_ui_instance_edit_form'.
+ */
+function commerce_product_form_field_ui_field_instance_edit_form_alter(array &$form, FormStateInterface $form_state) {
+  $instance = $form_state['instance'];
+  // If the current field instance is attached to a product type,
+  // and of a field type that defines an options list.
+  $allowed_fields = array("options", "taxonomy", "entity_reference");
```
bojanz: These should be single quotes (around options, taxonomy, entity_reference)

bojanz:
This is ready to be merged (once the quotes nitpick is addressed), Josh can we get a quick rebase?

Jbekker:
I fixed the quotes and rebased this.

Jbekker:
Alright, I think it's good to go.

PR #67, opened 2014-09-04
Issue #2327283 Create translations for a currency

Jbekker:
- Refactor commerce_price.install to use the currency importer service.
- Rename the importer service.
- The Symbol field is translatable as well.
- Create translations for currency entities when a new language is added or when the user adds
  a new currency.

### DIFF
--- a/modules/price/commerce_price.module
+++ b/modules/price/commerce_price.module
@@ -1,6 +1,7 @@
 <?php
 
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\language\Entity\ConfigurableLanguage;
 
 /**
  * Implements hook_entity_operation().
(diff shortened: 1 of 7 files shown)


modules/price/commerce_price.module
```diff
@@ -1,7 +1,7 @@
 <?php
 
 use Drupal\Core\Entity\EntityInterface;
-
```
bojanz, line 4: There's supposed to be a blank line after the use declarations, I'm guessing you removed it by mistake.

modules/price/src/CurrencyImporter.php
```diff
@@ -77,8 +71,41 @@ public function importCurrency($currency_code) {
...
     return $entity;
   }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function importCurrencyTranslations($currencies = array(), $languages = array()) {
+    // Skip importing translations if the site it not multilingual.
+    if (!$this->languageManager->isMultilingual()) {
+      RETURN FALSE;
```
bojanz: return shouldn't be uppercase.

PR #69, opened 2014-09-09
Add a owner field to the store.

Jbekker:
- Move some of the form field definitions to the BaseFieldDefinition on the entity.
- Coding style fixes in the store tests


Jbekker:
@bojanz shall I also remove the currency field on the store with this pull? or create a seperate one.

PR #71, opened 2014-09-19
Adding hook_requirements() to commerce_price.install 

dwkitchen:
Adding hook_requirements() to commerce_price.install as uses the commerceguys/intl library which needs to be installed via composer first.

### DIFF
--- a/modules/price/commerce_price.install
+++ b/modules/price/commerce_price.install
@@ -16,3 +16,34 @@ function commerce_price_install() {
     $entity->save();
   }
 }
+
+/**
+ * Implements hook_requirements().
+ *
+ * Check that the Commerce Guys' Libraries have been installed.
+ */
+function commerce_price_requirements($phase) {
+  $requirements = array();
+
+  if ($phase == 'install') {
+    if (\Drupal::moduleHandler()->moduleExists('composer_manager')) {
+      /* @var $packages Drupal\composer_manager\ComposerPackages */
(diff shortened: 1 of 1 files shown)


modules/price/commerce_price.install
```diff
@@ -16,3 +16,34 @@ function commerce_price_install() {
...
+/**
+ * Implements hook_requirements().
+ *
+ * Check that the Commerce Guys' Libraries have been installed.
+ */
+function commerce_price_requirements($phase) {
+  $requirements = array();
+
+  if ($phase == 'install') {
+    if (\Drupal::moduleHandler()->moduleExists('composer_manager')) {
```
bojanz: Since composer_manager is a dependency of commerce.module, won't it always be present at this point?

dwkitchen: Was think that is someone tries to install all the commerce modules without having installed composer_manager, Drupal is going to ask do you want to install composer manager.

Not sure exactly when in that flow that hook_requirements() is called for each of the modules. Is it individually as it tries to install each of them or all of them at the start before trying to install any of them.

If we can guarantee that composer manager will have been installed through the module dependencies by the point we get to the next line then we don't need the if statement.

Jbekker:
I'm pretty sure composer_manager is always present when enabling commerce_price since drupal will first install commerce which depends on composer_manager.

dwkitchen:
Removed condition on composer_manager being installed.

PR #76, opened 2014-09-24
Apply formatters and widgets to Commerce base fields

pcambra:
Here's an initial version, regarding the getCurrentUserId() method, I need to ping Berdir to ask him about what's the future of it as it something that probably will end somewhere else, see https://www.drupal.org/node/1979260#comment-9131623


bojanz:
I just merged a PR that changed a single field weight, which introduced a conflict here... Sorry!

PR #81, opened 2014-09-30
 Issue #2333375 Format Price field using NumberFormatter. 

Jbekker:
Format the price field using the Intl library.

I still have one question.. what to do when the number format for the current locale isn't available? We could fallback to 'en' but then we need to make sure that the english number format is installed when the price module is enabled and make sure it can't be removed.


Jbekker:
It probably makes sense to move the price formatting to a service at some point. If someone would like to create a fancy formatter, they can use the service for formatting their prices.

bojanz:
I agree that the formatter should be in a service. Let's make that a followup.

bojanz:
I like the idea from the drupal.org issue, let's change the currency display mode select to a "Show the currency code instead of the currency symbol" checkbox, it's better UX.

PR #83, opened 2014-10-02
Fix step 4 where composer install doesn't work

joshuataylor:
See issue https://www.drupal.org/node/2348771 for full details about this issue.


bojanz:
As discussed on IRC, I fixed the issue by removing the symfony/yaml dependency from pricing.
We'll want to avoid depending on drush for now because it's still broken for us (the autoloader issue),

PR #85, opened 2014-10-02
Fix travis to use the new autoload workaround in settings.php

joshuataylor:
Also included a couple of tweaks from the redirect module, which might be good to look into it further in future.

Currently tests are failing which I'm looking into now - it's also happening locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 5.4
   - 5.5
+  - 5.6
 
 mysql:
   database: drupal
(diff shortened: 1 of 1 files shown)


joshuataylor:
Actually it seems that travis is still not working for this due to exit codes, please hold 5.

joshuataylor:
Does commerce have PHPUnit tests? These aren't firing.

Jbekker:
No units tests yet but those will come at some point.

joshuataylor:
@Jbekker I've gone ahead and disabled it until it's been added, so once this has been done re-enable it.

PR #94, opened 2014-10-03
fix the ui for commerce product editing

BBGuy:
Me and iAugur have made the following changes:
Fix the display of the price widget - make them inline with added css file
Move the Active form element under the price
Set admin theme for product edit

### DIFF
--- a/modules/price/css/commerce_price.css
+++ b/modules/price/css/commerce_price.css
@@ -1,0 +1,8 @@
+
+/**
+ * @file
+ * Styling for Commerce Price.
+ */
+.field-widget-price-simple .form-type-textfield, .field-widget-price-simple .form-type-select {
+  display: inline;
+}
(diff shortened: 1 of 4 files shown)


modules/price/css/commerce_price.css
```diff
@@ -0,0 +1,16 @@
+/*
+    Document   : commerce_price.css
```
bojanz: Let's remove this header, it has no purpose.

modules/price/css/commerce_price.css
```diff
@@ -0,0 +1,16 @@
...
+    Created on : Oct 3, 2014, 1:12:18 PM
+    Author     : guyshneerson
+    Description:
+        Purpose of the stylesheet follows.
+*/
+
+/*
+   TODO customize this sample style
```
bojanz: Sample style?

PR #99, opened 2014-10-07
Fix error payment installation and change the fields to be plain strings

Jbekker:
Text processing option was removed on the text field type and a new string field type was added: https://www.drupal.org/node/2329465


joshuataylor:
Ran into this same problem with tests (which is a pending PR) for other parts of commerce.

PR #100, opened 2014-10-08
scrutinizer fixes from scrutinizer-ci

joshuataylor:
https://scrutinizer-
ci.com/g/joshuataylor/commerce/inspections/75a2d2a7-4dc2-4f24-a1a7-55e572d13f6b/patches


modules/order/src/Entity/CommerceOrder.php
```diff
@@ -183,6 +183,7 @@ public function getOwner() {
 
   /**
    * {@inheritdoc}
+   * @return integer
```
bojanz, line 186: No.

joshuataylor, line 186: WTF, agreed.

PR #104, opened 2014-10-15
Add operation links to the default Products view for admins.

jazzdrive3:
This follows the conventions set forth in the Node module to Views field handlers for links to a product entity: view, edit, and delete. The operations are gathered in a Dropbutton field on the View itself. The entity_operations field handler no longer exists, and was returning as a broken handler.

This also creates a extension of EntityViewsData, which would be required anyway for more granular control of Views integration.

Once this is verified as the way to do things, I'll go ahead and do this for the other Commerce entities as well.

One issue is that the Dropbutton is rendering properly in default views, but that is a core issue that. Currently, you will see the fallback list of links instead of the nice looking ctools dropbutton.

TEST
- [ ] clear your caches
- [ ] Visit admin/commerce/products and you should see the Edit and Delete links in the Operations column.


bojanz:
https://www.drupal.org/node/2165989 has now landed, giving each entity type an automatic Operations handler. There are related issues about providing the same for the individual operations as well. So we don't need this PR anymore. Thank you for your efforts jazzdrive3.

PR #105, opened 2014-10-15
2352731: Import number formats for all variants of a language

StryKaizer:
2352731: Import number formats for all variants of a language


bojanz:
Thank you for your contribution StryKaizer, as always, it is very appreciated.

I'm going to have to close this PR instead of merging it because we've decided to stop using config entities for number formats. See you in another issue!

PR #109, opened 2014-11-25
Require entity reference for tests, otherwise it will fail.

joshuataylor:


joshuataylor:
Making sure this passes travisCI, if it does it should be good to go.

joshuataylor:
It is green so good to go :).

PR #111, opened 2014-12-07
Price formatters

dwkitchen:
A minor correction to the Price Field Formatter and two additional field formatters.
1. Currency to display the currency information
2. Amount to only display the numeric amount

I would like to add decimal place formatting to the Amount Formatter to use the currency setting or a fixed number of decimal places.

### DIFF
--- a/modules/price/src/Plugin/Field/FieldFormatter/PriceAmountFormatter.php
+++ b/modules/price/src/Plugin/Field/FieldFormatter/PriceAmountFormatter.php
@@ -1,0 +1,118 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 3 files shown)


bojanz:
Needs a rebase.

dwkitchen:
Rebased

bojanz:
Committed the initial price formatters in 331a6985bcf4d937ee15eb007c28942021acfb7d
One uses commerceguys/intl, one uses a theme function.

After discussing the two formatters in this PR with both dwkitchen and rszrama I've decided not to commit them, their primary purpose is exports via views, so we should make sure the amount and currency_code are available via views instead, with the proper handlers.
We'll come back to it if necessary.

PR #114, opened 2014-12-20
Line item Entity

dwkitchen:
My initial development of a line item type

There are three issues that is see and are all related to Drupal core.
1.  Entity reference configuration requires the selection of the target entity and bundles, there is no option for all bundles. The line items field on orders needs to be able to reference all line item types; products, shipping, etc.
2.  The same applies to the the product line item type which needs to be able to refference all product types.
3.  The config is only able to create the line items field on the order bundle, if a user creates a new type of order type from the UI it will not have the line items field on it which will be required. (This also applies to the price field on products)

### DIFF
--- a/modules/line_item/src/Form/CommerceLineItemTypeForm.php
+++ b/modules/line_item/src/Form/CommerceLineItemTypeForm.php
@@ -1,0 +1,102 @@
+<?php
+
+/**
+ * @file
+ * Contains Drupal\commerce_line_item\Form\CommerceLineItemTypeForm.
+ */
+
+namespace Drupal\commerce_line_item\Form;
+
+use Drupal\Core\Entity\EntityForm;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class CommerceLineItemTypeForm extends EntityForm {
+
+  /**
+   * The line_item type storage.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $lineItemTypeStorage;
+
+  /**
+   * Create an CommerceLineItemTypeForm object.
+   *
+   * @param \Drupal\Core\Entity\EntityStorageInterface $line_item_type_storage
+   *   The line_item type storage.
+   */
+  public function __construct(EntityStorageInterface $line_item_type_storage) {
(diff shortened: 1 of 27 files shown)


bojanz:
"Entity reference configuration requires the selection of the target entity and bundles, there is no option for all bundles. " -> A regression compared to D7 entityreference then? Did you open a core issue?

modules/order/src/CommerceOrderInterface.php
```diff
@@ -120,25 +120,6 @@ public function getRevisionAuthor();
   public function setRevisionAuthorId($uid);
 
   /**
-   * Returns the line items associated with this order.
-   *
-   * @return array
-   *   The line items of this order.
-   */
-  public function getLineItems();
```
bojanz, line 128: Why were these two methods removed? They make sense even if entityreference is used under the hood, no?

modules/line_item/src/Form/CommerceLineItemTypeForm.php
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * The line_item type storage.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $lineItemTypeStorage;
+
+  /**
+   * Create an IndexForm object.
```
bojanz: IndexForm is a copy-paste mistake.

modules/line_item/src/Form/CommerceLineItemTypeForm.php
```diff
@@ -0,0 +1,102 @@
...
+      '#type' => 'machine_name',
+      '#default_value' => $line_item_type->id(),
+      '#machine_name' => array(
+        'exists' => array($this->line_itemTypeStorage, 'load'),
+        'source' => array('label'),
+      ),
+      '#disabled' => !$line_item_type->isNew(),
+    );
+
+    $form['description'] = array(
```
bojanz, line 70: Do we want line item types to have a description? We've never shown it anywhere.

dwkitchen, line 70: When an admin is adding a line item to an order in the back office.

dwkitchen:
It is possible to define an entity reference field in baseFieldDefinitions(), this also avoids having to set the bundle type as this is only required in the UI, if it is not set it is still all bundle types.
I have also removed the bundle type setting from the product line item type entity reference field to products.

PR #115, opened 2014-12-20
Moving the price field to a BaseFieldDefinition

dwkitchen:
So it is included on all product types


dwkitchen:
As per our conversation, the default price field now has a configuration setting that can enable/disable it.

dwkitchen:
After some further thought and investigation after you mentioned multi-currency, I have tried creating a field for each enabled currency. The price widget needs some additions like the ability to set the widget to only one currency, you can't set the default currency code without setting the value ATM.
This has given be some thoughts on how we can mange tax rates on products as well.

bojanz:
Postponing this for now. The price field still makes more sense as a configurable field.

PR #121, opened 2014-12-24
Add field storage for commerce_line_items field.

googletorp:
We should have a field for referencing line items on orders.


dwkitchen:
We can now define entity reference fields as base fields see https://github.com/commerceguys/commerce/blob/8.x-2.x/modules/order/src/Entity/CommerceOrder.php#L343

I started with a field like you have but this is much better as it will create the field on new order types that the user creates as well.

bojanz:
Correct, we also discussed this on irc.

PR #110, opened 2014-11-25
Tax

ralt:
- Adds the tax module
- Adds the tax type entity and its Forms friends

What's left:
- Adding the `CommerceTaxRate` entity and its friends
- Adding the `CommerceTaxRateAmount` entity and its friends
- Figure out how to do an entity reference form to select tax rates in tax types

PR #122, opened 2014-12-29
Uses core formatter for the language

ralt:
Integrating https://www.drupal.org/node/2393007

Not sure we should do it for products too...


ralt:
For the record, the failing tests are because #110 relies on Zone being updated.

PR #129, opened 2015-01-15
Update commerce for the Change Record "Entity annotation link templates ...

joshuataylor:
...are paths again"

Also fixes various broken routes, as found throughout.

One thing to note is that CommercePaymentTransaction doesn't have a form either.


joshuataylor:
I removed the extra use as core doesn't use them, but PHPStorm whinges about them..

PR #139, opened 2015-01-28
Revert rename camelCase for controller

Jbekker:
This partially reverts the camelCase rename on the controller function parameters
to fix upcasting.

### DIFF
--- a/modules/line_item/src/Controller/CommerceLineItemDevelController.php
+++ b/modules/line_item/src/Controller/CommerceLineItemDevelController.php
@@ -19,23 +19,23 @@ class CommerceLineItemDevelController extends DevelController {
   /**
    * Dump devel information for a Commerce Line item Type.
    *
-   * @param \Drupal\commerce_line_item\CommerceLineItemTypeInterface $lineItemType
+   * @param \Drupal\commerce_line_item\CommerceLineItemTypeInterface $commerce_line_item_type
    *
    * @return string
    */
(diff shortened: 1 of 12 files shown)


joshuataylor:
Absolutely, plus we need to update tests to figure out why they were passing.

A programmer has a problem. "I know", they think, "I'll use regular expressions!".

Now they have two problems.

Jbekker:
I agree joshua, tests are needed. I'll take a look this weekend if you didnt got to it first.

bojanz:
I've investigated this problem a bit. The upcasting is documented at https://www.drupal.org/node/2310425 and performed by the EntityConverter class.
Now, in order to do the upcasting, the converter needs the entity_type parameter.
By default, this is done by looking at the controller arguments, making sure that they match the entity type name. However, if we specify the entity_type directly in the route, we can avoid this.

See the example from the link:

```
route_with_two_nodes:
  path: '/foo/{node1}/{node2}
  defaults:
    _controller: '\Drupal\example\Controller\ExampleController::foo'
  options:
    parameters:
      node1:
        type: entity:node
      node2:
        type: entity:node
```

We do the same for our own entity types, and name the controller arguments however we want ($currency instead of $commerce_currency, $numberFormat instead of $commerce_numberFormat, etc)

bojanz:
jbekker and I explored our idea, but had problems with Drupal that will need to be fixed first. So I'm going to merge this PR as is for now.

PR #140, opened 2015-01-28
Adds a missing route for the payment info type entity

Jbekker:


bojanz:
Merging this one, but we can skip payments for a while, the existing code is very unfinished.

PR #141, opened 2015-01-28
Issue #2416141: Adds a 'digital' property to the product type

Jbekker:
https://www.drupal.org/node/2416141


dwkitchen:
What is the propose of the digital property? I don't agree with the use of this and we should work in the same way as Commerce 1.x did by have an option of marking a product as physical when enabling shipping.

bojanz:
David and I discussed this and agreed to improve the label & options of this setting. To be done in a followup.

PR #143, opened 2015-02-01
Issue #2418085: Move entity list builders out of the controller folder

Jbekker:
https://www.drupal.org/node/2418085

### DIFF
--- a/modules/price/src/CommerceCurrencyListBuilder.php
+++ b/modules/price/src/CommerceCurrencyListBuilder.php
@@ -2,10 +2,10 @@
 
 /**
  * @file
- * Contains \Drupal\commerce_price\Controller\CommerceNumberFormatListBuilder.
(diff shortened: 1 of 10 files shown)


modules/price/src/CommerceCurrencyListBuilder.php
```diff
@@ -2,10 +2,10 @@
 
 /**
  * @file
- * Contains \Drupal\commerce_price\Entity\Controller\CommerceCurrencyListBuilder.
+ * Contains \Drupal\commerce_price\Entity\CommerceCurrencyListBuilder.
```
bojanz: the "Entity\" part of this namespace is incorrect, let's fix that as well. So it should be: Contains \Drupal\commerce_price\CommerceCurrencyListBuilder.

bojanz:
Apart from that one docblock, the rest looks good.

PR #147, opened 2015-02-19
Handle default currencies in Store entiyu

googletorp:
https://www.drupal.org/node/2423233


bojanz:
We can actually convert the field to an entityreference, since currencies are config entities.

bojanz:
Opened an issue to track this: https://www.drupal.org/node/2464763

Closing this PR because the implementation is wrong.

PR #148, opened 2015-02-19
Add missing route

googletorp:
https://www.drupal.org/node/2429785

### DIFF
--- a/commerce.routing.yml
+++ b/commerce.routing.yml
@@ -44,6 +44,14 @@ entity.commerce_store.collection:
   requirements:
     _permission: 'administer stores'
 
+entity.commerce_store.canonical:
+  path: '/admin/commerce/config/store/{commerce_store}/edit'
+  defaults:
+    _entity_form: commerce_store.edit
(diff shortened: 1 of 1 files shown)


commerce.routing.yml
```diff
@@ -44,6 +44,15 @@ entity.commerce_store.collection:
...
 
+entity.commerce_store.canonical:
+  path: '/admin/commerce/config/store/{commerce_store}/edit'
+  defaults:
+    _entity_form: commerce_store.edit
+    _title: 'Edit a store'
+  requirements:
+    _entity_access: 'commerce_store.edit'
+
+
```
bojanz: Nitpick of the year: let's kill this extra newline :)

Jbekker:
I just tested this without this fix on a new drupal 8 install and it works just fine. Also, it doesn't make any sense to copy the edit route and use it as a canonical route.

bojanz:
@googletorp Is the crash reproducible with the latest Drupal core -dev?

PR #150, opened 2015-02-24
Update run-tests to latest reroll

joshuataylor:
From https://www.drupal.org/node/2189345#comment-9636131


joshuataylor:
Failure is not related to patch, so should be good to go.

PR #154, opened 2015-02-28
Add enum library as composer dependency

pcambra:


bojanz:
Technically we don't need to, because Composer will pick up that enum is required by addressing, but I like to have our libraries listed anyway. Thanks.

PR #155, opened 2015-02-28
Removed destination from admin listings to ensure that the cancel link points to the right listing

pcambra:


bojanz:
This PR is adding back  _admin_route: TRUE to routes, which I removed because core adds that automatically to any route that starts with admin/

pcambra:
That's right! I looked for this but couldn't find it.
We've got an inconsistent route I think, every other entity have the routes for delete and edit in admin but products are '/product/{commerce_product}/delete'. I'll open an additional PR with this.

bojanz:
\Drupal\system\EventSubscriber\AdminRouteSubscriber is the class that adds _admin_route.

The products are following the node example, so if it matches node, we're fine.

bojanz:
The views have been rebuilt, this will require the PR to be recreated :/

PR #156, opened 2015-03-05
Issue #2416141: Remove digital boolean from product type

dwkitchen:
There is no purpose for the field, it is to simplistic for tax, and anything for this requirement should be added by the Tax module.
There is also no other provision for digital services in Commerce core, so these should be provided by the module implementing the digital service.


joshuataylor:
drush was broken for this commit, which has since been fixed.

Please rerun the test @bojanz :)

bojanz:
Committed in 3d508e5944e972dd354e6b5676f19e2363f4fb08.

PR #160, opened 2015-03-11
TaxResolver service

dwkitchen:
Example use:

``` PHP
$resolver = \Drupal::service('commerce_tax.tax_resolver')
$resolver->resolveAmounts(TaxableInterface $taxable, Context $context);
```

### DIFF
--- a/modules/tax/src/TaxResolver.php
+++ b/modules/tax/src/TaxResolver.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_tax\TaxResolverFactory.
+ */
+
+namespace Drupal\commerce_tax;
+
+use CommerceGuys\Tax\Resolver\TaxResolver as BaseTaxResolver;
+use CommerceGuys\Tax\Resolver\Engine\TaxTypeResolverEngine;
+use CommerceGuys\Tax\Resolver\Engine\TaxRateResolverEngine;
+use Drupal\commerce_tax\TaxTypeRepository;
+use Drupal\commerce_tax\Plugin\CommerceTax\TaxTypeResolver;
+
+class TaxResolver extends BaseTaxResolver {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct() {
+    // Tax Type Resolver.
+    $taxTypeRepository = new TaxTypeRepository();
+    $this->taxTypeResolverEngine = new TaxTypeResolverEngine();
+
+    $tax_type_resolver_manager = \Drupal::service('plugin.manager.commerce_tax.tax_type_resolver');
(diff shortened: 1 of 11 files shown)


modules/tax/src/TaxTypeRepository.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Drupal\commerce_tax;
+
+use CommerceGuys\Tax\Repository\TaxTypeRepositoryInterface;
+
+/**
+ * Manages tax types based on JSON definitions.
```
bojanz, line 8: That's not true :) We should say something like "Exposes Drupal data to the tax library." Or use "entities" instead of "data".

modules/tax/src/TaxTypeRepository.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+class TaxTypeRepository implements TaxTypeRepositoryInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function get($id)
+    {
+        return entity_load('commerce_tax_type', $id);
```
bojanz, line 17: entity_load\* functions are deprecated and to be removed from core soon. There are two ways to load an entity in D8:
1) Use the static method on the entity class. For example Node::load(), Product::load(), etc. This is mostly used in cases where dependency injection is impossible (procedural code, form api callbacks)
2) Inject the entity storage class. Plenty of examples of that in Commerce

In this case we need to do #2.

modules/tax/src/TaxResolver.php
```diff
@@ -0,0 +1,41 @@
...
+use Drupal\commerce_tax\Plugin\CommerceTax\TaxTypeResolver;
+
+class TaxResolver extends BaseTaxResolver {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct() {
+    // Tax Type Resolver.
+    $taxTypeRepository = new TaxTypeRepository();
```
bojanz, line 23: Forget about "new" and \Drupal inside a class. All that needs to be injected.

bojanz:
According to Drupal conventions, classes should have a blank line at the beginning and at the end (after the opening brace before the closing brace). You have the first, but not the last newline in your class. Something to fix in the next rebase.

bojanz:
We initially tried to use plugins here, and events for resolving the store/locale.
DrupalCon discussions showed that we should try to unify the different patterns, and that tagged services represent the optimal solution. This also means we don't need to wrap the library resolvers.

I just pushed the store resolver code: https://github.com/commerceguys/commerce/commit/783d0bc8dfca7c83c2285e18c7ea1b4058df8474
The country/locale one will follow today as well.

I can then come back and do the same for taxes.

Thank you for your work David, it allowed us to evaluate an approach and get to the final idea.

PR #164, opened 2015-04-01
Issue #2462957 by bojanz, vasike: Add a store_id entity reference field ...

vasike:
...to products and orders within Entity baseFieldDefinitions.

### DIFF
--- a/modules/product/src/Entity/Product.php
+++ b/modules/product/src/Entity/Product.php
@@ -337,6 +337,23 @@ class Product extends ContentEntityBase implements ProductInterface {
         ),
       ));
 
+    $fields['store_id'] = BaseFieldDefinition::create('entity_reference')
+      ->setLabel(t('Store'))
+      ->setDescription(t('Store field for products'))
+      ->setCardinality(1)
+      ->setRequired(TRUE)
+      ->setRevisionable(TRUE)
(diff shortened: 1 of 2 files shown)


bojanz:
Committed in df8c5231745e0a57483c6b674c28a2aa6d93c26e

modules/product/src/Entity/Product.php
```diff
@@ -337,6 +337,23 @@ public static function baseFieldDefinitions(EntityTypeInterface $entityType) {
         ),
       ));
 
+    $fields['store_id'] = BaseFieldDefinition::create('entity_reference')
+      ->setLabel(t('Store'))
+      ->setDescription(t('Store field for products'))
+      ->setCardinality(1)
```
dwkitchen, line 343: Should the Cardinality only be 1? This means a product can only be related to one store.

bojanz, line 343: Yes, that is correct. We don't have a product model implemented yet. In a product + product variations system, each variation can belong to only one store, while the product (containing marketing info) can be shared.

dwkitchen, line 343: So in the multi store marketplace where stores can sell the products, the product is "owned" by the store that created it - but we have another reference field/entity that allows other stores to sell it (this makes sense as it would also need their price on it).

PR #169, opened 2015-04-06
Price display weird

googletorp:
https://www.drupal.org/node/2423237


bojanz:
This is a good bandaid for the problem for now. Thanks

PR #175, opened 2015-04-16
Add inline_entity_form to travis

joshuataylor:
As this is not on Drupal.org, the module can not be downloaded. Once this is on drupal.org, remove this line.


joshuataylor:
Related issue - https://github.com/webflo/inline_entity_form/issues/25

PR #179, opened 2015-04-16
Fix commerce product third party settings + tests

joshuataylor:


modules/product/src/Tests/ProductAdminTest.php
```diff
@@ -50,11 +50,13 @@ function testAddCommerceProductAdmin() {
...
    */
   function testAddCommerceProductExistingSkuAdmin() {
-    $product = $this->createEntity('commerce_product', array(
+    $product = $this->createEntity(
+      'commerce_product', array(
       'sku' => $this->randomMachineName(),
       'title' => $this->randomMachineName(),
       'type' => 'product'
-    ));
+    )
```
bojanz: This should still be 1 line. We can switch to using [ ] to define the array, making it clearer.
(The plan was to do that across the codebase soon anyway)

joshuataylor: The formatting is a little wonky for this entire file, will reformat this at the same time.

modules/product/commerce_product.module
```diff
@@ -100,24 +101,24 @@ function commerce_product_form_field_ui_field_edit_form_alter(array &$form, Form
...
         'visible' => array(
-          ':input[name="field[commerce_product][attribute_field]"]' => array('checked' => TRUE),
+          ':input[name="attribute_field"]' => array('checked' => TRUE),
         ),
       ),
-      '#default_value' => $default_attribute_widget,
+      '#default_value' => $defaultAttributeWidget,
     );
-    // Determine the default attribute widget title.
+//    // Determine the default attribute widget title.
```
bojanz: Accidental change?

joshuataylor: Fixed.

PR #180, opened 2015-04-16
Added information to download inline_entity_form as prerequisite

JulienD:
Because the 8.x branch of Inline_entity_form does not live on Drupal.org it needs to be installed manually from Github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,9 @@ Installation
 
 Preparation:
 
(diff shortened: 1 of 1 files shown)


bojanz:
We have moved inline_entity_form to drupal.org

joshuataylor:
Do you mind merging PR185 then?

PR #181, opened 2015-04-16
store resolver event

stijndmd:
Needs review + resolve() method to be complete


bojanz:
Explained in https://www.drupal.org/node/2464759#comment-10017425 why we went with another implementation.

Thank you for your work stijndmd, it definitely helped us arrive at the final idea.

PR #182, opened 2015-04-16
Issues #2472263 by vasike: Create Shopping Cart block (View and block Plugin)

vasike:


modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Display a list of line items added to cart.
+   *
+   * @return array
+   *   A view with cart form content.
+   */
+  public function build() {
+    // Get the Cart Order id.
+    $cart_order_id = 1;
+    // Use the block cart settings for the View to bse used for this block.
```
joshmiller83: Typo here in the comments. "for the View to be" vs. "bse"

bojanz:
No longer relevant.

PR #185, opened 2015-04-17
Now that inline_entity_form is on drupal.org, we can remove the clone

joshuataylor:
Thanks to their migration across to drupal.org, a clone is no longer needed! Yay!


joshuataylor:
Oh great this is hitting a bug with drush. Will follow up but for now I'll add a drush dl.

PR #186, opened 2015-04-17
Enable default currency based on country on install commerce_price

StryKaizer:
On enabling the commerce_price module, this will enable the currency for the default country.
If no default country has been set, USD will be used.

### DIFF
--- a/modules/price/commerce_price.install
+++ b/modules/price/commerce_price.install
@@ -6,16 +6,20 @@
  * Import the default currencies from the library.
  */
 function commerce_price_install() {
+  $currency_code = "USD";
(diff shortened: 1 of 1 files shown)


pjcdawkins:
Little style issues:
- change the (pre-existing) $currency_code into $currencyCode
- add spaces on `if($default_country){`
- change $default_country to $defaultCountry
- rebase into one commit, and reference the d.o [issue](https://www.drupal.org/node/2472525) in the commit message, i.e. `Issue #2472525:`

bojanz:
Same reasoning as for #187 (country not known early enough, needs to be an explicit action).

PR #187, opened 2015-04-17
2472087: Create a default store on installation

StryKaizer:


joshuataylor:
Also going to need schema for this.

Some like https://github.com/joshuataylor/commerce/commit/503f8f45a39558fa7cb699cce98c0f68756720f3

StryKaizer:
Thanks for the input joshuataylor.

Not sure if I'm getting your suggestion.
The property for available currencies in a store is "currency" at this moment, not plural "currencies".  

I did change it to use an array as suggested.

joshuataylor:
Hi @StryKaizer ,

I'm pretty sure it's `currencies`:

`modules/store/config/install/core.entity_form_display.commerce_store.online.default.yml`

```
- field.field.commerce_store.online.currencies
currencies:`
```

modules/store/config/install/field.storage.commerce_store.currencies.yml:
    id: commerce_store.currencie
    field_name: currencies

modules/store/config/install/field.field.commerce_store.online.currencies.yml:
      - field.storage.commerce_store.currencies
    id: commerce_store.online.currencies
    field_name: currencies
   description: 'All currencies allowed for the store.'

StryKaizer:
08:01 (StryKaizer) joshtau: I'm a bit confused about the currency/currencies stuff at https://github.com/commerceguys/commerce/pull/187#issuecomment-97636488
08:02 (StryKaizer) joshtau: when I try the code using currency, it works, when trying currencies, it does not work
08:03 (StryKaizer) joshtau: guessing there's some code elsewhere using currency too, instead of currencies?   Will look into is later
08:04 (StryKaizer) joshtau: I used firebug in the "create store" form to find out the property name, which also uses currency instead of currencies at this moment
08:05 (StryKaizer) joshtau: have to go now, will look into is later

StryKaizer:
Turned out the property was changed last week in HEAD, renamed to currencies

bojanz:
The problem I've found is that the country default variable isn't set early enough during Drupal install.
The module install hooks run first. So this code will only work if installing Commerce after Drupal is already functional.

I've been thinking about moving this code to a service that can be called by the Kickstart installer, or a Console/Drush command, sidestepping the problem.

bojanz:
I've reworked the currency importer, and added a Console command. I think that's as far as we can go.

Thank you for your exploration of this problem.

PR #192, opened 2015-04-27
Issue #2472143 - Revamp the Add Product admin form

joshuataylor:
This needs a review to make sure the code is okay.

I also recommend merging https://github.com/commerceguys/commerce/pull/187 once https://github.com/StryKaizer/commerce/commit/12ea52b26b06b94ca107bcfa0e8ff305b4434937#commitcomment-10904958 is fixed.


joshuataylor:
Also not sure if relying on https://www.drupal.org/node/2468609 would be easier..

bojanz:
As discussed with Josh, implementing the new product architecture (https://www.drupal.org/node/2506163) which allows products to belong to multiple stores means that this PR will have to be abandoned.

Thank you for your work Josh, and making us understand this problem space better.
If nothing else, it can be used as an example for similar future efforts.

PR #194, opened 2015-04-27
Issue #2478141 - Non admin users can see admin/commerce/products

joshuataylor:


modules/product/config/install/views.view.commerce_products.yml
```diff
@@ -381,17 +382,6 @@ display:
...
-        - languages
-      cacheable: false
-  page_1:
-    display_plugin: page
-    id: page_1
-    display_title: Page
-    position: 1
-    display_options:
-      display_extenders: {  }
-      path: admin/commerce/products
```
bojanz, line 393: We lost the path / whole page? Accidentally?

joshuataylor, line 393: Not sure what happened, fixed.

Also updated the other views, and removed UUID.

PR #195, opened 2015-04-27
Issue #2478631: SKUs should only be unique on a per-store basis

joshuataylor:


joshuataylor:
Actually on second thought, this will require tests.

PR #196, opened 2015-04-27
Issue #2478631: SKUs should only be unique on a per-store basis

joshuataylor:


modules/product/src/Plugin/Validation/Constraint/ProductSkuConstraintValidator.php
```diff
@@ -22,9 +22,11 @@ public function validate($items, Constraint $constraint) {
     $sku = $items->first()->value;
     if (isset($sku) && $sku !== '') {
       $productId = $items->getEntity()->product_id->value;
+      $storeId = $items->getEntity()->getStore()->id();
```
bojanz: we can use getStoreId() to skip loading it.

modules/product/src/Tests/ProductTest.php
```diff
@@ -34,7 +34,8 @@ function testAddCommerceProductExistingSku() {
     $values = [
       'sku' => $this->randomMachineName(),
       'title' => $this->randomMachineName(),
-      'type' => 'product'
+      'type' => 'product',
+      'store_id' => $this->commerce_store->id()
```
bojanz, line 38: Can we turn $this->commerce_store into $this->store?

joshuataylor, line 38: Yes, but let's do that on https://github.com/commerceguys/commerce/pull/192 to prevent making https://github.com/commerceguys/commerce/pull/192 a pain to merge.

PR #197, opened 2015-04-27
Fix issues with conversion to short arrays

joshuataylor:


bojanz:
Huh. That file got saved accidentally, it wasn't supposed to be fixed because it should be replaced by IEF. Anyway, good catch. One more reason to finish a provisional UI and get some tests running there.

PR #205, opened 2015-05-16
Add "use EntityChangedTrait" for #2428795

joshuataylor:


bojanz:
There should be a newline before the use statements, to match what core does.

bojanz:
Committed in 051c078a6d8fc64daf674c16442210e321e263ae

PR #210, opened 2015-06-10
entity_create has been deprecated

joshuataylor:


joshuataylor:
Failure looks like a testbot bug, can you requeue it to confirm @bojanz

bojanz:
Let's do the same thing we did in address, use the entity class wherever the entity type is known and static.

joshuataylor:
+1

joshuataylor:
Done

PR #212, opened 2015-06-18
Fixes use and @param

joshuataylor:


joshuataylor:
Actually there are still problems with this.

PR #214, opened 2015-06-22
Removes unused imports, fixes StoreType entity

joshuataylor:


modules/store/src/Form/StoreForm.php
```diff
@@ -9,7 +9,7 @@
 
 use Drupal\Core\Entity\ContentEntityForm;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\Language\Language;
+
```
bojanz: We don't put two newlines before the class, the added one is unneeded.

PR #215, opened 2015-07-02
Add price to commerce product testing

joshuataylor:
Once the price is on the product page we can test that the price is correct.


bojanz:
Merged the other PR, we can now expand this test.

joshuataylor:
Unless I'm slow, is there no way to test the field in the frontend in a better way?

bojanz:
We can create a test for the actual formatter class, but that's out of scope for the product test.

bojanz:
Prices are now on the variation so the test will have to live somewhere else. Closing this for now

PR #220, opened 2015-07-11
Order tests

joshuataylor:
This takes PR https://github.com/commerceguys/commerce/pull/142 , and adds order tests, order admin tests and fixes the YML files.

### DIFF
--- a/modules/order/src/Tests/CommerceOrderTestBase.php
+++ b/modules/order/src/Tests/CommerceOrderTestBase.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * @file
+ * Definition of \Drupal\commerce_order\Tests\CommerceOrderTestBase.
+ */
+
+namespace Drupal\commerce_order\Tests;
+
+use Drupal\Component\Utility\String;
+use Drupal\simpletest\WebTestBase;
+
+/**
+ * Defines base class for shortcut test cases.
+ */
+abstract class CommerceOrderTestBase extends WebTestBase {
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  public static $modules = array('commerce', 'commerce_order');
+
+  /**
+   * User with permission to administer products.
+   */
+  protected $adminUser;
+
+  protected function setUp() {
+    parent::setUp();
+    $this->adminUser = $this->drupalCreateUser(array(
+      'administer orders',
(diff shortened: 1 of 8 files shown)


modules/order/src/Tests/CommerceOrderTestBase.php
```diff
@@ -0,0 +1,107 @@
...
+namespace Drupal\commerce_order\Tests;
+
+use Drupal\commerce_product\Entity\Product;
+use Drupal\commerce_store\Entity\Store;
+use Drupal\commerce_store\Tests\StoreTestBase;
+use Drupal\Component\Utility\SafeMarkup;
+use Drupal\simpletest\WebTestBase;
+
+/**
+ * Defines base class for shortcut test cases.
```
bojanz: Not shortcut :)

modules/order/src/Tests/CommerceOrderTestBase.php
```diff
@@ -0,0 +1,107 @@
...
+ */
+abstract class CommerceOrderTestBase extends WebTestBase {
+
+  public $product;
```
bojanz: These two variables can be protected, and need documentation.

modules/order/src/Tests/CommerceOrderTestBase.php
```diff
@@ -0,0 +1,107 @@
...
+  public static $modules = [
+    'commerce',
+    'commerce_order',
+    'commerce_price',
+    'commerce_line_item',
+    'inline_entity_form'
+  ];
+
+  /**
+   * User with permission to administer products.
```
bojanz: Missing @var

modules/order/src/Tests/CommerceOrderTestBase.php
```diff
@@ -0,0 +1,107 @@
...
+      'access administration pages',
+    ));
+    // Create a store
+    $values = [
+      'name' => t('Default store'),
+      'uid' => 1,
+      'mail' => \Drupal::config('system.site')->get('mail'),
+      'type' => 'default',
+      'default_currency' => "USD",
+      'currency' => "USD",
```
bojanz: there's no "Currency" key anymore. Also, don't we need to set the address? Maybe not

joshuataylor:
Updated PR as per feedback

PR #221, opened 2015-07-17
Consider to use lionsad/drupal_ti

fabianx-ai:
Hello,

This is a test PR to see if drupal_ti would work for commerce.

drupal_ti wants to be a generic solution to empower D7 and D8 modules to enable automated testing without having to deal with all the setup of travis-ci. (especially hhvm and PHP 5.3 had been difficult to support for the docker containers)

phpunit, simpletest, behat is all supported with various browsers (chrome / firefox on selenium) and drivers ( selenium / goutte / phantomjs).

drupal_ti itself is tested on travis for all possible combinations

Running phpunit tests as part of Drupal 8 core is supported, though right now needs some custom code. (e.g. https://github.com/md-systems/pathauto/pull/51/files) That will be incorporated into the code base soon in the 1.3 branch. Given commerce does not yet have such tests that is probably acceptable.

Currently (in this PR) the cache has been disabled, but drupal_ti supports an advanced caching layer in dev-master (will be in 1.4.0), hence I would do that separately.

### DIFF
--- a/.travis-before-script.sh
+++ b/.travis-before-script.sh
@@ -17,6 +17,7 @@ cd "$DRUPAL_TI_DRUPAL_DIR"
 
 	git clone --branch 8.x-1.x http://git.drupal.org/project/composer_manager.git
 	git clone --branch 8.x-1.x http://git.drupal.org/project/address.git
+	git clone --branch 8.x-1.x http://git.drupal.org/project/inline_entity_form.git
 )
 
 # Enable and run composer_manager.
@@ -33,9 +34,6 @@ rm -rf vendor
 composer drupal-rebuild
 composer update --prefer-source -n --verbose
 cd ..
-
-# Download more dependencies.
-drush dl -y inline_entity_form
 
 # Enable main module and submodules.
 drush en -y commerce commerce_product commerce_order
--- a/.travis-before-script.sh
+++ b/.travis-before-script.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -e $DRUPAL_TI_DEBUG
+
+# Ensure the right Drupal version is installed.
+# Note: This function is re-entrant.
+drupal_ti_ensure_drupal
+
+# Add custom modules to drupal build.
+cd "$DRUPAL_TI_DRUPAL_DIR"
+
+# Download custom branches of address and composer_manager.
+(
+	# These variables come from environments/drupal-*.sh
+	mkdir -p "$DRUPAL_TI_MODULES_PATH"
+	cd "$DRUPAL_TI_MODULES_PATH"
+
+	git clone --branch 8.x-1.x http://git.drupal.org/project/composer_manager.git
+	git clone --branch 8.x-1.x http://git.drupal.org/project/address.git
+)
+
+# Enable and run composer_manager.
+drush pm-enable composer_manager --yes
+drush composer-manager-init
+
+# Ensure the module is linked into the codebase.
+drupal_ti_ensure_module_linked
+
+# Rebuild core dependencies.
+# @todo Is that really needed?
+cd core
+rm -rf vendor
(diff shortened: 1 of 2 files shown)


fabianx-ai:
The tests failed (as HEAD is broken), but drupal_ti works well.

Is it really necessary to delete all of core/vendor?

Is running composer update not enough?

joshuataylor:
When commerce first started to use TravisCI, everything was great. Recently we've seen many failures in terms of travisci being buggy, not completing setup and other issues. So I'm all for changing to use something better than using .travis.yml, I'll look into drupal_ti as this looks like an exciting solution moving forward.

The reason we delete core/vendor is there is a 'bug' in composer where if you don't it'll try and download the latest source instead of cloning, and unless you're authorized with Github your IP can get banned from downloading zips (which TravisCI is) so deleting the vendor file is necessary for a full clone.

bojanz:
Ping @mglaman

fabianx-ai:
@joshuataylor Yes, that is one reason why I worked hard to ensure everything works on the container infrastructure. That infrastructure has been way more stable for myself.

## 

Oh, but once those are .git directories composer would correctly git fetch those?

e.g. we could cache the result and just replace core/vendor with the cached version?

fabianx-ai:
I just added the unit tests, those work fine, too :).

joshuataylor:
@LionsAd  , see https://github.com/composer/composer/issues/3826 .

Caching is also not recommended, given how often the dependencies update (addressing, intl, tax, pricing), and we might test against an older version as right now we use dev-master instead of a tagged release.

joshuataylor:
Also failures are not caused by drupal_ti , instead a patch that changed a lot of things but had no tests (https://www.drupal.org/node/2506163).

Testing against more versions of PHP is also wonderful.

bojanz:
I'm going to get the current tests passing again, then we can rebase this PR, make sure it works, and merge it.

bojanz:
@LionsAd Can you rebase? The build has been fixed.

fabianx-ai:
@bojanz Rebased and ported over the inline_entity_form change :).

fabianx-ai:
Uhm, MySQL server has gone away. panopoly/panopoly has some settings for that, but I wonder if we need to up max-packet size or if that was just a travis glitch.

Making a comment change to trigger a new test.

fabianx-ai:
@bojanz The mysql optimization did the trick, which I will now push to upstream into a bugfix release.

But tests fail again (PHP-7):

```
Interface 'CommerceGuys\Tax\Model\TaxRateInterface' not found
```

Fatal error: Interface 'CommerceGuys\Tax\Model\TaxRateInterface' not found in /home/travis/build/commerceguys/commerce/modules/tax/src/Entity/TaxRate.php on line 54

Maybe that interface was missed?

bojanz:
@LionsAd The interface is a part of the library retrieved via Composer. Anyway, I'm fine with removing PHP 7 testing for now, since 7 itself seems too unstable to be reliable.

joshuataylor:
Also php7 tests were passing locally with nightly

fabianx-ai:
@bojanz @joshuataylor: I was stupid and misunderstood something with how composer-manager works.

Sorry for the confusion it works now.

I will move some things upstream, release a 1.4.1 bug fix release, then update this PR.

Should get to that on the weekend.

fabianx-ai:
@bojanz @joshuataylor Rebased, passing tests and cleaned up.

All good now and ready for merging :).

I did not see any timeouts on the container infrastructure in relation to composer, but in case we see some we can easily add travis_retry to all composer commands in the future.

joshuataylor:
Great! Yes, lets keep an eye out for composer issues (as we've had quite a few in the past), this is a great step forward to solving the pains with CI. :+1: x1000

PR #222, opened 2015-07-21
Fix commerce product tests

joshuataylor:
Waiting for green, tests are green locally.

### DIFF
--- a/modules/product/src/Tests/ProductAdminTest.php
+++ b/modules/product/src/Tests/ProductAdminTest.php
@@ -8,6 +8,7 @@
 namespace Drupal\commerce_product\Tests;
 
 use Drupal\commerce_product\Entity\Product;
+use Drupal\commerce_product\Entity\ProductVariation;
 
 /**
  * Create, view, edit, delete, and change products and product types.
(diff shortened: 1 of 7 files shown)


modules/product/src/Tests/ProductAdminTest.php
```diff
@@ -43,37 +50,15 @@ function testAddCommerceProductAdmin() {
...
-      ->count()
+    // Test product variations
+    $productVariation = \Drupal::entityQuery('commerce_product_variation')
+      ->condition("sku", $productVariationValues['variations[form][inline_entity_form][sku][0][value]'])
+      ->range(0, 1)
       ->execute();
-    $this->assertEqual($duplicateCommerceProductSkus, 1, "Only one product exists");
 
-    $this->assertText("is already in use", "Commerce Product failure text is showing");
+    $productVariation = entity_load("commerce_product_variation", current($productVariation));
```
bojanz: we should use ProductVariation::load() here. entity_load() is deprecated.

modules/store/config/install/commerce_store.settings.yml
```diff
@@ -1 +1 @@
-default_store: ''
+default_store: 0229d4d9-abd4-4a8a-8c05-5ad016240bab
```
bojanz: Accidental?

PR #225, opened 2015-07-28
Issue #2511266 by cbanman: Create a service for determining variation availability

crbanman:


modules/order/src/AvailabilityCheckerInterface.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Drupal\commerce\LineItemSourceInterface;
+
+/**
+ * Defines the interface for availability checkers.
+ */
+interface AvailabilityCheckerInterface {
+
+  /**
+   * Checks if a source applies to this checker.
```
bojanz: Determines whether the checker can be used with the given source entity.

modules/order/src/AvailabilityCheckerInterface.php
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * @param \Drupal\commerce\LineItemSourceInterface $source
+   *   The source.
+   *
+   * @return bool
+   *   Returns TRUE if the source applies to this checker, returns FALSE if the
```
bojanz: Reformulate the @return to match the suggestion above.

crbanman:
Made the requested doc fixes.

PR #230, opened 2015-08-05
Entity link templates should be paths

floretan:
Entity link templates should be paths instead of route names, and Drupal throws fatal errors if that's not the case.

At the moment, if anyone enables the Commerce Payment module (just to see what's the current state), they put their Drupal installation in an unusable state. The fatal errors even prevent the module from being uninstalled.


bojanz:
I've been telling people that the Payment module is broken, incomplete, waiting for a rewrite, so it shouldn't be enabled, but there's no harm in fixing this crash. Thanks.

PR #234, opened 2015-08-28
Moved Availability Manager and Checker to commerce core

pcambra:


src/AvailabilityCheckerInterface.php
```diff
@@ -2,10 +2,10 @@
...
 /**
  * @file
- * Contains \Drupal\commerce_order\AvailabilityCheckerInterface.
+ * Contains \Drupal\commerce\AvailabilityCheckerInterface.
  */
 
-namespace Drupal\commerce_order;
+namespace Drupal\commerce;
 
 use Drupal\commerce\PurchasableEntityInterface;
```
bojanz: We don't need this use statement anymore then?

bojanz:
Let's fix the use statements and merge it :)

pcambra:
True, those classes are now in the same namespace*, removed them :)

PR #235, opened 2015-08-28
Updated composer manager instructions for install

pcambra:


bojanz:
This will need some more work, I'll leave it be until I tag a new composer_manager release.

bojanz:
We iterated on this in #240

PR #244, opened 2015-09-25
A line item matcher

dwkitchen:
Is this the basics of what you are after?

### DIFF
--- a/modules/order/src/Event/LineItemComparisonFieldsEvent.php
+++ b/modules/order/src/Event/LineItemComparisonFieldsEvent.php
@@ -1,0 +1,61 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_order\Event\LineItemComparisonFieldsEvent.
+ */
+
+namespace Drupal\commerce_order\Event;
+
+use Drupal\commerce\PurchasableEntityInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Defines the available countries event.
+ *
+ * @see \Drupal\address\Event\OrderEvents
+ */
+class LineItemComparisonFieldsEvent extends Event {
+
+  /**
+   * The comparison fields.
+   *
+   * A list of fields.
+   *
+   * @var array
+   */
+  protected $comparisonFields;
+
+  /**
+   * The purchasable entity.
+   *
+   * @var \Drupal\commerce\PurchasableEntityInterface
+   */
+  protected $entity;
+
+  /**
+   * Constructs a new LineItemComparisonFieldsEvent object.
+   *
+   * @param \Drupal\commerce\PurchasableEntityInterface $entity
+   *   The Purchasable Entity
+   *
+   * @param array $comparisonFields
+   *   An initial list of comparison fields.
+   */
+  public function __construct(PurchasableEntityInterface $entity, array $comparisonFields) {
+    $this->entity = $entity;
+    $this->comparisonFields = $comparisonFields;
+  }
+
+  /**
+   * Gets the comparison fields.
+   *
+   * @return array
+   *   The comparison fields.
+   */
+  public function getComparisonFields() {
+    return $this->comparisonFields;
+  }
+
+}
+
--- a/modules/order/src/LineItemMatcher.php
+++ b/modules/order/src/LineItemMatcher.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * @file
+ * Contains \Drupal\commerce_order\LineItemMatcher.
+ */
+
+namespace Drupal\commerce_order;
+
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Drupal\commerce\PurchasableEntityInterface;
+use Drupal\commerce_order\Entity\OrderInterface;
+use Drupal\commerce_order\Entity\LineItemInterface;
+use Drupal\commerce_order\Event\OrderEvents;
+use Drupal\commerce_order\Event\LineItemComparisonFieldsEvent;
+
+/**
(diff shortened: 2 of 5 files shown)


dwkitchen:
What about customisable line items? I guess this is handled with `drupal_alter('commerce_cart_product_comparison_properties', $comparison_properties, clone($line_item));` in Commerce 1.x

How will customisations be handled by the Purchasable Entity?

dwkitchen:
I still don't get where a customisable product would have its customisations. It doesn't look like they are in the PurchasableEntity, so I think we need a second variable to this matcher if a dummy line item is not created.

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,63 @@
...
+   */
+  protected $order;
+
+  /**
+   * Creates a LineItemMatcherBase instance.
+   *
+   * @param OrderInterface $order
+   */
+  public function __construct($order)
+  {
```
bojanz: Opening brace on wrong position.

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * The order.
+   *
+   * @var Drupal\commerce_order\Entity\OrderInterface
+   */
+  protected $order;
+
+  /**
+   * Creates a LineItemMatcherBase instance.
```
bojanz: Constructs a new $X object.

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * @param OrderInterface $order
```
bojanz: Param must have full namespace, description in the line below.

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,60 @@
...
+namespace Drupal\commerce_order;
+
+use Drupal\commerce\PurchasableEntityInterface;
+use Drupal\commerce_order\Entity\OrderInterface;
+use Drupal\commerce_order\Entity\LineItemInterface;
+use Drupal\commerce_order\Event\OrderEvents;
+use Drupal\commerce_order\Event\LineItemComparisonFieldsEvent;
+
+/**
+ * Class LineItemMatcher
```
bojanz, line 17: Missing docs.

modules/order/src/Event/LineItemComparisonFieldsEvent.php
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Constructs a new LineItemComparisonFieldsEvent object.
+   *
+   * @param \Drupal\commerce\PurchasableEntityInterface $entity
+   *   The Purchasable Entity
+   */
+  public function __construct(PurchasableEntityInterface $entity) {
+    $this->entity = $entity;
+    $this->comparisonFields = ['type', 'id'];
```
bojanz: We should pass these to the constructor, to keep the logic in the Matcher.

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function matchAll(PurchasableEntityInterface $entity, OrderInterface $order) {
+
+    // Get the comparison fields from all modules altering them.
+    $event = new LineItemComparisonFieldsEvent($entity);
+    $event_dispatcher = \Drupal::service('event_dispatcher');
```
bojanz: Inject the event dispatcher.

dwkitchen: This is how it is done in the address module. How should it be done?

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,81 @@
...
+ * @package Drupal\commerce_order
+ */
+class LineItemMatcher implements LineItemMatcherInterface {
+
+  /**
+   * An Event Dispatcher.
+   *
+   * @var \Symfony\Component\EventDispatcher\EventDispatcherInterface
+   */
+  protected $event_dispatcher;
```
bojanz, line 30: Class attributes must be camelCase.

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,81 @@
...
+ * Class LineItemMatcher
+ *
+ * Match an existing line item for the PurchasableEntity.
+ *
+ * @package Drupal\commerce_order
+ */
+class LineItemMatcher implements LineItemMatcherInterface {
+
+  /**
+   * An Event Dispatcher.
```
bojanz, line 26: "The event dispatcher."

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @var \Symfony\Component\EventDispatcher\EventDispatcherInterface
+   */
+  protected $event_dispatcher;
+
+  /**
+   * Constructs a new LineItemMatcher.
```
bojanz, line 33: Constructs a new X object.

modules/order/src/LineItemMatcher.php
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * @param \Symfony\Component\EventDispatcher\EventDispatcherInterface $event_dispatcher
```
bojanz, line 35: Missing description below param ("The event dispatcher.")

bojanz:
Committed a more advanced version of this in d25878624dd27bc0f7090102a1f37e1e1399cc38.

PR #247, opened 2015-09-25
Create a country cache context per d.o issue 2571393

jaredsmith:
This is my first time seeing cache contexts, so I feel a bit overwhelmed (and uninformed) if I'm doing this correctly, but I tried to model my patch on the existing cache contexts in core.

### DIFF
--- a/src/Cache/Context/CountryCacheContext.php
+++ b/src/Cache/Context/CountryCacheContext.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce\Cache\Context\CountryCacheContext
+ */
+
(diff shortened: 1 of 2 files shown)


src/Cache/Context/CountryCacheContext.php
```diff
@@ -0,0 +1,57 @@
...
+   * 
+   * @var \Drupal\commerce\Country
+   */
+  protected $countryCode;
+  
+  /**
+   * Constructs a new CountryCacheContext service.
+   * 
+   * @param \Drupal\commerce\Country $country
+   *   The theme manager.
```
bojanz: Not the theme manager :)

PR #248, opened 2015-09-25
Issue #2574793: Tax rate machine name was not able to be changed if default

mikaelkundert:
## Problem

See validation for "Default" field:

``` php
  /**
   * Validates that there is only one default per tax type.
   */
  public function validateDefault(array $element, FormStateInterface $form_state, array $form) {
    $taxRate = $this->getEntity();
    $default = $element['#value'];
    if ($default) {
      $loadedTaxRates = $this->taxRateStorage->loadByProperties([
        'type' => $form_state->getValue('type'),
      ]);
      foreach ($loadedTaxRates as $rate) {
        if ($rate->getId() !== $taxRate->getId() && $rate->isDefault()) {
          $form_state->setError($element, $this->t('Tax rate %label is already the default.', [
            '%label' => $rate->label(),
          ]));
          break;
        }
      }
    }
  }
```

We can see, that there is already an valid condition for this, but it turns out that entity gets updated prior to form validation in `\Drupal\Core\Form\FormBuilder::processForm()`.

See how `doBuildForm()` method gets called before `validateForm()` in http://cgit.drupalcode.org/drupal/tree/core/lib/Drupal/Core/Form/FormBuilder.php#n537
## Proposed solution

I tried XDebug original entity or values from `$form_state` and tried to seek for any type of `originalEntity` traits or something without success.

I ended up implementing `TaxRateForm::originalTaxRate` property with getter and setter methods and populate it on form build, and use its value in the validation.

### DIFF
--- a/modules/tax/src/Form/TaxRateForm.php
+++ b/modules/tax/src/Form/TaxRateForm.php
@@ -8,6 +8,7 @@
 namespace Drupal\commerce_tax\Form;
 
 use Drupal\Core\Entity\EntityForm;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityStorageInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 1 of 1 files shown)


mikaelkundert:
Drupal.org issue link: https://www.drupal.org/node/2574793

mikaelkundert:
Why it failed? Connectivity issues to Github during build?

mikaelkundert:
Turns out that we may just use getOriginalId() which. Simplified the code, it became onliner! \o/

PR #250, opened 2015-09-26
Return the cache contexts from price formatters, issue 2571395

jaredsmith:
I'm flying a bit blind here, but I _think_ this is what you were looking for.

I'm assuming I don't need to also add this to modules/price/src/Plugin/Field/FieldFormatter/PriceDefaultFormatter.php.


bojanz:
We need it in both places. The current change looks good.

jaredsmith:
Updated to add to PriceDefaultFormatter.php as well.

PR #254, opened 2015-09-26
Issue #2575485: When deleting tax types or tax rates, delete their children

mikaelkundert:


mikaelkundert:
Drupa.org issue: https://www.drupal.org/node/2575485

PR #257, opened 2015-09-27
Add field_access to commerce_cart module

jaredsmith:
Here's my initial stab at a pull request for this issue.  

I'm not familiar with the Field API, so I'm mostly just copying/pasting from the 1.x code, so feel free to point out anything that might have changed in for d8.

### DIFF
--- a/modules/cart/commerce_cart.module
+++ b/modules/cart/commerce_cart.module
@@ -122,6 +122,44 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
 }
 
 /**
+ * Implements hook_form_FORM_ID_alter() for 'field_ui_field_edit_form'.
+ */
+function commerce_cart_form_field_config_edit_form_alter(array &$form, FormStateInterface $form_state) {
+  $field = $form_state->getFormObject()->getEntity();
+  if (!$field->isLocked() && $field->get('entity_type') == 'commerce_line_item') {
+
+    // Get the current instance's line item form settings for use as default values.
+    $instance = $form['#instance'];
+    if (empty($instance['commerce_cart_settings']) || !is_array($instance['commerce_cart_settings'])) {
+      $commerceCartSettings = array();
+    }
+    else {
+      $commerceCartSettings = $instance['commerce_cart_settings'];
+    }
+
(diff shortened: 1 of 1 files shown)


modules/cart/commerce_cart.module
```diff
@@ -122,6 +122,44 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
 }
 
 /**
+ * Implements hook_form_FORM_ID_alter() for 'field_ui_field_edit_form'.
+ */
+function commerce_order_form_field_config_edit_form_alter(array &$form, FormStateInterface $form_state) {
```
bojanz: Wrong function name. Needs to be commerce_cart_form_field_config_edit_form_alter. The current version never gets called :)

modules/cart/commerce_cart.module
```diff
@@ -122,6 +122,44 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
...
+  $field = $form_state->getFormObject()->getEntity();
+  if ($field->isLocked() === false && $field->get('entity_type') == 'commerce_line_item') {
```
bojanz: we can do !$field->isLocked()

modules/cart/commerce_cart.module
```diff
@@ -122,6 +122,44 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
...
+
+    // Get the current instance's line item form settings for use as default values.
+    $instance = $form['#instance'];
```
bojanz, line 132: I doubt $form['#instance'] exists. Also, we don't prefix variables with commerce

jaredsmith, line 132: OK, as I said -- I don't understand how this works in D8, so I can definitely use either some other examples to look at, or some guidance on how this works.

modules/cart/commerce_cart.module
jaredsmith: Fixed.

modules/cart/commerce_cart.module
jaredsmith: Fixed.

bojanz:
We decided to implement this differently, by using a form mode instead.

PR #261, opened 2015-10-07
 Issues #2464347 by bojanz, vasike: Create Cart page - Rebuild the cart view for cart page.

vasike:
Rebuild (put back) the cart view for cart page.

### DIFF
--- a/modules/cart/commerce_cart.info.yml
+++ b/modules/cart/commerce_cart.info.yml
@@ -8,4 +8,5 @@ dependencies:
   - commerce_order
   - commerce_product
 config_devel:
+  - commerce_cart.settings
   - views.view.commerce_cart_form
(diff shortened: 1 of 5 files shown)


joshuataylor:
Can you do a rebase so it's easier to review?

vasike:
@joshuataylor is it now better?

joshuataylor:
Awesome, we also need to double check that the config schema etc actually exists, which I'm having problems with at the moment in a related issue.

vasike:
@joshuataylor i'm afraid i do not know what you mean. where do we need this double check, etc?

joshuataylor:
Sorry, I was talking nonsense. I'm looking at this now.

bojanz:
Cart settings need to be per order type, not global.

bojanz:
Committed the finished version in 18e3e521b6f44ed8d60d9f7f4c4d5238215f874c.

PR #263, opened 2015-10-08
Fix tests to pass again

joshuataylor:


bojanz:
These langcodes are actually supposed to be lowercase, that's how Drupal passes them around. We need to see why/when they get uppercased.

modules/product/commerce_product.module
```diff
@@ -141,7 +141,7 @@ function commerce_product_add_variations_field($product_type) {
         'handler' => 'default',
         'handler_settings' => [
           'target_bundles' => [
-            $product_type->getVariationType(),
+            $product_type->id() => $product_type->id(),
```
bojanz, line 144: That can't be true, because we need to specify bundles of the commerce_product_variation entity type, not commerce_product.

bojanz:
Pushed a fix for the locale codes, the PR can now be rebased with that part removed.

joshuataylor:
I'll get this updated today.

bojanz:
Fixed in 8887d059f134c97560ab3235a808abd42ff7aa04 and 37993feb0cfbacc32d192361fd349ad204f8cf14.

The store type test was failing because of the "title as block" change, we weren't placing the title block in the setUp(). The product tests were throwing exceptions because we were calling the add variation field function on an incomplete product type (no variationType specified).

Thanks for getting this work started!

PR #264, opened 2015-10-21
Update readme to account for change in composer_manager

NickDickinsonWilde:


bojanz:
I need to update the README to point to docs.drupalcommerce.org instead, but let's fix this first anyway.

Thanks!

PR #265, opened 2015-10-26
Configure composer before requiring drupal-ti

webflo:


bojanz:
Latest IRC idea is to include the token unencrypted, and make sure it's read-only.
That way it will work for forks as well.

I have no clue why removing --prefer-source fails, looks like it will have to stay.

PR #266, opened 2015-10-26
Just another composer test

webflo:


bojanz:
Let's go back to #265.

PR #267, opened 2015-10-26
Modernize forms #2591829

czigor:
https://www.drupal.org/node/2591829

### DIFF
--- a/modules/product/src/Form/ProductTypeForm.php
+++ b/modules/product/src/Form/ProductTypeForm.php
@@ -7,12 +7,14 @@
 
 namespace Drupal\commerce_product\Form;
 
-use Drupal\Core\Entity\EntityForm;
+use Drupal\Core\Entity\BundleEntityFormBase;
 use Drupal\Core\Entity\EntityManagerInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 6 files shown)


czigor:
Other forms are coming soon.

bojanz:
Payment info won't be translatable, so we shouldn't add language settings there.

For other forms, make sure $language_configuration is $languageConfiguration.

czigor:
TaxTypeForm seems a bit weird, should I touch it? For id it has  a textfield instead of machine_name and I could not find a TaxType entity. Otherwise I'm ready.

bojanz:
No, you don't need to touch that one.

modules/product/src/Form/ProductTypeForm.php
```diff
@@ -83,7 +86,27 @@ public function form(array $form, FormStateInterface $form_state) {
       '#description' => t('Products of this type represent digital services.')
     ];
 
-    return $form;
+    if ($this->moduleHandler->moduleExists('language')) {
+      $form['language'] = array(
```
bojanz, line 90: Can we use the short array syntax here? [ instead of array(

bojanz:
Committed as 0e675b7e5e49ddeafeb7b67975129a1f402e2238, made the short array syntax change manually, as well as the #disabled fix (using $type->isLocked()) that was forgotten.

PR #268, opened 2015-11-02
Disable machine name field for existing bundles #2606524

czigor:
This was changed to isLocked() by accident.


bojanz:
Sorry, I already reverted this in 533b79fb704e68039d6065cd96447a3b9fc26922

czigor:
Hmm... the revert only removes the isLocked() calls. We need isNew() .

PR #269, opened 2015-11-04
Issue #2600564: Integrate profile into commerce

subhojit777:
This will cover:
- Add the profile module dependency to order
- Create the “billing” profile bundle and export it to order/config/install
- Create the "address" field on the "billing" profile and export it to order/config/install
- Create a billing_profile reference field on the order (base field definition), add a getter and a setter

### DIFF
--- a/modules/order/config/install/field.field.profile.billing.address.yml
+++ b/modules/order/config/install/field.field.profile.billing.address.yml
@@ -1,0 +1,32 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.profile.address
+    - profile.type.billing
+  module:
+    - address
+id: profile.billing.address
+field_name: address
+entity_type: profile
+bundle: billing
+label: Address
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  available_countries: {  }
+  fields:
+    administrativeArea: administrativeArea
+    locality: locality
+    dependentLocality: dependentLocality
+    postalCode: postalCode
+    sortingCode: sortingCode
+    addressLine1: addressLine1
+    addressLine2: addressLine2
+    organization: organization
+    recipient: recipient
+  langcode_override: ''
+field_type: address
--- a/modules/order/config/install/field.field.profile.billing.address.yml
+++ b/modules/order/config/install/field.field.profile.billing.address.yml
@@ -12,7 +12,7 @@ entity_type: profile
 bundle: billing
 label: Address
 description: ''
-required: false
+required: true
 translatable: false
 default_value: {  }
 default_value_callback: ''
--- a/modules/order/config/install/field.storage.profile.address.yml
+++ b/modules/order/config/install/field.storage.profile.address.yml
@@ -10,7 +10,7 @@ entity_type: profile
 type: address
 settings: {  }
 module: address
-locked: false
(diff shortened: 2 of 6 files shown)


bojanz:
Since we're providing the field ourselves, can we name it address instead of field_address?

subhojit777:
@bojanz corrected.

modules/order/config/install/field.field.profile.billing.address.yml
```diff
@@ -0,0 +1,32 @@
...
+    - profile.type.billing
+  module:
+    - address
+id: profile.billing.address
+field_name: address
+entity_type: profile
+bundle: billing
+label: Address
+description: ''
+required: false
```
bojanz: Should be required :)

modules/order/config/install/field.storage.profile.address.yml
```diff
@@ -0,0 +1,18 @@
...
+  module:
+    - address
+    - profile
+id: profile.address
+field_name: address
+entity_type: profile
+type: address
+settings: {  }
+module: address
+locked: false
```
bojanz: Let's lock it. This field should never be removed.

modules/order/src/Entity/Order.php
```diff
@@ -463,6 +494,21 @@ public static function baseFieldDefinitions(EntityTypeInterface $entityType) {
       ->setLabel(t('Data'))
       ->setDescription(t('A serialized array of additional data.'));
 
+    $fields['billing_address'] = BaseFieldDefinition::create('entity_reference')
```
bojanz: We are referencing the billing profile, not the billing address. You'll need to correct the field name and getters/setters accordingly.

bojanz:
Almost there!

subhojit777:
@bojanz corrected :) is it okay now?

bojanz:
Almost there!

You added methods to the Order entity, but not to the OrderInterface. Did those changes get lost by mistake?

modules/order/src/Entity/Order.php
```diff
@@ -463,6 +494,21 @@ public static function baseFieldDefinitions(EntityTypeInterface $entityType) {
       ->setLabel(t('Data'))
       ->setDescription(t('A serialized array of additional data.'));
 
+    $fields['billing_profile'] = BaseFieldDefinition::create('entity_reference')
+      ->setLabel(t('Billing profile'))
+      ->setDescription(t('Billing profile'))
+      ->setRequired(TRUE)
+      ->setSetting('target_type', 'profile')
+      ->setSetting('handler', 'default')
+      ->setTranslatable(TRUE)
```
bojanz, line 504: We also need a ->setSetting('handler_settings', ['target_bundles' => ['profile']]); This will limit the reference field to the appropriate bundle.

subhojit777:
@bojanz changes made.

modules/order/src/Entity/OrderInterface.php
```diff
@@ -151,4 +151,40 @@ public function getEmail();
    */
   public function setEmail($mail);
 
+  /**
+   * Get billing profile.
```
bojanz: Verbs in docblocks must be in the third person. "Gets the billing profile."

subhojit777:
@bojanz comment corrected :)

PR #270, opened 2015-11-06
EntityManager

czigor:
Apart from one occurence we replace EntityManager with EntityTypeManager.

### DIFF
--- a/modules/product/src/Tests/ProductAdminTest.php
+++ b/modules/product/src/Tests/ProductAdminTest.php
@@ -82,7 +82,7 @@ class ProductAdminTest extends ProductTestBase {
     $this->assertText(t("Are you sure you want to delete the product @product?", ['@product' => $product->getTitle()]), "Commerce Product deletion confirmation text is showing");
     $this->assertText(t('This action cannot be undone.'), 'The product deletion confirmation form is available');
     $this->drupalPostForm(NULL, NULL, t('Delete'));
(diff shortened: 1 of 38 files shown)


bojanz:
Are the tests passing for you locally? The travis build failed (The StoreTest, to be precise).

czigor:
No, tests are failing locally as well. Will work on it next week.

bojanz:
Great! This is now ready. Can you please rebase & squash so that we're left with one commit?

czigor:
Tests fail because of OrderAdminTest::testCreateOrder(). They fail in 8.x-2.x as well since the billing profile field has been made required for the order entity in 6e07c8984155800d84438084a0b82efcfbe09602.

modules/product/src/Tests/ProductAdminTest.php
```diff
@@ -71,6 +71,56 @@ function testAddProductAdmin() {
   }
 
   /**
+   * Tests adding/removing stores to a product via the admin.
+   */
+  function testUpdateProductAdmin() {
```
bojanz: You are adding back a removed test by accident.

bojanz:
Let's fix OrderAdminTest elsewhere. Remove the test and this is ready to go.

czigor:
Thanks! I have removed the test.

bojanz:
Thanks for the big effort, András!

PR #271, opened 2015-11-13
Order admin test fix

czigor:
https://www.drupal.org/node/2612672
The bundle of the Billing profile field on the order form has been changed from 'profile' to 'billing' as well.

### DIFF
--- a/modules/order/src/Tests/OrderAdminTest.php
+++ b/modules/order/src/Tests/OrderAdminTest.php
@@ -8,6 +8,7 @@
 namespace Drupal\commerce_order\Tests;
 
 use Drupal\commerce_order\Entity\Order;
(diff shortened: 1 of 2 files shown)


modules/order/src/Tests/OrderAdminTest.php
```diff
@@ -17,6 +18,26 @@
 class OrderAdminTest extends OrderTestBase {
 
   /**
+   * The profile to test against
+   *
+   * @var \Drupal\profile\Entity\Profile
+   */
+  protected $billing_profile;
```
bojanz: this should be $billingProfile, to match the coding standard.

mglaman:
:+1:

czigor:
This test is failing in 8.x-2.x as well.

PR #272, opened 2015-11-19
EntityBundle views handler

czigor:
https://www.drupal.org/node/2616450

The EntityBundle views field handler is available for every ContentEntity. The handler displays the bundle name if there are more than one bundles for that entity. If there's only one bundle then it displays nothing.

### DIFF
--- a/commerce.views.inc
+++ b/commerce.views.inc
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * Implements hook_views_data_alter().
+ */
+function commerce_views_data_alter(array &$data) {
+  // Override the default bundle views field handler.
+  $entity_types = \Drupal::service('entity_type.manager')->getDefinitions();
+  foreach ($entity_types as $entity_type) {
+    // Only override for commerce content entities.
+    if ($entity_type instanceof \Drupal\Core\Entity\ContentEntityType && strpos($entity_type->id(), 'commerce_') === 0) {
+      // Bundle is stored either in the data or the base table.
+      if (!empty($data[$entity_type->getDataTable()])) {
+        $data[$entity_type->getDataTable()][$entity_type->getKey('bundle')]['field']['id'] = 'entity_bundle';
+      }
+      else {
+        $data[$entity_type->getBaseTable()][$entity_type->getKey('bundle')]['field']['id'] = 'entity_bundle';
+      }
+    }
+  }
+}
--- a/src/Plugin/views/field/EntityBundle.php
+++ b/src/Plugin/views/field/EntityBundle.php
@@ -1,0 +1,122 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce\Plugin\views\field\EntityBundle.
+ */
+
+namespace Drupal\commerce\Plugin\views\field;
+
+use Drupal\views\Plugin\views\field\Field;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Entity\EntityManagerInterface;
+use Drupal\Core\Field\FieldTypePluginManagerInterface;
+use Drupal\Core\Field\FormatterPluginManager;
+use Drupal\Core\Language\LanguageManagerInterface;
+use Drupal\Core\Render\RendererInterface;
+use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
+
+/**
+ * Displays the entity bundle.
+ *
+ * Can be configured to show nothing when there's only one possible bundle,
+ * allowing the 'Hide empty column' table setting to hide the column.
+ *
+ * @ingroup views_field_handlers
+ *
+ * @ViewsField("entity_bundle")
+ */
+class EntityBundle extends Field {
+
+  /**
+   * The entity type bundle info service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface
+   */
+  protected $entityTypeBundleInfo;
+
+  /**
+   * Constructs a \Drupal\commerce\Plugin\views\field\EntityBundle object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   *   The field formatter plugin manager.
+   * @param \Drupal\Core\Field\FormatterPluginManager $formatter_plugin_manager
+   *   The field formatter plugin manager.
+   * @param \Drupal\Core\Field\FieldTypePluginManagerInterface $field_type_plugin_manager
+   *   The field plugin type manager.
+   * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
+   *   The language manager.
+   * @param \Drupal\Core\Render\RendererInterface $renderer
+   *   The renderer.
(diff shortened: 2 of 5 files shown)


src/Plugin/views/field/EntityBundle.php
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function render(ResultRow $row) {
+    // Get entity.
+    $entity_type = $this->getEntityType();
+    // Get the bundles of this entity.
+    $bundles = \Drupal::service('entity_type.bundle.info')->getBundleInfo($entity_type);
```
bojanz: We can inject this service.

src/Plugin/views/field/EntityBundle.php
```diff
@@ -0,0 +1,47 @@
...
+
+    // If there's only one bundle, don't dislay it.
```
bojanz: There should be a setting (enabled by default) in the views handler for this behavior, so that it can be turned off.

commerce.views.inc
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Implements hook_views_data_alter().
+ */
+function commerce_views_data_alter(array &$data) {
+  // Views field handler to display the entity bundle. If there's only one
+  // bundle for the entity type, don't display anything.
+  $content_entities = array_keys(\Drupal::service('entity_type.repository')->getEntityTypeLabels(TRUE)['Content']);
+  foreach ($content_entities as $content_entity) {
+    $data[$content_entity]['entity_bundle'] = [
```
bojanz: We should be overriding the handler for the existing/real bundle fields instead.

czigor:
@bojanz Fixed all the things, can you have a look?

src/Plugin/views/field/EntityBundle.php
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * The entity type bundle info service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface
+   */
+  protected $entityTypeBundleInfo;
+
+  /**
+   * Constructs a \Drupal\field\Plugin\views\field\Field object.
```
bojanz: "Constructs a new EntityBundle object."

src/Plugin/views/field/EntityBundle.php
```diff
@@ -0,0 +1,117 @@
...
+      '#title' => t('Hide if there\'s only one bundle.'),
+      '#default_value' => $this->options['hide_one_bundle'],
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function renderItems($items) {
+    // Check option setting if we want to display bundle name.
```
bojanz: "// Check option setting if we want to display bundle name." -> this comment is redundant, it didn't tell us anything new. When commenting focus on the "why", not the "what". In this case you can simply omit the comment.

src/Plugin/views/field/EntityBundle.php
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildOptionsForm(&$form, FormStateInterface $form_state) {
+    parent::buildOptionsForm($form, $form_state);
+
+    // Option to hide the bundle name if there's only one bundle.
+    $form['hide_one_bundle'] = [
```
bojanz: Let's go with "hide_single_bundle", sounds better :)

src/Plugin/views/field/EntityBundle.php
```diff
@@ -0,0 +1,117 @@
...
+   *   The field formatter plugin manager.
+   * @param \Drupal\Core\Field\FieldTypePluginManagerInterface $field_type_plugin_manager
+   *   The field plugin type manager.
+   * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
+   *   The language manager.
+   * @param \Drupal\Core\Render\RendererInterface $renderer
+   *   The renderer.
+   *  @param \Drupal\Core\Entity\EntityTypeBundleInfoInterface $entityTypeBundleInfo
+   *    The entity type bundle info.
+   *
```
bojanz: Extra new line.

src/Plugin/views/field/EntityBundle.php
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+namespace Drupal\commerce\Plugin\views\field;
+
+use Drupal\views\Plugin\views\field\Field;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Field handler to display the entity bundle. It displays nothing if there's
```
bojanz: Docblocks like this should have a short single line description, then an empty line, then the full description.

"Displays the entity bundle.

Can be configured to show nothing when there's only one possible bundle, allowing the 'Hide empty column' table setting to hide the column."

commerce.views.inc
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Implements hook_views_data_alter().
+ */
+function commerce_views_data_alter(array &$data) {
+  // Override the default bundle views handler for all content entity types.
+  // Get all entity types.
+  $entity_types = \Drupal::service('entity_type.manager')->getDefinitions();
+  foreach ($entity_types as $entity_type) {
+    // We only want content entities.
```
bojanz: Careful, we should limit ourselves to commerce content entities only (id starts with "commerce_").

commerce.views.inc
```diff
@@ -0,0 +1,16 @@
+<?php
+
+/**
+ * Implements hook_views_data_alter().
+ */
+function commerce_views_data_alter(array &$data) {
+  // Override the default bundle views handler for all content entity types.
+  // Get all entity types.
```
bojanz: "// Get all entity types." is redundant.

bojanz:
Almost ready! You'll also need to re-export the relevant views.

czigor:
Thanks for the review! Fixed everything. The views export at /admin/config/development/configuration/single/export did not work flawlessly (it was trying to set the checkbox value to '1' instead of 'true' and other issues) so I did it manually. 

Note that I have found the "Hide empty column" setting itself to be buggy as well, see https://www.drupal.org/node/2621874.

czigor:
Tests are failing because the config schema is not right. This is also confirmed by the https://www.drupal.org/project/config_inspector module. I still have not figured out how to fix this.

czigor:
Tests fail in 8.x-2.x as well.

bojanz:
Can you rebase this, please?

czigor:
Rebased, seems to work.

commerce.views.inc
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Implements hook_views_data_alter().
+ */
+function commerce_views_data_alter(array &$data) {
+  // Override the default bundle views field handler.
+  $entity_types = \Drupal::service('entity_type.manager')->getDefinitions();
+  foreach ($entity_types as $entity_type) {
+    // Only override for commerce content entities.
+    if ($entity_type instanceof \Drupal\Core\Entity\ContentEntityType && strpos($entity_type->id(), 'commerce_') === 0) {
+      $data[$entity_type->getDataTable()][$entity_type->getKey('bundle')]['field']['id'] = 'entity_bundle';
```
bojanz: $entity_type->getDataTable() is empty for commerce_order. We need additional logic (use the base table in that case).

czigor: Fixed it.

bojanz:
Committed in 144f4d6b8b29f5e31037ca3105f67853a5ed814e, with some tweaks to commerce.views.inc

PR #273, opened 2015-11-20
Issue #1385826 by bgilhome: Separate delete permissions out of edit permissions

mglaman:
…ermissions


mglaman:
This needs work. Need to implement access control handler, like NodeAccessControlHandler.

bojanz:
We'll be using the new Entity API permissions system.

PR #274, opened 2015-11-23
Fix broken product type handler in products view

czigor:


czigor:
https://www.drupal.org/node/2621428

PR #275, opened 2015-11-24
Issue #2610220: Create a views argument plugin for filtering by active store

jazzdrive3:
This also adds a StoreCacheContext, so we can use is in the argument definition.

### DIFF
--- a/modules/store/src/Cache/Context/StoreCacheContext.php
+++ b/modules/store/src/Cache/Context/StoreCacheContext.php
@@ -1,0 +1,59 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_store\Cache\Context\StoreCacheContext.
+ */
+
+namespace Drupal\commerce_store\Cache\Context;
+
+use Drupal\commerce_store\StoreContextInterface;
+use Drupal\Core\Cache\Context\CacheContextInterface;
+use Drupal\Core\Cache\CacheableMetadata;
+
+/**
+ * Defines the StoreCacheContext service, for "per store" caching.
+ *
+ * Cache context ID: 'store'.
+ */
+class StoreCacheContext implements CacheContextInterface {
+
+  /**
+   * The store context.
+   *
+   * @var \Drupal\commerce_store\StoreContextInterface
+   */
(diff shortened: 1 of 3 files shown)


jazzdrive3:
What is the process for reviewing these?

modules/store/src/Cache/Context/StoreCacheContext.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_store\Cache\Context\StoreCacheContext.
+ */
```
borisson, line 6: The `@file` docblocks for files that have only one class/interface can be removed, so let's remove this.

modules/store/src/Cache/Context/StoreCacheContext.php
```diff
@@ -0,0 +1,58 @@
...
+   *
+   * @var \Drupal\commerce_store\StoreContextInterface
+   */
+  protected $storeContext;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getLabel() {
+    return t('Store');
```
borisson, line 42: Let's use the `StringTranslationTrait` and use `$this->t(` here instead if that's possible.

modules/store/src/Plugin/views/argument_default/ActiveStore.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_store\Plugin\views\argument_default\ActiveStore.
+ */
```
borisson, line 6: The @file docblocks for files that have only one class/interface can be removed, so let's remove this.

borisson:
I think this is a good addition, but it'll probably need a rebase and you should also squash the commits after resolving the nitpicks.

modules/store/src/Cache/Context/StoreCacheContext.php
mglaman: It's a static method

bojanz:
Continued in https://www.drupal.org/node/2610220.

PR #281, opened 2015-12-12
Move the store selector to the right sidebar on the product edit form

JulienD:
I've moved the store field widget to the right bar and added control to display or not the fieldset by checking the visibility of the store field widget. https://www.drupal.org/node/2626972#comment-10662764


bojanz:
Posting an updated patch to https://www.drupal.org/node/2626972 along with screenshots for other reviewers. Thanks, Julien.

PR #282, opened 2015-12-12
Issue #2633906: Integrate with Entity API

mglaman:


modules/order/src/Entity/LineItem.php
```diff
@@ -148,23 +149,14 @@ public function setData($data) {
    * {@inheritdoc}
    */
   public static function baseFieldDefinitions(EntityTypeInterface $entityType) {
+    $fields = self::entityKeysBaseFieldDefinitions($entityType);
+
     $fields['line_item_id'] = BaseFieldDefinition::create('integer')
```
bojanz: Now you have two ids, line_item_id and id :)

PR #284, opened 2015-12-12
The store email isn't validated

mglaman:


mglaman:
@bojanz I'm ¯_(ツ)_/¯. Apparently, it seems, once we let the Field API touch the value constraints are useless on a data definition. It only validates programmatically.

PR #285, opened 2015-12-12
Add test class for currency

petebarnett:


bojanz:
Continued in https://www.drupal.org/node/2615398

PR #286, opened 2015-12-12
2612572 bulk ops

fgm:
Refactored routes and actions using a deriver. "Patch 6" is the one from d.o.


vasike:
it seems there is an issue with the form id and there is no deletion (confirm form) for VBO actions.
there is patch that could help for this PR.
The patch contains also some small code "corrections".
[pr_286-2612572_bulk_ops_fixes.txt](https://github.com/commerceguys/commerce/files/77332/pr_286-2612572_bulk_ops_fixes.txt)

bojanz:
Continued in #334

PR #288, opened 2015-12-23
Export entity form and view display for billing profile bundle

mglaman:
…le bundle


bojanz:
Added the config_devel entry and committed: c6c068ffbdd5bdf0cd2cd884fd76a1b89ef145c7

PR #289, opened 2015-12-25
Issue #2640724: Use access handler to prevent edit field access to st…

mglaman:
…ore if entity has been saved

### DIFF
--- a/modules/order/src/Entity/Order.php
+++ b/modules/order/src/Entity/Order.php
@@ -24,6 +24,7 @@ use Drupal\profile\Entity\ProfileInterface;
  *   id = "commerce_order",
  *   label = @Translation("Order"),
  *   handlers = {
+ *     "access" = "Drupal\commerce_order\OrderAccessControlHandler",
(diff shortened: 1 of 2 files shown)


bojanz:
We should remove the store widget completely, it has no business on the edit page.
Same for uid / email.

mglaman:
Should it just be viewable on the edit, for reference? Make a meta info pane like node and seven?

Should uid, email be added to access handler and then remove hide from form display?

mglaman:
Also.. I think email has a place for edit. What if its an admin created order and the notifications email needs to be updated. I'd want to see what others do for email

bojanz:
It makes sense to show the user / mail in the right sidebar. I guess the store as well? We need to discuss what goes there.

As for the second question, I envisioned a "Reassign order" action that has the same Existing user | New user UI as the add form. So it's editable, just outside of the actual edit screen.

mglaman:
I like that idea. More buttons and specific actions than just fields.

mglaman:
We're going to just manage the form, not access handler.

PR #290, opened 2015-12-25
Issue #2636550: Update the order edit form

mglaman:


mglaman:
Related to #289

modules/order/commerce_order.module
```diff
@@ -10,6 +10,17 @@ use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
 
 /**
+ * Implements hook_theme().
+ */
+function commerce_order_theme($existing, $type, $theme, $path) {
+  return [
+    'order_edit_form' =>  [
```
bojanz: Should this be commerce_order_edit_form? Are we supposed to namespace the theme functions like that? I know I didn't for the product form, just wondering if that makes sense.

modules/order/src/Form/OrderForm.php
```diff
@@ -32,26 +38,44 @@ public function form(array $form, FormStateInterface $form_state) {
...
+    else {
+      // @todo: This should be hidden if only one store.
+      $form['order_state']['store'] = $this->fieldAsReadOnly(t('Store'), $order->getStore()->label());
+    }
+
     if (isset($form['state'])) {
       $form['state']['#group'] = 'order_state';
     }
+    else {
+      // @todo: We need an action button for working with states and transitions.
```
bojanz: The action button won't go in the edit form (view instead?), so we can kill the @todo for now.

modules/order/src/Form/OrderForm.php
```diff
@@ -32,26 +38,44 @@ public function form(array $form, FormStateInterface $form_state) {
 
     $form['order_state'] = [
```
bojanz: Why are we keeping "order_state" as the key? Doesn't match the "Order information" label.

modules/order/commerce_order.module
mglaman: I'll do it anyways. I'm honestly not sure. But I remember someone opened a ticket about Layout plugin module not doing the theme namespacing.

bojanz:
Committed in 46f7970d179e94ee77aec64137ad281967b66907.

PR #291, opened 2015-12-28
Issue #2641490: Create a "placed" timestamp to save when order entere…

mglaman:
…d validation state


mglaman:
Related to a todo in #290

bojanz:
Committed in b0e908235b6143fcbcc70c5e1817914fce669863.

I've made placed without a widget for now, and will extend the OrderAddForm to include it. That seems like the best place to cover the use case of "I am creating an order for a past date".

PR #292, opened 2015-12-28
Issue #2641492: Create an action to reassign an order

mglaman:


mglaman:
Updated to add trait for the user selection elements.

modules/order/src/Form/OrderReassignForm.php
```diff
@@ -0,0 +1,89 @@
+<?php
+/**
```
bojanz, line 3: That's not a cool docblock :)

modules/order/src/OrderUserSelectTrait.php
```diff
@@ -0,0 +1,110 @@
...
+      ];
+    }
+  }
+
+  /**
+   * Rebuild form for customer user type selection.
+   */
+  public function validateCustomerUserType(array $form, FormStateInterface $form_state) {
+    return $form['customer'];
+  }
```
bojanz: You need a newline before the closing class brace.

modules/order/src/OrderUserSelectTrait.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+/**
+ * @file Contains \Drupal\commerce_order\OrderUserSelectTrait.
```
bojanz: This docblock is not standard either (missing newline after @file)

modules/order/src/Entity/Order.php
```diff
@@ -51,6 +51,7 @@
  *     "canonical" = "/admin/commerce/orders/{commerce_order}",
  *     "edit-form" = "/admin/commerce/orders/{commerce_order}/edit",
  *     "delete-form" = "/admin/commerce/orders/{commerce_order}/delete",
+ *     "reassign-form" = "/admin/commerce/orders/{commerce_order}/reassign",
```
bojanz: You don't need to define the link template, it won't be used by anything (they're usually used by route providers).

mglaman: I used it for

````
$reassign_route_info = $order->toUrl('reassign-form');
````

Also wasn't sure if we'd add this to operations dropbutton, which simplifies referencing the route then

Comment (bojanz) [inline modules/order/src/Entity/Order.php]: Okay, nevermind then.

File: modules/order/src/OrderListBuilder.php [hunk 5]
```diff
@@ -110,4 +110,22 @@ public function buildRow(EntityInterface $entity) {
     return $row + parent::buildRow($entity);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  protected function getDefaultOperations(EntityInterface $entity) {
+    $operations = parent::getDefaultOperations($entity);
+
+    if ($entity->access('update') && $entity->hasLinkTemplate('reassign-form')) {
```
mglaman, line 119: I left the "hasLinkTemplate" in case someone overrides entity class, or removes this action in favor of something else. Then they do not have to override this as well (if they did not want to.)

bojanz:
Committed as 53db1e28df667d49a430c960be857661348fb39b.

Some of the changes made:
1) Moved the trait from src/ to src/Form, so that it lives next to the classes that use it.
2) Renamed the trait to CustomerFormTrait, felt it's more clear.
3) Moved the related submit code to the trait.
4) Added dependency injection where missing.
5) Added a redirect to the reassign form. 
6) Changed the weight of the reassign action so that it appears below Edit.

PR #296, opened 2016-01-05
Replaced fulfilment by fulfillment

pcambra:
Drupal is normally in American English, so we should be consistent and use fulfillment, even if it looks bad :)


bojanz:
I tell you old chap, we never should have given the colonies independence.

Committed in 395bc60c37800d78ee585824e0548d5b7838660d :)

PR #298, opened 2016-01-05
Issue #2644790: All order base fields that are viewable on "view" dis…

mglaman:
…play must be configurable


mglaman:
Provides the following. I made some adjustments previous which is why store is hidden. But this way we can control it much more easily for user view.
screen shot 2016-01-05 at 3 56 11 pm

PR #299, opened 2016-01-07
Add missing dependency on entity module for commerce_store.

DuaelFr:
Causes this kind of error after having enabled commerce_store.
PHP Fatal error:  Interface 'Drupal\entity\Entity\EntityDescriptionInterface' not found in /web/modules/contrib/commerce/modules/store/src/Entity/StoreTypeInterface.php on line 16

### DIFF
--- a/modules/store/commerce_store.info.yml
+++ b/modules/store/commerce_store.info.yml
@@ -7,6 +7,7 @@ dependencies:
   - address
   - commerce
   - commerce_price
+  - entity
   - options
 config_devel:
   - commerce_store.commerce_store_type.default


bojanz:
commerce_store depends on commerce, which depends on entity.
Thus, commerce_store already depends on entity.

Sounds like we might be running into a core bug? Can you provide a set of steps you took?

DuaelFr:
I just took the repository we created during the sprint then git pulled commerce to get the last version. After that I ran `drush si standard && drush en -y commerce commerce_cart commerce_order commerce_price commerce_product commerce_store commerce_tax`. I wasn't asked to download or enable entity.

I tried on a brand new install with a freshly downloaded drupal and commerce and I cannot reproduce so I suppose we can close this PR :/

PR #302, opened 2016-01-12
Issue #2646364: Create the line item reference view formatter

mglaman:


modules/order/src/Plugin/Field/FieldFormatter/LineItemTable.php
```diff
@@ -0,0 +1,47 @@
...
+      '#arguments' => [$order->id()],
+      '#embed' => TRUE,
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function isApplicable(FieldDefinitionInterface $field_definition) {
+    return $field_definition->getTargetEntityTypeId() == 'commerce_order';
```
bojanz: We should restrict further by field name, otherwise it will show up as an option on other entity references, like billing profile.

modules/order/src/Plugin/Field/FieldFormatter/LineItemTable.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\commerce_order\Plugin\Field\FieldFormatter.
```
bojanz: Missing class name at the end of the namespace.

modules/order/config/install/views.view.line_item_summary.yml
```diff
@@ -0,0 +1,397 @@
...
+dependencies:
+  module:
+    - commerce_order
+    - commerce_price
+id: line_item_summary
+label: 'Line item summary'
+module: views
+description: ''
+tag: ''
+base_table: commerce_line_item
```
bojanz: Doesn't line_item_table make more sense, cause it mirrors the formatter name (and 1.x)?

I was initially surprised that this is not an order view, but I see 1.x did the same. I guess that means the order total is not desired in the footer?

bojanz:
Committed in 0fe712371a4fc40c6e63ee2cde443d354bf913fc.

Renamed the formatter from line_item_table to commerce_line_item_table, added @todo for making the selected view configurable, cleaned up isApplicable().

PR #306, opened 2016-01-18
Namespace the plugins (prepend 'commerce_').

bojanz:
Some notes:
- Decided to do commerce_entity_bundle instead of commerce_bundle because core mostly adds entity_ to entity specific handlers
- Removed the _action suffix from the product actions, core does it everywhere for actions, but no other plugin type does that.
- I'm unhappy about commerce_edit_quantity and commerce_edit_remove, but no better ideas for now (edit_quantity and edit_remove are names taken from 1.x)


bojanz:
Committed in af9f2dbacc2f6670d84fcb80cfc03bc59e7cf71f with the quantity/remove_button renames we discussed internally, and a CSS class update.

PR #308, opened 2016-01-21
Port display_submitted to product types with product theme/template implementation

vasike:
Issue #2652704 by vasike, bojanz

### DIFF
--- a/commerce.module
+++ b/commerce.module
@@ -57,3 +57,22 @@ function commerce_field_widget_form_alter(&$element, FormStateInterface $form_st
     $element[$child]['#description'] = '';
   }
 }
+
+/**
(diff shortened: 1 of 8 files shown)


bojanz:
It feels like the template & related code that we pulled from node has a ton of legacy cruft, that makes the whole idea of controlling "submitted by" seem less attractive. I'm going to gather some feedback from the community.

bojanz:
Talked to mglaman and berdir. Matt agreed that showing the author information isn't needed by default on a product. Both him and berdir suggested this to be done via view modes and formatters instead.

So the only thing that we need from this PR at this point is the suggestions hook, but we can add that in the "show product rendered in the cart" issue.

Thank you for exploring this Tavi.

PR #312, opened 2016-01-21
Issue #2646196 by mglaman: Create the order view route and mode for u…

mglaman:
…sers


mglaman:
Didn't make test yet for same reason user order view doesn't have tests: permissions

bojanz:
Committed.

PR #315, opened 2016-01-27
Issue #2472263: Add cart icon

joshmiller83:
Per request by @mglaman here is the cart icon sans Druplicon.

https://www.drupal.org/node/2472263


joshmiller83:
Also posted as a patch here (use whichever you want): https://www.drupal.org/node/2472263#comment-10791812

PR #320, opened 2016-02-02
Issue #2464371: Create an add to cart form/block

mglaman:
Adds attribute selection and AJAXy goodness. Mostly a copy and paste from 1.x, however concept of "Matching products" has been removed since we're not supporting a dropdown for attribute-less variations.

### DIFF
--- a/modules/product/src/Form/AddToCartForm.php
+++ b/modules/product/src/Form/AddToCartForm.php
@@ -9,11 +9,16 @@ namespace Drupal\commerce_product\Form;
 
 use Drupal\commerce_cart\CartManagerInterface;
 use Drupal\commerce_cart\CartProviderInterface;
+use Drupal\commerce_product\Entity\ProductInterface;
+use Drupal\commerce_product\Entity\ProductVariationInterface;
 use Drupal\commerce_store\StoreContextInterface;
+use Drupal\Core\Entity\EntityFieldManagerInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Link;
+use Drupal\Core\Render\Element;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
@@ -50,6 +55,18 @@ class AddToCartForm extends FormBase {
   protected $storeContext;
 
   /**
+   * The entity field manager.
+   *
+   * @var \Drupal\Core\Entity\EntityFieldManager
+   */
+  protected $entityFieldManager;
+
+  /**
+   * @var \Drupal\commerce_product\Entity\ProductVariationInterface
+   */
+  protected $defaultVariation;
+
+  /**
    * Constructs a new AddToCartForm object.
    *
    * @param \Drupal\commerce_cart\CartManagerInterface $cart_manager
@@ -57,15 +74,18 @@ class AddToCartForm extends FormBase {
    * @param \Drupal\commerce_cart\CartProviderInterface $cart_provider
    *   The cart provider.
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
-  49   The entity type manager.
+   *   The entity type manager.
    * @param \Drupal\commerce_store\StoreContextInterface $store_context
    *   The store context.
+   * @param \Drupal\Core\Entity\EntityFieldManager $entity_field_manager
+   *   The entity field manager.
    */
-  public function __construct(CartManagerInterface $cart_manager, CartProviderInterface $cart_provider, EntityTypeManagerInterface $entity_type_manager, StoreContextInterface $store_context) {
+  public function __construct(CartManagerInterface $cart_manager, CartProviderInterface $cart_provider, EntityTypeManagerInterface $entity_type_manager, StoreContextInterface $store_context, EntityFieldManagerInterface $entity_field_manager) {
     $this->cartManager = $cart_manager;
     $this->cartProvider = $cart_provider;
     $this->variationStorage = $entity_type_manager->getStorage('commerce_product_variation');
     $this->storeContext = $store_context;
+    $this->entityFieldManager = $entity_field_manager;
   }
 
   /**
@@ -76,7 +96,8 @@ class AddToCartForm extends FormBase {
       $container->get('commerce_cart.cart_manager'),
       $container->get('commerce_cart.cart_provider'),
       $container->get('entity_type.manager'),
-      $container->get('commerce_store.store_context')
+      $container->get('commerce_store.store_context'),
+      $container->get('entity_field.manager')
     );
   }
 
@@ -91,28 +112,23 @@ class AddToCartForm extends FormBase {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state, $product = NULL, array $settings = NULL) {
-    $variations = [];
-    foreach ($product->variations->referencedEntities() as $variation) {
-      $variations[$variation->id()] = $variation->label();
-    }
-
+    /** @var \Drupal\commerce_product\Entity\Product $product */
     $form['#settings'] = $settings;
+    $variations = $product->variations->referencedEntities();
+
+    // Set the first variation as the default to be used.
+    $this->defaultVariation = reset($variations);
+    $form_state->set('default_variation', $this->defaultVariation->id());
 
     // If there are multiple variations, display ability to select a variation.
     if (count($variations) > 1) {
-      $form['variation'] = [
-        '#type' => 'select',
-        '#title' => $this->t('Select variation:'),
-        '#options' => $variations,
-      ];
+      $this->variationForm($form, $form_state, $variations);
     }
-    // Otherwise set the variation and only display "add to cart".
-    else {
-      $form['variation'] = [
-        '#type' => 'value',
-        '#value' => key($variations),
-      ];
-    }
+
+    $form['variation'] = [
+      '#type' => 'value',
+      '#value' => $form_state->get('default_variation'),
+    ];
 
     if (!empty($settings['show_quantity'])) {
       $form['quantity'] = [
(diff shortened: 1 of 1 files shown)


modules/product/src/Form/AddToCartForm.php
```diff
@@ -57,15 +68,19 @@ class AddToCartForm extends FormBase {
...
+   *   The entity field manager.
    */
-  public function __construct(CartManagerInterface $cart_manager, CartProviderInterface $cart_provider, EntityTypeManagerInterface $entity_type_manager, StoreContextInterface $store_context) {
+  public function __construct(CartManagerInterface $cart_manager, CartProviderInterface $cart_provider, EntityTypeManagerInterface $entity_type_manager, StoreContextInterface $store_context, EntityFieldManagerInterface $entity_field_manager) {
     $this->cartManager = $cart_manager;
     $this->cartProvider = $cart_provider;
     $this->variationStorage = $entity_type_manager->getStorage('commerce_product_variation');
     $this->storeContext = $store_context;
+    $this->entityFieldManager = $entity_field_manager;
+
```
bojanz: Extra new line.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -91,29 +107,20 @@ public function getFormId() {
...
    */
   public function buildForm(array $form, FormStateInterface $form_state, $product = NULL, array $settings = NULL) {
-    $variations = [];
-    foreach ($product->variations->referencedEntities() as $variation) {
-      $variations[$variation->id()] = $variation->label();
-    }
-
     $form['#settings'] = $settings;
 
+    $form_state->set('default_variation', $product->variations->first()->entity);
```
bojanz: Can this store the variation id as well? We want to keep the form and form state as slim as possible.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +173,224 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+   * @param \Drupal\commerce_product\Entity\ProductInterface $product
+   *   The product that has the variations.
+   */
+  protected function variationForm(array &$form, FormStateInterface $form_state, ProductInterface $product) {
+    $default_variation = $form_state->get('default_variation');
+    $attribute_names = [];
+    $unchanged_attributes = [];
+    $user_input = $form_state->getUserInput();
+
+    $variations = $product->variations->referencedEntities();
```
bojanz: use $product->getVariations()

mglaman: This doesn't exist? Only in ProductVariationManager? So use the service instead of the field directly?

modules/product/src/Form/AddToCartForm.php
```diff
@@ -50,6 +55,18 @@ class AddToCartForm extends FormBase {
   protected $storeContext;
 
   /**
+   * The entity field manager.
+   *
+   * @var \Drupal\Core\Entity\EntityFieldManager
+   */
+  protected $entityFieldManager;
```
bojanz, line 62: The code ended up not using $entityFieldManager, so this needs to be removed (along with unused use statements, constructor changes).

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+   * @param $name
+   *   The attribute field name.
+   * @param $value
+   *   The attribute value.
+   *
+   * @return bool
+   *   Return TRUE on value match.
+   */
+  protected function isMatchingVariation(ProductVariationInterface $variation, $name, $value) {
+    return $variation->{$name}->first()->entity->id() == $value;
```
bojanz, line 305: $variation->{$name}->first()->entity->id() can be replaced with $variation->get($name)->target_id everywhere.

We'll want to introduce $variation->getAttributeValueId($name) in the future, but the target_id way is short enough not to need a helper function such as isMatchingVariation.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+   *   The product variations.
+   *
+   * @return array
+   *   Information about the attributes.
+   */
+  protected function getAttributesInfo(array $variations) {
+    $attributes = [];
+    /** @var \Drupal\Core\Field\FieldConfigInterface $field */
+    foreach ($this->defaultVariation->getAttributeFieldDefinitions() as $field) {
+      $third_party_settings = $field->getThirdPartySettings('commerce_product');
```
bojanz, line 269: This method calls $field->getName() many times, would be cleaner to just declare $field_name before $third_party_settings and use it.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+        'name' => $field->getName(),
+        'type' => $third_party_settings['attribute_widget'],
+        'title' => $attribute_label,
+        'required' => $field->isRequired(),
+        'options' => [],
+      ];
+
+      // Get possible field options based on current variations.
+      foreach ($variations as $variation) {
+        $attribute_value = $variation->{$field->getName()}->first()->entity;
```
bojanz, line 284: Just like in the target_id example, the first() here isn't needed, TypedData does that for you.

modules/product/src/Form/AddToCartForm.php
mglaman, line 305: Beautiful. I thought there had to be an easier way. That should improve performance as well, since we are accessing the entity.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+  /**
+   * Gets an array of attributes and their settings.
+   *
+   * @param \Drupal\commerce_product\Entity\ProductVariationInterface[] $variations
+   *   The product variations.
+   *
+   * @return array
+   *   Information about the attributes.
+   */
+  protected function getAttributesInfo(array $variations) {
```
bojanz, line 265: getAttributeInfo() is nicer

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+   * Builds the variation selection form.
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   * @param \Drupal\commerce_product\Entity\ProductVariationInterface[] $variations
+   *   An array of product variations.
+   */
+  protected function variationForm(array &$form, FormStateInterface $form_state, array $variations) {
```
bojanz, line 195: The body of this method belongs in buildForm(), it's essential to the form structure so we gain nothing by splitting it out.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+    // Resolve the default variation from current attribute values.
+    $this->resolveDefaultVariation($variations, $form_state);
+
+    // Get qualified attribute options and add them to the form.
+    foreach ($this->resolveQualifyingAttributes($variations) as $name => $data) {
+      $this->attributeElement($form, $data, $data['used_options']);
+    }
+
+    if (!empty($form['attributes'])) {
+      $form['attributes'] += [
```
bojanz, line 205: This should be defined before the foreach that calls $this->attributeElement(), since that method fills up the children of those keys.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+   * @param \Drupal\commerce_product\Entity\ProductVariationInterface[] $variations
+   *   An array of product variations.
+   */
+  protected function variationForm(array &$form, FormStateInterface $form_state, array $variations) {
+    // Resolve the default variation from current attribute values.
+    $this->resolveDefaultVariation($variations, $form_state);
+
+    // Get qualified attribute options and add them to the form.
+    foreach ($this->resolveQualifyingAttributes($variations) as $name => $data) {
+      $this->attributeElement($form, $data, $data['used_options']);
```
bojanz, line 201: If we're already passing $data, then there's no need to pass $data['used_options'] separately.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param array $attribute
+   *   The attribute information array.
+   * @param $used_options
+   *   An array of options in use.
+   */
+  protected function attributeElement(array &$form, array $attribute, array $used_options) {
+    $field_name = $attribute['name'];
+    $options = array_intersect_key($attribute['options'], array_combine($used_options[$field_name], $used_options[$field_name]));
```
bojanz, line 229: The parent method passes $attribute['options'] as $used_options, so this array_intersect_key does nothing.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -166,4 +182,253 @@ function submitForm(array &$form, FormStateInterface $form_state) {
...
+   * @return array
+   *   Information about the attributes.
+   */
+  protected function getAttributesInfo(array $variations) {
+    $attributes = [];
+    /** @var \Drupal\Core\Field\FieldConfigInterface $field */
+    foreach ($this->defaultVariation->getAttributeFieldDefinitions() as $field) {
+      $third_party_settings = $field->getThirdPartySettings('commerce_product');
+
+      $attribute_label = !empty($third_party_settings['attribute_widget_title']) ?
```
bojanz, line 271: A multiline ternary has no point, cleaner to go with an if / else pair in that case.

bojanz:
Spent a ton of time trying to simplify this, because the existing logic took sick-me 5 reads to understand.
Here's the end result: 91bd75dd87394eee464198055d95e9f69d6d403c
It isolates the variation part of the form so that it can be easily moved into a widget once we turn the entire form into a line item creation form.

Added logic:

``` php
      // 1 required value -> Disable the element to skip unneeded ajax calls.
      if ($attribute['required'] && count($attribute['values']) === 1) {
        $form['attributes'][$field_name]['#disabled'] = TRUE;
      }
```

Logic that's still missing: clearing the non-triggering attribute values on ajax refresh. I ported the much better logic from Address module, but couldn't get it to work. It's late, I'll revisit this from tomorrow.

PR #321, opened 2016-02-04
Fix and improve product variation service

mglaman:
ProductVariationManagerInterface had a bad namespace, the service was broken. Also added QueryFactory as an injecting to get entity query.


bojanz:
We'll want to move this to a storage, but let's fix it first. Thanks.

PR #322, opened 2016-02-04
Issue #2660274: Disable the variation type field when editing product…

mglaman:
… types

### DIFF
--- a/modules/product/src/Form/ProductTypeForm.php
+++ b/modules/product/src/Form/ProductTypeForm.php
@@ -100,6 +100,7 @@ class ProductTypeForm extends BundleEntityFormBase {
       '#default_value' => $product_type->getVariationType(),
       '#options' => $variation_types,
       '#required' => TRUE,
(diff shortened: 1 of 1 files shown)


modules/product/src/Form/ProductTypeForm.php
```diff
@@ -100,6 +100,8 @@ public function form(array $form, FormStateInterface $form_state) {
       '#default_value' => $product_type->getVariationType(),
       '#options' => $variation_types,
       '#required' => TRUE,
+      '#description' => $this->t('This value cannot be changed once you have saved this product type.'),
```
bojanz: We don't provide such a description for other similar cases (such as machine names), so this might be unneeded?

mglaman: Sounds fair, I'll remove it

bojanz:
Committed the patch: d1c8bc1a83eedd4e057dc49edb2a0a4cf802d468

PR #323, opened 2016-02-08
2659080: Make cart scrollable if greater then window + Better positio…

haringsrob:
This improvement has the following 2 effects.
1. When the cart is going out of the window, it will change the positioning.
2. When the link inside the cart is clicked, it will go to the cart.

### DIFF
--- a/modules/cart/css/commerce_cart.layout.css
+++ b/modules/cart/css/commerce_cart.layout.css
@@ -9,13 +9,12 @@
 .cart-block--contents {
   display: none;
   position: absolute;
-  overflow: hidden;
   z-index: 300;
-}
-.cart-block--contents.is-scrollable {
+  overflow-x: hidden;
   overflow-y: scroll;
+  max-height: 300px;
 }
-.cart-block--contents.is-outside {
+.cart-block--contents.is-outside-horizontal {
   right: 0;
 }
 .cart-block--contents__expanded {
--- a/modules/cart/css/commerce_cart.layout.css
+++ b/modules/cart/css/commerce_cart.layout.css
@@ -12,6 +12,13 @@
   overflow: hidden;
   z-index: 300;
 }
+.cart-block--contents.is-scrollable {
+  overflow-y: scroll;
+  max-height: 200px;
+}
+.cart-block--contents.is-outside {
+  right: 0;
+}
 .cart-block--contents__expanded {
   overflow: visible;
 }
--- a/modules/cart/css/commerce_cart.layout.css
+++ b/modules/cart/css/commerce_cart.layout.css
@@ -10,6 +10,9 @@
   display: none;
   position: absolute;
   z-index: 300;
+  overflow: hidden;
+}
+.cart-block--contents__items {
   overflow-x: hidden;
   overflow-y: scroll;
   max-height: 300px;
--- a/modules/cart/css/commerce_cart.layout.css
+++ b/modules/cart/css/commerce_cart.layout.css
@@ -14,7 +14,6 @@
 }
 .cart-block--contents.is-scrollable {
   overflow-y: scroll;
-  max-height: 200px;
 }
 .cart-block--contents.is-outside {
   right: 0;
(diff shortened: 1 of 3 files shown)


mglaman:
I feel like there is a way we can simplify this, for accounting page resize. I also like accounting for viewport space to the right. Going to sit and think on it. 

I wonder if we can leverage Backbone and make the cart block an object, where resizing logic is a method on the object itself. Basically, how can we leverage what core does with toolbar here?

haringsrob:
Hi Mglaman,

Thank you for your feedback. Unfortunately, I have not the best JS (let alone frameworks) experience.

But if you could give me a starting point or example, then I'm willing to look into it.

Regards,

modules/cart/js/commerce_cart.js
```diff
@@ -8,19 +8,30 @@
 
   Drupal.behaviors.commerceCartBlock = {
     attach: function (context) {
-      var $context = $(context);
-      var $cart = $context.find('.cart--cart-block');
-      var $cartContents = $cart.find('.cart-block--contents');
+      var $context = $(context),
+          $cart = $context.find('.cart--cart-block'),
```
mglaman, line 12: We don't need this variable anymore, technically, correct?

modules/cart/js/commerce_cart.js
```diff
@@ -8,19 +8,30 @@
...
+          $cartContents = $cart.find('.cart-block--contents');
 
       if ($cartContents.length > 0) {
-        $cart.click(function (e) {
+        // Expand the block when the link is clicked.
+        $cartButton.on('click', function (e) {
+          // Prevent it from going to the cart.
           e.preventDefault();
+          // Get the shopping cart width + the offset to the left.
+          var $windowWidth = $(window).width(),
```
mglaman, line 22: As a general nitpick, I only place `$` on variables if they are a jQuery object. The variables defined in this section are all integers, so we can just use `windowWidth`, `cartOffsetLeft`, `cartWidth` respectively.

modules/cart/js/commerce_cart.js
haringsrob, line 12: You are talking about $cart?

We use it currently in 2 places. We could change it, but that would things more complicated I guess.

modules/cart/js/commerce_cart.js
haringsrob, line 22: Changed this. Thanx.

modules/cart/js/commerce_cart.js
mglaman, line 12: If it's used then we can keep it :)

modules/cart/css/commerce_cart.layout.css
```diff
@@ -9,8 +9,16 @@
 .cart-block--contents {
   display: none;
   position: absolute;
-  overflow: hidden;
```
mglaman, line 12: Flipflopping this back, reduce diff and unneeded changed.

modules/cart/js/commerce_cart.js
mglaman, line 22: Changed `$windowWidth` and `$cartOffsetLeft` to not have `$`, they're integers.

modules/cart/js/commerce_cart.js
```diff
@@ -8,19 +8,30 @@
...
+            $cartOffsetLeft = $cart.offset().left,
```
mglaman, line 23: Since we only use `$cartOffsetLeft` once, I just put that into the next line

modules/cart/js/commerce_cart.js
mglaman, line 22: Drupal ESLint triggered that there should be one `var` for each variable definition, not joined via `,`, just FYI :)

modules/cart/js/commerce_cart.js
```diff
@@ -8,19 +8,30 @@
 
   Drupal.behaviors.commerceCartBlock = {
     attach: function (context) {
-      var $context = $(context);
-      var $cart = $context.find('.cart--cart-block');
-      var $cartContents = $cart.find('.cart-block--contents');
+      var $context = $(context),
```
mglaman, line 11: Drupal ESLint specifies one `var` for each variable, no `,` for multiple

mglaman:
Committed as @62cf8d8201784f7b634cc0b705ff907022978646

PR #327, opened 2016-02-11
Fixed deprecated entity_create.

facine:


bojanz:
Thanks! Let's use $this->createEntity() like the rest of the test instead?

facine:
Fixed!

PR #329, opened 2016-02-11
Fixed deprecated entity_get_display.

facine:


mglaman:
Looks like this is being taken care of in #332, and has passing tests.

bojanz:
Yep, continued in #332

PR #331, opened 2016-02-11
2663652 Removed entity type from ::load

haringsrob:


bojanz:
Committed in dc8989d1197d60c787de42a44587ee3d95e4d5e4.

PR #332, opened 2016-02-15
2660280 product variation display

haringsrob:


bojanz:
I like the general approach of using a helper function. Moved it to commerce.module so that we can use it for commerce_order as well. Committed with cleanups in cdbab180d6e8685555215a02e2003548e3e0c5c6.

PR #335, opened 2016-02-17
2659096 improve interface tests

haringsrob:
Hi,

The full implementation is not complete yet, as I'm having some issues adding an address to an entity.

The only methods yet to be implemented are:

```
  public function getAddress();
  public function setAddress(AddressInterface $address);
  public function getBillingCountries();
  public function setBillingCountries(array $countries);
```

All other tests for Product, ProductVariation and Store are there.

I'm already opening this as we could already take advantage of the tests that are there.

I also included a small fix in product::getOwnerId.

Regards,


bojanz:
Rebased, finished, committed in 62d3f8095b3983736563ee9da32f664da18e5093

PR #336, opened 2016-02-18
Issue #2669120 by mglaman: Allow line item fields to be exposed on th…

mglaman:
…e add to cart form

### DIFF
--- a/modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.add_to_cart.yml
+++ b/modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.add_to_cart.yml
@@ -1,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - commerce_order.commerce_line_item_type.product_variation
+    - core.entity_form_mode.commerce_line_item.add_to_cart
+  module:
+    - commerce_product
+id: commerce_line_item.product_variation.add_to_cart
+targetEntityType: commerce_line_item
+bundle: product_variation
+mode: add_to_cart
+content:
+hidden:
+  created: true
+  purchased_entity: true
+  quantity: true
+  status: true
+  uid: true
+  unit_price: true
--- a/modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.default.yml
+++ b/modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.default.yml
@@ -4,7 +4,7 @@ dependencies:
   config:
     - commerce_order.commerce_line_item_type.product_variation
   module:
-    - commerce_price
+    - commerce_product
 id: commerce_line_item.product_variation.default
 targetEntityType: commerce_line_item
 bundle: product_variation
@@ -22,14 +22,10 @@ content:
     type: number
     weight: 1
     settings:
-      placeholder: ''
-    third_party_settings: {  }
-  unit_price:
-    type: commerce_price_default
-    weight: 2
-    settings: {  }
+      placeholder: '1'
     third_party_settings: {  }
 hidden:
   created: true
   status: true
   uid: true
+  unit_price: true
--- a/modules/product/src/Form/AddToCartForm.php
+++ b/modules/product/src/Form/AddToCartForm.php
@@ -240,6 +240,8 @@ class AddToCartForm extends FormBase {
             'callback' => '::ajaxRefresh',
             'wrapper' => $form['#wrapper_id'],
           ],
+          // Options may change and not contain the current #value when processed.
+          '#validated' => TRUE,
         ];
         // Convert the _none option into #empty_value.
         if (isset($form['attributes'][$field_name]['options']['_none'])) {
@@ -298,11 +300,9 @@ class AddToCartForm extends FormBase {
         }
       }
 
+      // Check if we have valid matches, and change the current variation.
       $matches = array_diff_key($matches, $invalid_matches);
-
-      if (!empty($matches)) {
-        $current_variation = reset($matches);
-      }
+      $current_variation = reset($matches);
     }
 
     return $current_variation;
--- a/modules/product/src/Form/AddToCartForm.php
+++ b/modules/product/src/Form/AddToCartForm.php
@@ -7,6 +7,7 @@ use Drupal\commerce_cart\CartProviderInterface;
 use Drupal\commerce_product\Entity\ProductVariationInterface;
 use Drupal\commerce_store\StoreContextInterface;
 use Drupal\Component\Utility\Html;
+use Drupal\Core\Entity\Entity\EntityFormDisplay;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
@@ -17,6 +18,8 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  * Provides the add to cart form for product variations.
  */
 class AddToCartForm extends FormBase {
+
+  const LINE_ITEM_FORM_DISPLAY = 'add_to_cart';
 
   /**
    * The cart manager.
@@ -45,13 +48,6 @@ class AddToCartForm extends FormBase {
    * @var \Drupal\commerce_store\StoreContextInterface
    */
   protected $storeContext;
-
-  /**
-   * The current variation.
-   *
-   * @var \Drupal\commerce_product\Entity\ProductVariationInterface
-   */
-  protected $currentVariation;
 
   /**
    * Constructs a new AddToCartForm object.
@@ -95,6 +91,8 @@ class AddToCartForm extends FormBase {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state, $product = NULL, array $settings = NULL) {
+    /** @var \Drupal\commerce_product\Entity\ProductInterface $product */
+
     $wrapper_id = Html::getUniqueId('commerce-product-add-to-cart-form');
     $form += [
       '#tree' => TRUE,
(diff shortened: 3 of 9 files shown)


modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.add_to_cart.yml
```diff
@@ -0,0 +1,26 @@
+uuid: 4d6eaaee-5d0d-41e5-b100-da5451fad534
```
bojanz: We don't commit the uuid in default config.

modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.add_to_cart.yml
```diff
@@ -0,0 +1,26 @@
...
+    - commerce_order.commerce_line_item_type.product_variation
+    - core.entity_form_mode.commerce_line_item.add_to_cart
+  module:
+    - commerce_cart
+id: commerce_line_item.product_variation.add_to_cart
+targetEntityType: commerce_line_item
+bundle: product_variation
+mode: add_to_cart
+content:
+  quantity:
```
bojanz: Shouldn't we hide the quantity by default? (It used to be disabled by default before)

modules/product/src/Form/AddToCartForm.php
```diff
@@ -150,9 +149,29 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
-    $quantity = $form_state->getValue('quantity');
+
     $combine = $form['#settings']['combine'];
-    $this->cartManager->addEntity($cart, $variation, $quantity, $combine);
+
+    /** @var \Drupal\commerce_product\Entity\Product $product */
+    $product = $form['#product'];
+    $line_item_form_display = $this->getLineItemFormDisplay($product);
+
+    $line_item = LineItem::create([
```
bojanz: This needs to be $this->cartManager->createLineItem().
LineItem:: is generally only allowed in tests, real classes must use the storage, but in this case we actually need the cartManager method.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -234,6 +253,38 @@ protected function variationElement(array $form, FormStateInterface $form_state)
...
+
+    $form_display = $this->getLineItemFormDisplay($product);
+
+    $line_item_stub = LineItem::create([
+      'type' => $form_display->getTargetBundle(),
+    ]);
+
+    $form_display->buildForm($line_item_stub, $form, $form_state);
+
+    if (isset($form['quantity'])) {
```
bojanz: We need to remove all quantity settings. The default quantity can be set in the field settings, we don't need to care about it.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -365,4 +416,35 @@ protected function getAttributeValues(array $variations, $field_name, callable $
...
+  protected function getLineItemFormDisplay(ProductInterface $product) {
+    /** @var \Drupal\commerce_product\Entity\ProductType $product_type */
+    $product_type = ProductType::load($product->bundle());
+    /** @var \Drupal\commerce_product\Entity\ProductVariationType $variation_bundle */
+    $variation_bundle = ProductVariationType::load($product_type->getVariationType());
+    /** @var \Drupal\commerce_order\Entity\LineItemType $line_item_bundle */
+    $line_item_bundle = LineItemType::load($variation_bundle->getLineItemType());
+
+    /** @var \Drupal\Core\Entity\Entity\EntityFormDisplay $form_display */
+    $form_display = EntityFormDisplay::load('commerce_line_item.' . $line_item_bundle->id() . '.' . self::LINE_ITEM_FORM_DISPLAY);
```
bojanz: Use EntityFormDisplay::collectRenderDisplay() instead of this whole method.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -105,16 +113,7 @@ public function buildForm(array $form, FormStateInterface $form_state, $product
...
-      '#type' => 'number',
-      '#title' => $this->t('Quantity'),
-      '#default_value' => $settings['default_quantity'],
-      '#min' => 1,
-      '#max' => 9999,
-      '#step' => 1,
-      '#access' => !empty($settings['show_quantity']),
-      '#weight' => 99,
-    ];
+    $form = $this->lineItemElement($form, $form_state);
```
bojanz: No lineItemElement, that logic belongs in the main place

modules/product/src/Form/AddToCartForm.php
```diff
@@ -105,16 +111,24 @@ public function buildForm(array $form, FormStateInterface $form_state, $product
...
-      '#weight' => 99,
-    ];
+
+    /** @var \Drupal\commerce_product\Entity\ProductType $product_type */
+    $product_type = $this->entityTypeManager
+      ->getStorage('commerce_product_type')
+      ->load($product->bundle());
+    /** @var \Drupal\commerce_product\Entity\ProductVariationType $variation_type */
+    $variation_type = $this->entityTypeManager
+      ->getStorage('product_variation_type')
```
bojanz: Yeah, 'product_variation_type' is not exactly right :)

modules/product/src/Form/AddToCartForm.php
```diff
@@ -105,16 +111,24 @@ public function buildForm(array $form, FormStateInterface $form_state, $product
...
+      ->load($product_type->getVariationType());
+
+    $line_item_stub = LineItem::create([
```
bojanz: just call it a $line_item

modules/product/src/Form/AddToCartForm.php
```diff
@@ -105,16 +111,24 @@ public function buildForm(array $form, FormStateInterface $form_state, $product
...
+      'type' => $variation_type->getLineItemType(),
+    ]);
+
+    $form_display = EntityFormDisplay::collectRenderDisplay($line_item_stub, self::LINE_ITEM_FORM_DISPLAY);
```
bojanz: According to ContentEntityForm, once we have $form_display we should stick it in $form_state so that we don't need to repeat this in validate/submit

modules/product/src/Form/AddToCartForm.php
```diff
@@ -150,9 +164,20 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
     }
-    $quantity = $form_state->getValue('quantity');
+
     $combine = $form['#settings']['combine'];
-    $this->cartManager->addEntity($cart, $variation, $quantity, $combine);
+
+    $line_item = $this->cartManager->createLineItem($variation);
+    $line_item_form_display = EntityFormDisplay::collectRenderDisplay($line_item, self::LINE_ITEM_FORM_DISPLAY);
+
+    $line_item_values = $line_item_form_display->extractFormValues($line_item, $form, $form_state);
```
bojanz: AFAIK this will populate the fields on $line_item, you don't need to call ->set() yourself afterwards.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -0,0 +1,223 @@
+<?php
```
mglaman, line 1: I've reverted changes but kept this file. We have a few problems. I had some code that went further in with it, but it all went nowhere
- [ ] Variation comes back as null during validation, even if "variation" is "target_id"
- [ ] We need to set the unit price field, or else validation fails

mglaman:
@bojanz this deviates from the issue, but I couldn't get the `ProductVariationAttributesWidget` to cooperate.

mglaman/commerce@3cf3c43a07053fa1122340bfe427dc551ed2d1c6 fixes the selecting of an attribute from user input, not mimics how we get attributes. This wasn't passing my use case.

mglaman/commerce@cb127883e0e072f2630d2a8c019b949e0714039f fixes validation error when an attribute, that was selected, disappears.

Here is a gif of what the PR provides for me, locally.

dgfdfdfdfgdgf

mglaman:
Arg, also fresh install locally thows

```
$ drush si -y --account-pass=admin
You are about to DROP all tables in your 'data' database. Do you want to continue? (y/n): y
Starting Drupal installation. This takes a while. Consider using the --notify global option.                                                                                                                                                              [ok]

Fatal error: Call to a member function getPurchasableEntityType() on a non-object in /Users/mglaman/Drupal/sites/commerce2x/www/modules/contrib/commerce/modules/order/src/Entity/LineItem.php on line 259
Drush command terminated abnormally due to an unrecoverable error.                                                                                                                                                                                        [error]
Error: Call to a member function getPurchasableEntityType() on a non-object in /Users/mglaman/Drupal/sites/commerce2x/www/modules/contrib/commerce/modules/order/src/Entity/LineItem.php, line 259
```

modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.default.yml
```diff
@@ -22,14 +22,10 @@ content:
     type: number
     weight: 1
     settings:
-      placeholder: ''
```
bojanz, line 25: We should not be removing the unit price here, that means you can't use it in the IEF on the order add/edit page.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -160,17 +189,14 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
   }
 
   /**
-   * Builds the form elements for selecting a variation.
+   * Create variation selection widget.
```
bojanz, line 192: Why change the phpdoc to one that doesn't conform to standards (s/Create/Creates, etc)

modules/product/config/optional/core.entity_form_display.commerce_line_item.product_variation.add_to_cart.yml
```diff
@@ -0,0 +1,20 @@
...
+dependencies:
+  config:
+    - commerce_order.commerce_line_item_type.product_variation
+    - core.entity_form_mode.commerce_line_item.add_to_cart
+  module:
+    - commerce_product
+id: commerce_line_item.product_variation.add_to_cart
+targetEntityType: commerce_line_item
+bundle: product_variation
+mode: add_to_cart
```
bojanz, line 12: Note how this form display references the "add_to_cart" form mode, but that form mode is not exported.
Config dependencies are static, you can't rely on the result of hook_entity_bundle_create() being available. That's probably the cause of one of your crashes.

bojanz:
We'll need to remove both 3cf3c43 and cb12788, they are unrelated (and not ready) bug fixes that also need tests. They make the already-huge PR even harder to review and merge.

modules/product/src/Form/AddToCartForm.php
```diff
@@ -190,54 +215,47 @@ protected function variationElement(array $form, FormStateInterface $form_state)
...
-      $form['attributes'][$field_name] = [
-        '#type' => $attribute['type'],
-        '#title' => $attribute['title'],
-        '#options' => $attribute['values'],
-        '#required' => $attribute['required'],
-        '#default_value' => $selected_variation->get($field_name)->target_id,
-        '#ajax' => [
-          'callback' => '::ajaxRefresh',
-          'wrapper' => $form['#wrapper_id'],
+    else {
```
bojanz, line 219: This entire change seems to be style-only (moving the code under the else)? Wrong PR for that (and I'd disagree on the style choice, cause the cause was to lower the number of indentation levels)

bojanz:
Continued in #340 btw

PR #337, opened 2016-02-20
2664386 fixed incorrect storage and rate amount form

haringsrob:


modules/tax/src/Entity/TaxType.php
```diff
@@ -34,7 +34,6 @@
  *   config_export = {
  *     "id",
  *     "name",
- *     "genericLabel",
```
haringsrob, line 37: I'm not sure about this one, but it gave me some trouble. I removed it as it is a generated thing.

bojanz:
Committed in cbd1b22f9ef3d440730f40b7aaac9cfebb79ef1d. Thanks!

PR #340, opened 2016-02-25
Convert the add to cart form to a line item form

bojanz:


bojanz:
Looks like the add_to_cart form display isn't created properly, leading to add to cart tests failing completely.

PR #341, opened 2016-02-25
Issue #2659032: Add the cache contexts for the cart block

subhojit777:


subhojit777:
This is just the initial implementation. I'm sure the code can be refactored. Also `$view->getCacheContexts()` is always empty, even though I am doing cache settings inside `commerce_cart_block` view.

modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -207,4 +208,32 @@ protected function getCartViews(array $carts) {
   public function getCacheMaxAge() {
     return 0;
```
mglaman: If we are adding cache contexts, we should not have a max age of `0`. We will also need tests on the cart's count to verify cache is busted.

PR #345, opened 2016-02-29
Issue #2664676: Add test cases for empty store.

miteshmap:
@mglaman: As per the task here are two test cases created for order and product. looking forward for your inputs. :)

### DIFF
--- a/modules/order/src/Tests/OrderEmptyStoreTest.php
+++ b/modules/order/src/Tests/OrderEmptyStoreTest.php
@@ -1,0 +1,78 @@
+<?php
+/**
+ * @file
+ * Contains \Drupal\commerce_order\Tests\OrderEmptyStoreTestBase,
+ * Contains \Drupal\commerce_order\Tests\OrderEmptyStoreTest.
+ */
+namespace Drupal\commerce_order\Tests;
+
+use Drupal\commerce\Tests\CommerceTestBase;
+use Drupal\Core\Url;
+use Drupal\field\Tests\EntityReference\EntityReferenceTestTrait;
+use Drupal\Core\Entity\EntityInterface;
+/**
+ * Defines base class for Empty store test cases.
+ */
+abstract class OrderEmptyStoreTestBase extends CommerceTestBase {
+
+  use EntityReferenceTestTrait;
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  public static $modules = [
+    'commerce_store',
+    'commerce_product',
+    'commerce_order',
+    'field_ui',
+    'options',
+    'taxonomy',
+  ];
+
+  /**
+   * The product to test against.
+   *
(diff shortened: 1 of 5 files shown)


modules/order/src/Tests/OrderEmptyStoreTest.php
```diff
@@ -0,0 +1,78 @@
+<?php
+/**
```
bojanz: We no longer have @file docblocks in Commerce, Drupal has removed them from the coding standard.

modules/order/src/Tests/OrderEmptyStoreTest.php
```diff
@@ -0,0 +1,78 @@
...
+ * Contains \Drupal\commerce_order\Tests\OrderEmptyStoreTestBase,
+ * Contains \Drupal\commerce_order\Tests\OrderEmptyStoreTest.
+ */
+namespace Drupal\commerce_order\Tests;
+
+use Drupal\commerce\Tests\CommerceTestBase;
+use Drupal\Core\Url;
+use Drupal\field\Tests\EntityReference\EntityReferenceTestTrait;
+use Drupal\Core\Entity\EntityInterface;
+/**
```
bojanz, line 8: Missing empty line between the use statements and the docblock.

modules/order/src/Tests/OrderEmptyStoreTest.php
```diff
@@ -0,0 +1,78 @@
...
+ * Defines base class for Empty store test cases.
+ */
+abstract class OrderEmptyStoreTestBase extends CommerceTestBase {
```
bojanz: You should not be creating a new test base class, use OrderTestBase instead, add the rest in OrderEmptyStoreTest

modules/order/src/Tests/OrderEmptyStoreTest.php
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+
+/**
+ * Empty Store order page test.
+ *
+ * @group commerce
+ */
+class OrderEmptyStoreTest extends OrderEmptyStoreTestBase {
```
bojanz: A file can't contain more than one class.

modules/product/src/Tests/ProductEmptyStoreTest.php
```diff
@@ -0,0 +1,79 @@
...
+namespace Drupal\commerce_product\Tests;
+
+use Drupal\commerce\Tests\CommerceTestBase;
+use Drupal\Core\Url;
+use Drupal\field\Tests\EntityReference\EntityReferenceTestTrait;
+use Drupal\Core\Entity\EntityInterface;
+/**
+ * Defines base class for Empty store test cases.
+ */
+abstract class ProductEmptyStoreTestBase extends CommerceTestBase {
```
bojanz: Same feedback as for orders.

PR #347, opened 2016-03-03
Add a ConfigUpdater service

mglaman:
Because hook_update_N + config == hard.

### DIFF
--- a/commerce.services.yml
+++ b/commerce.services.yml
@@ -34,6 +34,10 @@ services:
     tags:
       - { name: service_collector, tag: commerce.availability_checker, call: addChecker }
 
+  commerce.config_updater:
+    class: Drupal\commerce\Config\ConfigUpdater
+    arguments: ['@entity_type.manager', '@config.storage', '@config.factory']
+
   cache_context.country:
     class: Drupal\commerce\Cache\Context\CountryCacheContext
     arguments: ['@commerce.country_context']
--- a/src/Config/ConfigUpdateResult.php
+++ b/src/Config/ConfigUpdateResult.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Drupal\commerce\Config;
+
+/**
+ * Result object for displaying configuration update results.
+ */
+class ConfigUpdateResult {
+
+  /**
+   * The configuration names that succeeded to update.
+   *
+   * @var \Drupal\Component\Render\MarkupInterface[]
+   */
+  protected $succeeded = [];
+
+  /**
+   * The configuration names that failed to update.
+   *
+   * @var \Drupal\Component\Render\MarkupInterface[]
+   */
+  protected $failed = [];
+
+  /**
+   * Constructs a new ConfigUpdateResult object.
+   *
+   * @param array $succeeded
+   *   An array of configuration names that succeeded to update.
+   * @param array $failed
+   *   An array of configuration names that failed to update.
+   */
+  public function __construct(array $succeeded, array $failed) {
+    $this->succeeded = $succeeded;
+    $this->failed = $failed;
+  }
+
+  /**
+   * Gets the configuration object names that succeeded to update.
+   *
+   * @return \Drupal\Component\Render\MarkupInterface[]
+   *   An array of messages, keyed by configuration name.
+   */
+  public function getSucceeded() {
+    return $this->succeeded;
+  }
+
+  /**
+   * Gets the configuration object names that failed to update.
+   *
+   * @return \Drupal\Component\Render\MarkupInterface[]
+   *   An array of messages, keyed by configuration name.
+   */
+  public function getFailed() {
+    return $this->failed;
+  }
+
+}
--- a/src/Config/ConfigUpdater.php
+++ b/src/Config/ConfigUpdater.php
@@ -1,0 +1,276 @@
+<?php
+
+namespace Drupal\commerce\Config;
+
+use Drupal\Component\Utility\Crypt;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Config\ExtensionInstallStorage;
+use Drupal\Core\Config\InstallStorage;
+use Drupal\Core\Config\StorageInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+
+/**
+ * Default implementation of the ConfigUpdaterInterface.
+ */
+class ConfigUpdater implements ConfigUpdaterInterface {
+
+  use StringTranslationTrait;
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * The active config storage.
+   *
+   * @var \Drupal\Core\Config\StorageInterface
+   */
+  protected $activeConfigStorage;
+
+  /**
+   * The extension config storage for config/install config items.
+   *
+   * @var \Drupal\Core\Config\StorageInterface
+   */
+  protected $extensionConfigStorage;
+
+  /**
+   * The extension config storage for config/optional config items.
+   *
+   * @var \Drupal\Core\Config\ExtensionInstallStorage
+   */
+  protected $extensionOptionalConfigStorage;
+
+  /**
+   * The config factory.
+   *
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $configFactory;
+
+  /**
+   * List of current config entity types, keyed by prefix.
+   *
+   * @var string[]
+   */
+  protected $typesByPrefix = [];
+
+  /**
+   * Constructs a new ConfigUpdater object.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   * @param \Drupal\Core\Config\StorageInterface $active_config_storage
+   *   The active config storage.
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config factory.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager, StorageInterface $active_config_storage, ConfigFactoryInterface $config_factory) {
+    $this->entityTypeManager = $entity_type_manager;
+    $this->activeConfigStorage = $active_config_storage;
+    $this->extensionConfigStorage = new ExtensionInstallStorage($active_config_storage, InstallStorage::CONFIG_INSTALL_DIRECTORY);
+    $this->extensionOptionalConfigStorage = new ExtensionInstallStorage($active_config_storage, InstallStorage::CONFIG_OPTIONAL_DIRECTORY);
+    $this->configFactory = $config_factory;
+
+    foreach ($this->entityTypeManager->getDefinitions() as $entity_type => $definition) {
+      if ($definition->isSubclassOf('Drupal\Core\Config\Entity\ConfigEntityInterface')) {
+        /** @var \Drupal\Core\Config\Entity\ConfigEntityTypeInterface $definition */
+        $prefix = $definition->getConfigPrefix();
+        $this->typesByPrefix[$prefix] = $entity_type;
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function import(array $config_names) {
+    $succeeded = [];
+    $failed = [];
+
+    foreach ($config_names as $config_name) {
+      // Read the config from the file.
+      $value = $this->loadFromExtension($config_name);
+      if (!$value) {
+        $failed[$config_name] = $this->t('@config did not exist in extension storage', ['@config' => $config_name]);
+        continue;
+      }
+
+      if ($this->loadFromActive($config_name)) {
+        $failed[$config_name] = $this->t('@config already exists, use revert to update', ['@config' => $config_name]);
+        continue;
+      }
+
+      $type = $this->getConfigType($config_name);
+
+      // Save it as a new config entity or simple config.
+      if ($type == 'system.simple') {
+        $this->configFactory->getEditable($config_name)->setData($value)->save();
+      }
+      else {
+        /** @var \Drupal\Core\Config\Entity\ConfigEntityStorageInterface $entity_storage */
+        $entity_storage = $this->entityTypeManager->getStorage($type);
+        $entity = $entity_storage->createFromStorageRecord($value);
+        $entity->save();
+      }
+
+      $succeeded[$config_name] = $this->t('@config was successfully imported', ['@config' => $config_name]);
+    }
+
+    return new ConfigUpdateResult($succeeded, $failed);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function revert(array $config_names, $skip_modified = TRUE) {
+    $succeeded = [];
+    $failed = [];
+
+    foreach ($config_names as $config_name) {
+      // Read the config from the file.
+      $value = $this->loadFromExtension($config_name);
+      if (!$value) {
+        $failed[$config_name] = $this->t('@config did not exist in extension storage', ['@config' => $config_name]);
+        continue;
+      }
+
+      // Check the configuration object's hash and see if it has been modified.
+      if ($this->isModified($config_name) && $skip_modified) {
+        $failed[$config_name] = $this->t('@config has been modified and was not reverted', ['@config' => $config_name]);
+        continue;
+      }
+
+      $type = $this->getConfigType($config_name);
+      if ($type == 'system.simple') {
+        // Load the current config and replace the value.
+        $this->configFactory->getEditable($config_name)->setData($value)->save();
+      }
+      else {
+        // Load the current config entity and replace the value, with the
+        // old UUID.
+        $definition = $this->entityTypeManager->getDefinition($type);
+        $id_key = $definition->getKey('id');
+
+        $id = $value[$id_key];
+        /** @var \Drupal\Core\Config\Entity\ConfigEntityStorageInterface $entity_storage */
+        $entity_storage = $this->entityTypeManager->getStorage($type);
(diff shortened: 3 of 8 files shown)


mglaman:
@bojanz ready for review :) and proper commit name suggestion

tests/src/Unit/ConfigUpdaterResultTest.php
```diff
@@ -0,0 +1,53 @@
+<?php
+/**
+ * Created by PhpStorm.
```
bojanz: This docblock needs to go :)

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
+<?php
+
+namespace Drupal\commerce;
+
+
```
bojanz: Double newlines between the namespace and use statements. Your inner Python developer is trying to break free :P

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
...
+use Drupal\Core\Config\StorageInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+
+/**
+ * Class ConfigUpdater.
+ */
+class ConfigUpdater implements ConfigUpdaterInterface {
+
+  use StringTranslationTrait;
```
bojanz: Missing newline after use statement.

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
...
+use Drupal\Component\Utility\Crypt;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Config\ExtensionInstallStorage;
+use Drupal\Core\Config\InstallStorage;
+use Drupal\Core\Config\StorageInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+
+/**
+ * Class ConfigUpdater.
```
bojanz: We usually document services like this with "Default implementation of the ConfigUpdaterInterface.", then place the real docs on the ConfigUpdaterInterface.

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
...
+
+  /**
+   * List of current config entity types, keyed by prefix.
+   *
+   * @var string[]
+   */
+  protected $typesByPrefix = [];
+
+  /**
+   * Constructs a ConfigReverter.
```
bojanz: Constructs a new ConfigUpdater object.

tests/src/Unit/ConfigUpdaterResultTest.php
```diff
@@ -0,0 +1,53 @@
...
+      'fake.config.name.one' => $this->prophesize(MarkupInterface::class),
+      'fake.config.name.two' => $this->prophesize(MarkupInterface::class),
+      'fake.config.name.three' => $this->prophesize(MarkupInterface::class),
+      'fake.config.name.four' => $this->prophesize(MarkupInterface::class),
+    ];
+
+    $result = new ConfigUpdaterResult([], $messages);
+
+    $this->assertEquals(count($result->getSucceeded()), 4);
+  }
```
bojanz, line 45: Missing newline before class closing brace.

tests/src/Unit/ConfigUpdaterResultTest.php
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Created by PhpStorm.
+ * User: mglaman
+ * Date: 3/4/16
+ * Time: 8:21 AM
+ */
+
+namespace Drupal\Tests\commerce\Unit;
+
+
```
bojanz: Double newline between namespace and use statements.

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
...
+      $value = $this->extensionOptionalConfigStorage->read($config_name);
+    }
+    return $value;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isModified($config_name) {
+    // Read the configuration from active storage.
```
bojanz: "Read the configuration from active storage." says the same thing as the line of code it's describing. Therefore it's redundant. We need to document the WHY and the non-obvious things. The comments below about the hashes are good examples of that (documenting non-obvious things).

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
...
+    $active = $this->activeConfigStorage->read($config_name);
+
+    // Get the hash created when the config was installed.
+    $original_hash = $active['_core']['default_config_hash'];
+
+    // Remove export keys not used to generate default config hash.
+    unset($active['uuid']);
+    unset($active['_core']);
+
+    // Recalculate the hash.
```
bojanz: "Recalculate the hash." is redundant for the same reason. Let the rehash line be a part of the unset() block, explain everything in the "Remove export keys" comment.

src/ConfigUpdaterInterface.php
```diff
@@ -0,0 +1,102 @@
...
+ * hook_post_update_NAME(), since this involves the modification of entities.
+ *
+ * @see hook_post_update_NAME()
+ */
+interface ConfigUpdaterInterface {
+
+  /**
+   * Imports configuration from extension storage to active storage.
+   *
+   * This action triggers a ConfigRevertInterface::IMPORT event if the
```
bojanz: Let's remove the comments that lie about events being triggered, then discuss events in a followup.

src/ConfigUpdaterResult.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace Drupal\commerce;
+
+/**
+ * Result object for displaying configuration update results.
+ */
+class ConfigUpdaterResult implements ConfigUpdaterResultInterface {
```
bojanz: ConfigUpdateResult would make more sense, no?

Also, this is a value object. Value objects have no interfaces. You are instantiating it directly in the ConfigUpdater class, it's not going into the container, it's not replaceable.

src/ConfigUpdaterResult.php
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Array of configuration names that succeeded to update.
+   *
+   * @var \Drupal\Component\Render\MarkupInterface[]
+   */
+  protected $succeeded = [];
+
+  /**
+   * ConfigUpdaterResult constructor.
```
bojanz: Constructs a new ConfigUpdateResult object.

src/ConfigUpdaterResult.php
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @param array $failed
+   *   An array of configuration names that failed to update.
+   * @param array $succeeded
+   *   An array of configuration names that succeeded to update.
+   */
+  public function __construct(array $failed, array $succeeded) {
```
bojanz: Wouldn't it make more sense for $succeeded to go first (in both the constructor and the methods)?

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
...
+    $active_hash = Crypt::hashBase64(serialize($active));
+
+    return $original_hash !== $active_hash;
+  }
+
+  /**
+   * Returns the config type for a given config object.
```
bojanz: "Gets the config type for a given config object." -> the verb in the docblock should match the verb in the method name.

src/ConfigUpdaterInterface.php
```diff
@@ -0,0 +1,102 @@
...
+   * configuration could be imported.
+   *
+   * @param string[] $config_names
+   *   An array of configuration names.
+   *
+   * @return \Drupal\commerce\ConfigUpdaterResultInterface
+   *   Object containing failed and succeeded configuration object update.
+   *   May also throw exceptions if there is a problem during saving the
```
bojanz: There's no such thing as "may also throw exceptions". Either the method throws exceptions, which are explicitly documented with @throws, or it doesn't (cause the exceptions are called by a method called by this method, for example), in which case no mention is needed.

src/ConfigUpdaterResult.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Drupal\commerce;
+
+/**
+ * Result object for displaying configuration update results.
+ */
+class ConfigUpdaterResult implements ConfigUpdaterResultInterface {
+
+  /**
+   * Array of configuration names that failed to update.
```
bojanz: "Array of" is redundant, and something that should be told to us by @var. Just say "The configuration names that failed to update."

src/ConfigUpdater.php
```diff
@@ -0,0 +1,279 @@
+<?php
+
+namespace Drupal\commerce;
```
bojanz: We are adding ConfigUpdater, ConfigUpdaterInterface, plus the result object to src/.
Since we have more than one file connected by a common purpose, it makes sense to make a namespace for them.
Drupal\commerce\Config perhaps?

commerce.services.yml
```diff
@@ -34,6 +34,10 @@ services:
     tags:
       - { name: service_collector, tag: commerce.availability_checker, call: addChecker }
 
+  commerce.config_update:
```
bojanz: If the name of the class is ConfigUpdater, then the service name should match, commerce.config_updater

src/ConfigUpdaterInterface.php
```diff
@@ -0,0 +1,102 @@
...
+   *   configuration.
+   */
+  public function delete(array $config_names);
+
+  /**
+   * Gets the current active value of configuration.
```
bojanz: s/Gets/Loads, so that the verb matches the method name.

src/ConfigUpdaterInterface.php
```diff
@@ -0,0 +1,102 @@
...
+   * @param string $config_name
+   *   The configuration item's full name.
+   *
+   * @return array
+   *   The configuration value.
+   */
+  public function loadFromActive($config_name);
+
+  /**
+   * Gets the extension storage value of configuration.
```
bojanz: s/Gets/Loads, so that the verb matches the method name.

bojanz:
Looks almost ready. Great job!

src/ConfigUpdaterResult.php
mglaman: Makes sense

mglaman:
@bojanz all items addressed

src/Config/ConfigUpdater.php
```diff
@@ -0,0 +1,276 @@
...
+      // Check the configuration object's hash and see if it has been modified.
+      if ($this->isModified($config_name)) {
+        $failed[$config_name] = $this->t('@config has been modified and was not deleted', ['@config' => $config_name]);
+        continue;
+      }
+
+      $type = $this->getConfigType($config_name);
+      if ($type == 'system.simple') {
+        $config = $this->configFactory->getEditable($config_name);
+        if (!$config) {
```
bojanz, line 199: getEditable() always returns a config object, so the if (!$config) will never work. Furthermore, we already checked active storage above.

src/Config/ConfigUpdater.php
```diff
@@ -0,0 +1,276 @@
...
+   * {@inheritdoc}
+   */
+  public function delete(array $config_names) {
+    $succeeded = [];
+    $failed = [];
+
+    foreach ($config_names as $config_name) {
+      $value = $this->loadFromActive($config_name);
+      if (!$value) {
+        $failed[$config_name] = $this->t('@config did not exist in extension storage', ['@config' => $config_name]);
```
bojanz, line 186: active storage, not extension storage (wrong failure message)

src/Config/ConfigUpdater.php
```diff
@@ -0,0 +1,276 @@
...
+      $value = $this->extensionOptionalConfigStorage->read($config_name);
+    }
+    return $value;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isModified($config_name) {
+    $active = $this->activeConfigStorage->read($config_name);
```
bojanz, line 244: Doesn't account for $active potentially being FALSE.

src/Config/ConfigUpdater.php
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function revert(array $config_names, $skip_modified = TRUE) {
+    $succeeded = [];
+    $failed = [];
+
+    foreach ($config_names as $config_name) {
```
bojanz, line 134: Doesn't check that the config exists in active storage (which means that the loaded $entity could be NULL, and getEditable() might be creating a new object by accident).

src/Config/ConfigUpdateResult.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Result object for displaying configuration update results.
+ */
+class ConfigUpdateResult {
+
+  /**
+   * The configuration names that succeeded to update.
+   *
+   * @var \Drupal\Component\Render\MarkupInterface[]
```
bojanz, line 13: I consider MarkupInterface to be an internal Drupal implementation detail, it has a __toString() method, we should treat it as a string.

bojanz:
Addressed the feedback above, committing.

src/Config/ConfigUpdater.php
```diff
@@ -0,0 +1,276 @@
...
+        if (!$config) {
+          $failed[$config_name] = $this->t('@config did not exist in active storage', ['@config' => $config_name]);
+          continue;
+        }
+        $config->delete();
+      }
+      else {
+        $definition = $this->entityTypeManager->getDefinition($type);
+        $id_key = $definition->getKey('id');
+        $id = $value[$id_key];
```
bojanz, line 208: We should do the same thing ConfigManager does: 
$id = substr($config_name, strlen($definition->getConfigPrefix()) + 1);

tests/src/Unit/ConfigUpdaterResultTest.php
```diff
@@ -0,0 +1,47 @@
...
+use Drupal\commerce\Config\ConfigUpdateResult;
+use Drupal\Component\Render\MarkupInterface;
+use Drupal\Tests\UnitTestCase;
+
+/**
+ * @coversDefaultClass \Drupal\commerce\Config\ConfigUpdateResult
+ * @group commerce
+ */
+class ConfigUpdaterResultTest extends UnitTestCase {
```
bojanz, line 13: I don't think there's any value in testing a value object (no pun intended).

src/Config/ConfigUpdater.php
```diff
@@ -0,0 +1,276 @@
...
+
+    foreach ($config_names as $config_name) {
+      $value = $this->loadFromActive($config_name);
+      if (!$value) {
+        $failed[$config_name] = $this->t('@config did not exist in extension storage', ['@config' => $config_name]);
+        continue;
+      }
+
+      // Check the configuration object's hash and see if it has been modified.
+      if ($this->isModified($config_name)) {
```
bojanz, line 191: I'm not sure it makes sense to check isModified() in delete(). Delete is a rare operation and should mean "really delete". If that proves to be wrong we can readd the check, along with $skip_modified.

src/Tests/ConfigUpdaterTest.php
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Test isModified on the commerce_store_type.default config object.
+   */
+  public function testIsModified() {
+    $config_name = 'commerce_store.commerce_store_type.testing';
+    $this->assertFalse($this->configUpdater->isModified($config_name));
+
+    /** @var \Drupal\commerce_store\Entity\StoreTypeInterface $store_type */
+    $store_type = \Drupal::entityTypeManager()->getStorage('commerce_store_type')
+      ->load('testing');
```
bojanz, line 58: There's no need to break ->load() into another line, core keeps it in one line.

src/Tests/ConfigUpdaterTest.php
```diff
@@ -0,0 +1,169 @@
...
+namespace Drupal\commerce\Tests;
+
+use Drupal\simpletest\WebTestBase;
+
+/**
+ * Tests the ConfigUpdater class.
+ *
+ * @group commerce
+ */
+class ConfigUpdaterTest extends WebTestBase {
```
bojanz, line 12: This test doesn't cover all of the failure modes (trying to import a non-existent config name, etc).

src/Config/ConfigUpdaterInterface.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+interface ConfigUpdaterInterface {
+
+  /**
+   * Imports configuration from extension storage to active storage.
+   *
+   * @param string[] $config_names
+   *   An array of configuration names.
+   *
+   * @return \Drupal\commerce\Config\ConfigUpdateResult
```
bojanz, line 22: Saying "The result." here is enough, they can look up the actual object.

src/Config/ConfigUpdaterInterface.php
```diff
@@ -0,0 +1,89 @@
...
+   */
+  public function import(array $config_names);
+
+  /**
+   * Reverts configuration to the value from extension storage.
+   *
+   * @param string[] $config_names
+   *   An array of configuration names.
+   * @param bool $skip_modified
+   *   Whether to skip modified configuration, defaults to TRUE. If this is
```
bojanz, line 33: The description after "Whether to skip modified configuration" is redundant (obvious from context)

src/Tests/ConfigUpdaterTest.php
```diff
@@ -0,0 +1,169 @@
+<?php
+
+namespace Drupal\commerce\Tests;
+
+use Drupal\simpletest\WebTestBase;
```
bojanz, line 5: We're testing no UIs, so it would be much faster to use a kernel test for this. Can be a followup.

bojanz:
Addressed most of the feedback above, committed in af4e4bb3507cc1119e92eadf13893d50d6941a70.

Followup:
- Convert the ConfigUpdaterTest to a kernel test.
- Cover all possible failures in the tests.

PR #349, opened 2016-03-04
2674888 Added query string and populate add to cart

haringsrob:
We might want to improve by adding a constant for "dv" so that if we ever change the tactic, it will be an easy task.

### DIFF
--- a/modules/cart/src/Form/AddToCartForm.php
+++ b/modules/cart/src/Form/AddToCartForm.php
@@ -81,6 +81,13 @@ class AddToCartForm extends ContentEntityForm {
       $form['#access'] = FALSE;
     }
 
+    // If our default variation is set, we alter the add to cart form.
+    if ($default_variation = \Drupal::request()->query->get('dv')) {
+      if (is_numeric($default_variation)) {
+        $form['purchased_entity']['widget'][0]['attributes']['field_variation']['#default_value'] = $default_variation;
+      }
+    }
+
     return $form;
   }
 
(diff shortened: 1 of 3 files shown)


modules/cart/src/Form/AddToCartForm.php
```diff
@@ -81,6 +81,13 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       $form['#access'] = FALSE;
     }
 
+    // If our default variation is set, we alter the add to cart form.
+    if ($default_variation = \Drupal::request()->query->get('dv')) {
+      if (is_numeric($default_variation)) {
+        $form['purchased_entity']['widget'][0]['attributes']['field_variation']['#default_value'] = $default_variation;
```
mglaman: Should we check if this is a valid product variation identifier for the given product, too?

mglaman: Also, if this is given, we could set it earlier and bypass other logic, making it a bit more performant

haringsrob: I was thinking about validation as well. But how will we handle it then? Set a message? Should we allow this message to be altered?

I agree on doing this at an earlier stage, but I will need a suggestion where? viewElements?

mglaman: You'd want to move this to the new field widget, for product attributes. Should be able to check before it invokes its methods to discover the variation to be selected by default

PR #350, opened 2016-03-05
2680901 Store add link no longer has a destination

haringsrob:


modules/store/commerce_store.module
```diff
@@ -61,3 +61,32 @@ function template_preprocess_commerce_store(array &$variables) {
...
+/**
+ * Helper function to generate add store links.
+ *
+ * @param array $options
+ *   Array of options to pass to the route generation.
+ *
+ * @return \Drupal\Core\Link
+ *   The link object to be sent to.
+ */
+function _commerce_store_get_add_link($options = []) {
```
mglaman: Is there a way we could always link to .add_page, and have .add_page intelligently redirect to the single form if there is one store?

haringsrob: Looking into this. :)

haringsrob:
Hi,

As discussed with @bojanz , we removed all destination parameters from this. As there is an idea to have a sort of "setup".

bojanz:
Committed in b6e2283ec599ed50da45b5cfbc5e62bdf598e7f7. Thanks!

PR #352, opened 2016-03-09
2603482 order language behavior

haringsrob:
Hi,

I have made an attempt for the implementation that has the following requirements:

1) langcode field on the order
2) ensure it's populated (by CartProvider)
3) ensure it can be selected when adding a new order, if there are multiple languages


modules/cart/src/CartManager.php
```diff
@@ -56,6 +64,7 @@ public function __construct(EntityTypeManagerInterface $entity_type_manager, Lin
     $this->lineItemStorage = $entity_type_manager->getStorage('commerce_line_item');
     $this->lineItemMatcher = $line_item_matcher;
     $this->eventDispatcher = $event_dispatcher;
+    $this->languageManager = \Drupal::languageManager();
```
mglaman: Update `services.yml` to inject this into the service. Update `create` method to get it from service, make it an argument.

modules/cart/src/EventSubscriber/CartLanguage.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+class CartLanguage implements EventSubscriberInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    $events['commerce_cart.entity.add'][] = array('updateLanguage', 0);
+    $events['commerce_cart.line_item.update'][] = array('updateLanguage', 0);
+    $events['commerce_cart.line_item.remove'][] = array('updateLanguage', 0);
```
mglaman: Short syntax for arrays here

modules/cart/src/EventSubscriber/CartLanguage.php
```diff
@@ -0,0 +1,35 @@
...
+    return $events;
+  }
+
+  /**
+   * Update the language of the order.
+   */
+  public function updateLanguage($event) {
+    $cart = $event->getCart();
+
+    $order_language = \Drupal::languageManager()->getCurrentLanguage();
```
mglaman: Do events supported dependency injection? I have no idea, actually. Worth checking and doing so instead of `\Drupal::`

haringsrob:
Hi @mglaman,

I have updated the code.

Regards

modules/cart/src/CartManager.php
```diff
@@ -118,6 +129,13 @@ public function addLineItem(OrderInterface $cart, LineItemInterface $line_item,
       $needs_cart_save = TRUE;
     }
 
+    // If the language is different, we need to save it again. If we are already
+    // planning to trigger the save, there is no need to do the additional
+    // checks.
+    if (!$needs_cart_save && $cart->getLanguage()->getId() <> $this->languageManager->getCurrentLanguage()->getId()) {
```
borisson: I don't see any usage of `<>` in the php code, I think this should be replaced by `!==`

modules/cart/src/EventSubscriber/CartLanguage.php
```diff
@@ -0,0 +1,52 @@
...
+   * @var Drupal\Core\Language\LanguageManager
+   */
+  protected $languageManager;
+
+  /**
+   * CartLanguage constructor.
+   *
+   * @param Drupal\Core\Language\LanguageManager $language_manager
+   */
+  public function __construct(LanguageManager $language_manager) {
```
borisson: Can we typehint on the interface here instead of the concrete class? `Drupal\Core\Language\LanguageManagerInterface`

modules/cart/src/EventSubscriber/CartLanguage.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Drupal\commerce_cart\EventSubscriber;
+
+use Drupal\Core\Language\LanguageManager;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Class OrderSubscriber.
```
borisson: The class description here doesn't really add anything, and it's wrong. 

`Makes a cart language-aware.` or something similar would make more sense.

steveoliver:
This needs a re-roll, then we can do an updated review.

haringsrob:
@steveoliver @borisson,

I have updated the comments and rerolled onto latest 8.x

PR #353, opened 2016-03-13
Issue #2660266 by czigor: Provide way to create new variation bundle …

mglaman:
…when making new product type


bojanz:
Better to continue this on the issue, since the result needs work.

PR #354, opened 2016-03-14
2674888 change variation link

haringsrob:


modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -259,6 +264,30 @@ protected function selectVariationFromUserInput(array $variations, array $user_i
...
+   *
+   * @return \Drupal\commerce_product\Entity\ProductVariationInterface
+   *   The selected variation.
+   */
+  protected function selectVariationFromQueryString(array $variations) {
+    // Always set the default variation.
+    $current_variation = reset($variations);
+    // If our query string is present and it is one of the available variations
+    // we modify the current variation.
+    if ($query_string_id = \Drupal::request()->query->get('dv')) {
```
mglaman: We should inject the request, if it's not already an available property.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -259,6 +264,30 @@ protected function selectVariationFromUserInput(array $variations, array $user_i
...
+      if (array_key_exists($query_string_id, $variations)) {
```
andypost, line 284: isset($variations[$query_string_id])

haringsrob:
Changes by @agoradesign (https://github.com/agoradesign) look good.

I think they full fill the remaining requirements.

Tested the update and everything is still functional.

modules/product/src/Entity/ProductVariationInterface.php
```diff
@@ -12,6 +12,11 @@
 interface ProductVariationInterface extends PurchasableEntityInterface, EntityChangedInterface, EntityOwnerInterface {
 
   /**
+   * Query string parameter name for defining a canonical URL to a product variation.
+   */
+  const CANONICAL_QUERY_PARAM = 'dv';
```
mglaman: let's just make this `v`

modules/product/commerce_product.module
```diff
@@ -102,7 +103,16 @@ function commerce_product_commerce_product_view(array &$build, EntityInterface $
     /** @var \Drupal\commerce_product\ProductVariationFieldRendererInterface $variation_field_renderer */
     $variation_field_renderer = \Drupal::service('commerce_product.variation_field_renderer');
 
-    $rendered_fields = $variation_field_renderer->renderFields($entity->getDefaultVariation(), $view_mode);
+    // Check if a variation was specified via its canonical URL.
+    // @todo: Core has same problems in regard to translation, follow core's pattern.
```
mglaman, line 107: We need to process this at the route level. We need to throw a 404 if variation is not part of the product. And then set the global context if its available, retrieve it here.

bojanz:
Continued in the patch at https://www.drupal.org/node/2674888

PR #356, opened 2016-03-19
Issue #2684713 by skyredwang: WSOD if price amount is NULL

mglaman:


bojanz:
Committed in 579ef829cf9bcdfec2379269afeb2988e080dead.

PR #358, opened 2016-03-20
Issue #2689911 by andypost: Use a lazy builder in the AddToCartFormatter

andypost:
That's it

### DIFF
--- a/modules/cart/commerce_cart.services.yml
+++ b/modules/cart/commerce_cart.services.yml
@@ -14,3 +14,7 @@ services:
   commerce_cart.line_item_matcher:
       class: Drupal\commerce_cart\LineItemMatcher
       arguments: ['@event_dispatcher']
+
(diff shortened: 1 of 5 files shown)


andypost:
Missed https://www.drupal.org/node/2689905

bojanz:
This looks good. However, commerce_cart shouldn't be aware of commerce_product, only the opposite. Since the code in CartLazyBuilders is product specific, it makes more sense for it to live in commerce_product as ProductLazyBuilders (this also allows you to remove the $product_type_id parameter, which is misnamed since it actually gives you the entity type id -> 'commerce_product')

andypost:
Thanx for pointer, I got missed the point of mapping, will update latter

andypost:
Looks ready, should I add a patch to d.o issue or how to proceed?

PR #360, opened 2016-03-23
Allow the view in the LineItemTable formatter to be configured

joshmiller83:
Related Drupal issue:

https://www.drupal.org/node/2689919


modules/order/src/Plugin/Field/FieldFormatter/LineItemTable.php
```diff
@@ -17,7 +22,53 @@
...
-class LineItemTable extends FormatterBase {
+class LineItemTable extends FormatterBase implements FormatterInterface, ContainerFactoryPluginInterface {
+
+  /**
+   * @var \Drupal\Core\Entity\Query\QueryInterface
+   */
+  protected $entityQuery;
+
+  /**
+   * Constructs a new PriceDefaultFormatter object.
```
joshmiller83, line 33: This is clearly a copy+paste that needs to be updated.

bojanz:
Let's continue the discussion in https://www.drupal.org/node/2689919

PR #361, opened 2016-03-29
Require Drupal 8.1

bojanz:
New alphas of entity api and profile tagged, drupal_ti PR used by our travis.yml, no blockers left.


Berdir:
Note: You can specify the pull request commit hash when fetching drupal_ti, I did that also with redirect: https://github.com/md-systems/redirect/blob/8.x-1.x/.travis.yml#L105 (note that that is an old commit has from my own PR, don't use that).

Then you don't have to wait until it is merged.

bojanz:
Wow, that actually works for unmerged PRs. Thanks @Berdir!

PR #366, opened 2016-04-15
implement basic price calculation logic

agoradesign:
Here's my first draft for the price calculation logic. Here are a few remarks:
- I've chosen the name BasePriceResolver, like proposed by @bojanz last week.
- To be consistent with other resolvers, I've stuck to the pattern of basic and chainable resolvers, which also extend the basic interface and thererfore may be nested
- @bojanz also mentioned, it should basically have the same signature as the Availability API, which basically has a similar structure as the resolvers, but are flat and may not be nested. In contrary, the AvailabilityCheckerInterface also defines an applies() function. My default implementation of the BasePriceResolverInterface does have an function named applies(), howerver I restrained from putting it into the interface. Because as soon you make it chainable, every chained resolver also has to implement this function and further ask all children. At the time of actually resolving the price, you'd call the function twice therefore. However, these are implementation details, that can be easily changed/extended...

As the default implementation is meant for pulling the price field from ProductVariation entites, and the ProductVariation entity possesses a base field called "price", I'm convinced that there's not a more sophisticated approach necessary than checking for:

$entity->hasField('price') && !$entity->get('price')->isEmpty()

Because every other use case, like having different purchasable entities, more price fields, etc would  need a custom price resolver anyway...

### DIFF
--- a/modules/cart/src/Form/AddToCartForm.php
+++ b/modules/cart/src/Form/AddToCartForm.php
@@ -6,6 +6,7 @@ use Drupal\commerce\PurchasableEntityInterface;
 use Drupal\commerce_cart\CartManagerInterface;
 use Drupal\commerce_cart\CartProviderInterface;
 use Drupal\commerce_order\Resolver\OrderTypeResolverInterface;
+use Drupal\commerce_price\Resolver\ChainPriceResolverInterface;
 use Drupal\commerce_store\StoreContextInterface;
 use Drupal\Core\Entity\EntityManagerInterface;
 use Drupal\Core\Entity\ContentEntityForm;
@@ -47,6 +48,13 @@ class AddToCartForm extends ContentEntityForm {
   protected $storeContext;
 
   /**
+   * The chain base price resolver.
+   *
+   * @var \Drupal\commerce_price\Resolver\ChainPriceResolverInterface
+   */
+  protected $chainPriceResolver;
+
+  /**
    * Constructs a new AddToCartForm object.
    *
    * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
@@ -59,14 +67,17 @@ class AddToCartForm extends ContentEntityForm {
    *   The order type resolver.
    * @param \Drupal\commerce_store\StoreContextInterface $store_context
    *   The store context.
+   * @param \Drupal\commerce_price\Resolver\ChainPriceResolverInterface $chain_price_resolver
+   *   The chain base price resolver.
    */
-  public function __construct(EntityManagerInterface $entity_manager, CartManagerInterface $cart_manager, CartProviderInterface $cart_provider, OrderTypeResolverInterface $order_type_resolver, StoreContextInterface $store_context) {
+  public function __construct(EntityManagerInterface $entity_manager, CartManagerInterface $cart_manager, CartProviderInterface $cart_provider, OrderTypeResolverInterface $order_type_resolver, StoreContextInterface $store_context, ChainPriceResolverInterface $chain_price_resolver) {
     parent::__construct($entity_manager);
 
     $this->cartManager = $cart_manager;
     $this->cartProvider = $cart_provider;
     $this->orderTypeResolver = $order_type_resolver;
     $this->storeContext = $store_context;
+    $this->chainPriceResolver = $chain_price_resolver;
   }
 
   /**
@@ -78,7 +89,8 @@ class AddToCartForm extends ContentEntityForm {
       $container->get('commerce_cart.cart_manager'),
       $container->get('commerce_cart.cart_provider'),
       $container->get('commerce_order.chain_order_type_resolver'),
-      $container->get('commerce_store.store_context')
+      $container->get('commerce_store.store_context'),
+      $container->get('commerce_price.chain_price_resolver')
     );
   }
 
@@ -143,8 +155,7 @@ class AddToCartForm extends ContentEntityForm {
     $entity = parent::buildEntity($form, $form_state);
     // Now that the purchased entity is set, populate the title and price.
     $entity->setTitle($entity->getPurchasedEntity()->getLineItemTitle());
-    // @todo Remove once the price calculation is in place.
-    $entity->unit_price = $entity->getPurchasedEntity()->price;
+    $entity->setUnitPrice($this->chainPriceResolver->resolve($entity->getPurchasedEntity()));
 
     return $entity;
   }
--- a/modules/price/src/Resolver/DefaultPriceResolver.php
+++ b/modules/price/src/Resolver/DefaultPriceResolver.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Drupal\commerce_price\Resolver;
+
+use Drupal\commerce\PurchasableEntityInterface;
+
+/**
+ * Returns the price based on the purchasable entity's price field.
+ */
+class DefaultPriceResolver implements PriceResolverInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function resolve(PurchasableEntityInterface $entity, $quantity = 1) {
+    return $entity->getPrice();
+  }
+
+}
--- a/modules/cart/commerce_cart.info.yml
+++ b/modules/cart/commerce_cart.info.yml
@@ -6,6 +6,7 @@ core: 8.x
 dependencies:
   - commerce
   - commerce_order
+  - commerce_price
   - commerce_product
   - views
 config_devel:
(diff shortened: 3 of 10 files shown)


mglaman:
This would need to be updated with new pricing object

andreasradloff:
I am the maintainer of the commerce_pricelist module and I have some thoughts on this.

The suggested resolver method takes a PurchasableEntity and a quantity as argument. These are derived calling getPurchasedEntity() on the line item. However, quantity is just one context which could be useable to derive a price, for example in the case of Commerce Pricelist prices can differ based on the user (customer segments) or order date (for time limited pricelists, discounts etc.).

The current approach provides no way to find the order date or user that the order belongs to, as this can be a different user than the loged in user. For example an administrator editing a order owned by another user.

All of this could be solved by passing the line item to the resolver instead. The default resolver could then get the purchased entity, quantity and the price inside the resolve method.

agoradesign:
I've rebased the PR + updated code to use the pricing value object instead of the field as return value.

@andreasradloff the base price resolver is meant to calculate a price of any purchasable item, no matter if it's in an order context or not. Using line items would force you to always create dummy line items, only for calculating the price. E.g. one important thing that's currently missing in Commerce is to display the actual calculated price in the product templates: there's no line item and also no need for creating one.

Of course it makes sense to include other contextual information into the price calculation than just the product variation and the quantity. But then we should rather include an extra parameter specifiying such contextual parameters.

Anyway, these are decisions that only the Commerce maintainers can make...

andreasradloff:
Actually in D7 that is how it worked, whenever the calculated price was displayed a line item was created. Could the line item be an optional context? I have seen $context as an array parameter elsewhere in Drupal 8 which contains whatever context is needed. 
I'm just worried that if the line item or other context is optional then me and other maintainers of modules that alter the price based on some context are going to spend a lot of time posting issues to remind module xyz to include the line item as context. 

Skickat från min Samsung-enhet

-------- Originalmeddelande --------
Från: agoradesign notifications@github.com 
Datum: 29-08-2016  21:06  (GMT+01:00) 
Till: drupalcommerce/commerce commerce@noreply.github.com 
Kopia: andreasradloff andreas@radloff.se, Mention mention@noreply.github.com 
Rubrik: Re: [drupalcommerce/commerce] implement basic price calculation logic
  (#366) 

I've rebased the PR + updated code to use the pricing value object instead of the field as return value.

@andreasradloff the base price resolver is meant to calculate a price of any purchasable item, no matter if it's in an order context or not. Using line items would force you to always create dummy line items, only for calculating the price. E.g. one important thing that's currently missing in Commerce is to display the actual calculated price in the product templates: there's no line item and also no need for creating one.

Of course it makes sense to include other contextual information into the price calculation than just the product variation and the quantity. But then we should rather include an extra parameter specifiying such contextual parameters.

Anyway, these are decisions that only the Commerce maintainers can make...

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or mute the thread.

{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/drupalcommerce/commerce","title":"drupalcommerce/commerce","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/drupalcommerce/commerce"}},"updates":{"snippets":[{"icon":"PERSON","message":"@agoradesign in #366: I've rebased the PR + updated code to use the pricing value object instead of the field as return value.\r\n\r\n@andreasradloff the base price resolver is meant to calculate a price of any purchasable item, no matter if it's in an order context or not. Using line items would force you to always create dummy line items, only for calculating the price. E.g. one important thing that's currently missing in Commerce is to display the actual calculated price in the product templates: there's no line item and also no need for creating one.\r\n\r\nOf course it makes sense to include other contextual information into the price calculation than just the product variation and the quantity. But then we should rather include an extra parameter specifiying such contextual parameters.\r\n\r\nAnyway, these are decisions that only the Commerce maintainers can make..."}],"action":{"name":"View Pull Request","url":"https://github.com/drupalcommerce/commerce/pull/366#issuecomment-243222857"}}}

mglaman:
@agoradesign could you enable the "allow maintainers to push" for the PR? Or I wonder if we have to close and re-open to enable that. Either way, I'd like to start fiddling along with this :)

modules/pricing/commerce_pricing.services.yml
```diff
@@ -0,0 +1,10 @@
+services:
+  commerce_pricing.chain_base_price_resolver:
+    class: Drupal\commerce_pricing\Resolver\ChainBasePriceResolver
+    tags:
+      - { name: service_collector, call: addResolver, tag: commerce_pricing.base_price_resolver }
+
+  commerce_pricing.default_base_price_resolver:
+    class: Drupal\commerce_pricing\Resolver\DefaultBasePriceResolver
+    tags:
+      - { name: commerce_pricing.base_price_resolver, priority: -100 }
```
mglaman: This should be 100. Otherwise it'll always win

agoradesign: No ;) afaik Symfony weights in services are the opposite of the Drupal style weights - and I'm acutally using custom resolvers with a priority of "10"

mglaman:
@bojanz I think this is ready, so we can merge and bring into the order refresh process.

modules/price/src/Resolver/DefaultPriceResolver.php
```diff
@@ -0,0 +1,34 @@
...
+   * Determines whether the resolver applies to the given purchasable entity.
+   *
+   * @param \Drupal\commerce\PurchasableEntityInterface $entity
+   *   The purchasable entity.
+   *
+   * @return bool
+   *   TRUE if the resolver applies to the given purchasable entity, FALSE
+   *   otherwise.
+   */
+  public function applies(PurchasableEntityInterface $entity) {
```
bojanz: We need to put applies() on the interface, and have it be before resolve(). Invoked by the chain resolver, instead of called by resolve(). The current solution is ugly for no visible gain

modules/price/src/Resolver/DefaultPriceResolver.php
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Returns the price based on the purchasable entity's price field.
+ */
+class DefaultPriceResolver implements PriceResolverInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function resolve(PurchasableEntityInterface $entity, $quantity = 1) {
+    return $this->applies($entity) ? $entity->price->first()->toPrice() : NULL;
```
bojanz: We have no guarantee $entity->price exists. Let's add getPrice() to PurchasableEntityInterface. We can copy it from the variation interface.

modules/cart/src/Form/AddToCartForm.php
```diff
@@ -143,8 +155,7 @@ public function buildEntity(array $form, FormStateInterface $form_state) {
     $entity = parent::buildEntity($form, $form_state);
     // Now that the purchased entity is set, populate the title and price.
     $entity->setTitle($entity->getPurchasedEntity()->getLineItemTitle());
-    // @todo Remove once the price calculation is in place.
-    $entity->unit_price = $entity->getPurchasedEntity()->price;
+    $entity->setUnitPrice($this->chainPriceResolver->resolve($entity->getPurchasedEntity()));
```
bojanz, line 158: Maybe we should just remove this, and skip injecting the resolver? This is supposed to be done by the order refresh process, which needs to land today as well.

bojanz (changes requested):

modules/price/src/Resolver/DefaultPriceResolver.php
mglaman: Added \Drupal\commerce\PurchasableEntityInterface::getPrice

modules/price/src/Resolver/DefaultPriceResolver.php
mglaman: Removed

mglaman:
Merged

bojanz:
@andreasradloff
I agree about the need for a context. We'll add a context value object (with date, user, store) in a followup.

steveoliver:
See follow-up Drupal.org issue: [Create context value object for price resolvers and availability checkers](https://www.drupal.org/node/2803221)

PR #367, opened 2016-04-20
Issue #2710257 by agoradesign: Re-introduce hook_commerce_cart_attrib…

agoradesign:
Re-introduce hook_commerce_cart_attributes_refresh_alter() as hook_commerce_product_attributes_refresh_alter(), like described here: https://www.drupal.org/node/2710257

### DIFF
--- a/modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
+++ b/modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
@@ -357,7 +357,12 @@ class ProductVariationAttributesWidget extends WidgetBase implements ContainerFa
    * Ajax callback.
    */
   public static function ajaxRefresh(array $form, FormStateInterface $form_state) {
-    return $form;
+    /** @var \Drupal\Core\Ajax\AjaxResponse $response */
+    $response = \Drupal::service('main_content_renderer.ajax')->renderResponse($form, \Drupal::service('request_stack')->getCurrentRequest(), \Drupal::service('current_route_match'));
+
+    // Allow other modules to add arbitrary AJAX commands on the refresh.
+    \Drupal::moduleHandler()->alter('commerce_product_attributes_refresh', $response, $form, $form_state);
+    return $response;
   }
 
 }
--- a/modules/product/commerce_product.api.php
+++ b/modules/product/commerce_product.api.php
@@ -1,0 +1,42 @@
+<?php
+
+/**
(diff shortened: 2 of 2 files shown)


modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -357,7 +357,12 @@ protected function getAttributeValues(array $variations, $field_name, callable $
    * Ajax callback.
    */
   public static function ajaxRefresh(array $form, FormStateInterface $form_state) {
-    return $form;
+    /** @var \Drupal\Core\Ajax\AjaxResponse $response */
+    $response = \Drupal::service('main_content_renderer.ajax')->renderResponse($form, \Drupal::service('request_stack')->getCurrentRequest(), \Drupal::service('current_route_match'));
```
mglaman, line 361: Do any of the base classes not already inject this is some way?

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -357,7 +357,12 @@ protected function getAttributeValues(array $variations, $field_name, callable $
...
+
+    // Allow other modules to add arbitrary AJAX commands on the refresh.
+    \Drupal::moduleHandler()->alter('commerce_product_attributes_refresh', $response, $form, $form_state);
```
mglaman, line 364: See above

agoradesign:
It's a static function

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
agoradesign, line 364: It's a static function

mglaman, line 364: Ha :P I can read

agoradesign:
I guessed so, but how should an injected service be called on a static function? No instance, no injected service ;)

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
agoradesign, line 364: I guessed so, but how should an injected service be called within a static function? No instance, no injected service ;)

bojanz:
Continued in #377

PR #368, opened 2016-04-27
Issue #2713891 by mglaman: Ability to use fields on an commerce_produ…

mglaman:
…ct_attribute_value for selecting that attribute

Doesn't currently have a test.

### DIFF
--- a/modules/cart/src/Tests/AddToCartFormTest.php
+++ b/modules/cart/src/Tests/AddToCartFormTest.php
@@ -17,6 +17,7 @@ use Drupal\commerce_product\Entity\ProductVariationInterface;
 use Drupal\commerce_product\Entity\ProductVariationType;
 use Drupal\commerce_product\Entity\ProductVariationTypeInterface;
 use Drupal\Core\Entity\Entity\EntityFormDisplay;
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
 
@@ -270,6 +271,59 @@ class AddToCartFormTest extends OrderTestBase {
   }
 
   /**
+   * Tests that the add to cart form renders an attribute entity.
+   */
+  public function testRenderedAttributeElement() {
+    /** @var \Drupal\commerce_product\Entity\ProductVariationTypeInterface $variation_type */
+    $variation_type = ProductVariationType::load($this->variation->bundle());
+
+    $color_attribute_values = $this->createAttributeSet($variation_type, 'color', [
+      'cyan' => 'Cyan',
+      'magenta' => 'Magenta',
+    ]);
+    $color_attribute_values['cyan']->set('rendered_test', 'Cyan (Rendered)')->save();
+    $color_attribute_values['cyan']->save();
+    $color_attribute_values['magenta']->set('rendered_test', 'Magenta (Rendered)')->save();
+    $color_attribute_values['magenta']->save();
+
+    $color_attribute = ProductAttribute::load($color_attribute_values['cyan']->getAttributeId());
+
+    $variation1 = $this->createEntity('commerce_product_variation', [
+      'type' => 'default',
+      'sku' => $this->randomMachineName(),
+      'price' => [
+        'amount' => 999,
+        'currency_code' => 'USD',
+      ],
+      'attribute_color' => $color_attribute_values['cyan'],
+    ]);
+    $variation2 = $this->createEntity('commerce_product_variation', [
+      'type' => 'default',
+      'sku' => $this->randomMachineName(),
+      'price' => [
+        'amount' => 999,
+        'currency_code' => 'USD',
+      ],
+      'attribute_color' => $color_attribute_values['magenta'],
+    ]);
+    $product = $this->createEntity('commerce_product', [
+      'type' => 'default',
+      'title' => $this->randomMachineName(),
+      'stores' => [$this->store],
+      'variations' => [$variation1, $variation2],
+    ]);
+
+    $this->drupalGet($product->toUrl());
+    $this->assertOptionWithDrupalSelector('edit-purchased-entity-0-attributes-attribute-color', $color_attribute_values['cyan']->id());
+
+    $color_attribute->set('elementType', 'commerce_product_rendered_attribute')->save();
+
+    $this->drupalGet($product->toUrl());
+    $this->assertText('Cyan (Rendered)');
+    $this->assertText('Magenta (Rendered)');
+  }
+
+  /**
    * Creates an attribute field and set of attribute values.
    *
    * @param \Drupal\commerce_product\Entity\ProductVariationTypeInterface $variation_type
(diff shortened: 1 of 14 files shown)


modules/product/src/Element/CommerceProductRenderedAttribute.php
```diff
@@ -0,0 +1,78 @@
...
+use Drupal\Component\Utility\Html as HtmlUtility;
+
+/**
+ * Provides a form input element for rendering attributes as radio buttons.
+ *
+ * Example usage:
+ * @code
+ * $form['settings']['active'] = array(
+ *   '#type' => 'commerce_product_rendered_attribute',
+ *   '#title' => $this->t('Poll status'),
```
bojanz: Poll status is a bit of a weird example :) Looks like we need to rewrite the entire docblock.

modules/product/src/Element/CommerceProductRenderedAttribute.php
```diff
@@ -0,0 +1,78 @@
...
+      /** @var \Drupal\Core\Entity\EntityStorageInterface $storage */
+      $storage = \Drupal::service('entity_type.manager')->getStorage('commerce_product_attribute_value');
+      $attribute_values = $storage->loadMultiple(array_keys($element['#options']));
+
+      /** @var \Drupal\Core\Render\RendererInterface $renderer */
+      $renderer = \Drupal::service('renderer');
+
+      $weight = 0;
+      foreach ($element['#options'] as $key => $choice) {
+        $rendered_attribute = \Drupal::entityTypeManager()->getViewBuilder('commerce_product_attribute_value')->view($attribute_values[$key], 'add_to_cart');
```
bojanz: Let's move $view_builder =  \Drupal::entityTypeManager()->getViewBuilder('commerce_product_attribute_value') to its own line, to shorten this.

modules/product/src/Entity/ProductAttribute.php
```diff
@@ -70,6 +78,13 @@ public function getValues() {
   /**
    * {@inheritdoc}
    */
+  public function getElementType() {
+    return (!$this->elementType) ? 'select' : $this->elementType;
```
bojanz: Can't the default be provided on the class attribute?

modules/product/src/Entity/ProductAttributeInterface.php
```diff
@@ -16,4 +16,10 @@
    */
   public function getValues();
 
+  /**
+   * Gets the attribute element type.
+   * 
+   * @return string
+   */
+  public function getElementType();
```
bojanz, line 25: Missing empty line before }

modules/product/src/ProductAttributeFieldManager.php
```diff
@@ -143,7 +143,7 @@ protected function buildFieldMap() {
         $handler_settings = $definition->getSetting('handler_settings');
         $component = $form_display->getComponent($field_name);
 
-        $field_map[$bundle][] = [
+        $field_map[$bundle][$field_name] = [
```
bojanz: Why are we making this change?

bojanz:
We need to remove all mentions of attribute_widget_type (the widget setting).

We also need the JS for the element: http://cgit.drupalcode.org/commerce_fancy_attributes/tree/commerce_fancy_attributes.js

That, the CSS, the form element might still belong in commerce_fancy_attributes, since the user will need that module anyway (for the color swatch formatter).

modules/product/src/ProductAttributeFieldManager.php
mglaman: See https://github.com/drupalcommerce/commerce/pull/368/files#diff-354d71e20ba2cfced9d85f311aefa77cR303

So we can foreach, use key for field name and still have attribute name for that field (so we can load it, see how it should render.) I feel there is a better approach, but last night it was +1 to get it done.

mglaman:
@bojanz why do we need the JS? the CSS is hiding the radio button and it is being handled as is? I figured commerce_fancy_attributes would be our placeholder for crazy field formatters (like hex code to square swatch.)

mglaman:
@bojanz we have passing tests :)

modules/product/src/Form/ProductAttributeForm.php
```diff
@@ -3,13 +3,44 @@
...
+  protected $entityFieldManager;
+
+  /**
+   * Constructs a new ProductAttributeForm object.
+   *
+   * @param \Drupal\Core\Entity\EntityFieldManagerInterface $entity_field_manager
+   *   The entity field manager.
+   */
+  public function __construct(EntityFieldManagerInterface $entity_field_manager) {
+    $this->entityFieldManager = $entity_field_manager;
```
bojanz: You don't appear to be using the entity field manager anywhere?

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -96,57 +104,6 @@ public static function isApplicable(FieldDefinitionInterface $field_definition)
   /**
    * {@inheritdoc}
    */
-  public static function defaultSettings() {
-    return [
-      'attribute_widget_type' => 'select',
```
bojanz: There's also the schema in attribute_widget_type.
We also need to re-export core.entity_form_display.commerce_line_item.product_variation.add_to_cart.yml to remove the now-gone setting.

modules/product/src/Tests/ProductRenderedAttributeTest.php
```diff
@@ -0,0 +1,182 @@
...
+  protected function setUp() {
+    parent::setUp();
+
+    $this->attributeFieldManager = \Drupal::service('commerce_product.attribute_field_manager');
+  }
+
+  /**
+   * Tests that the add to cart form renders an attribute entity.
+   */
+  public function testRenderedAttributeElement() {
```
bojanz: Why not put this method in the old AddToCartFormTest?

mglaman: Good call. I'm actually having to borrow two methods from it anyways.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
mglaman: Done!

modules/cart/src/Tests/AddToCartFormTest.php
```diff
@@ -270,6 +271,60 @@ public function testMultipleVariations() {
...
+
+    $this->drupalGet($product->toUrl());
+    $this->assertOptionWithDrupalSelector('edit-purchased-entity-0-attributes-attribute-color', $color_attribute_values['cyan']->id());
+
+    $color_attribute->set('elementType', 'commerce_product_rendered_attribute')->save();
+
+    $this->drupalGet($product->toUrl());
+    $this->assertText('Cyan (Rendered)');
+    $this->assertText('Magenta (Rendered)');
+
```
bojanz: Extra newline.

modules/product/commerce_product.libraries.yml
```diff
@@ -3,3 +3,8 @@ form:
   css:
     theme:
       css/product.form.css: {}
+rendered-attributes:
```
bojanz, line 7: Let's add a newline above rendered-attributes.

modules/product/css/product.rendered-attributes.css
```diff
@@ -0,0 +1,14 @@
+.product--rendered-attribute .form-item {
```
bojanz: This file should be commerce_product.rendered-attributes.css and have a docblock on top.

modules/product/src/Entity/ProductAttributeInterface.php
```diff
@@ -16,4 +16,11 @@
    */
   public function getValues();
 
+  /**
+   * Gets the attribute element type.
+   *
+   * @return string
```
bojanz, line 22: Missing description below @return string

mglaman, line 22: `The element type name.`

modules/product/src/Element/CommerceProductRenderedAttribute.php
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * The options must be an array of attribute values, keyed by the entity's ID.
+ *
+ * Example usage:
+ * @code
+ * $form['rendered_attributes'] = array(
+ *   '#type' => 'commerce_product_rendered_attribute',
+ *   '#title' => $this->t('Attributes'),
+ *   '#default_value' => 1,
+ *   '#options' => array(0 => 'Red', 1 => 'Blue),
```
bojanz: Let's not use array() even in examples.

PR #374, opened 2016-05-03
2706207 optional attribute

haringsrob:


bojanz:
Let's unify testProductOptionalAttributeRequiredSingle and testOptionalProductAttributeHidden into one method (test with no values, then with all values)

bojanz:
Committed in bdcd17f713bd2e29adff019d81e024b6f23b882b.
This commit actually revealed another bug (#options type), merged that fix into the commit.

PR #377, opened 2016-05-04
Issue #2716417 by mglaman: Allow rendering of variation fields on the…

mglaman:
… product entity

### DIFF
--- a/modules/product/commerce_product.module
+++ b/modules/product/commerce_product.module
@@ -10,6 +10,8 @@ use Drupal\field\Entity\FieldStorageConfig;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\commerce_product\Entity\ProductType;
 
 /**
  * Implements hook_entity_create().
@@ -68,6 +70,43 @@ function commerce_product_theme() {
       'render element' => 'elements',
     ],
   ];
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_view().
+ */
+function commerce_product_commerce_product_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
+  /** @var \Drupal\commerce_product\Entity\ProductInterface $entity */
+  $product_type = ProductType::load($entity->bundle());
+  if ($product_type->shouldInjectVariationFields()) {
+    /** @var \Drupal\commerce_product\ProductVariationFieldRendererInterface $variation_field_renderer */
+    $variation_field_renderer = \Drupal::service('commerce_product.variation_field_renderer');
+
+    $rendered_fields = $variation_field_renderer->renderFields($entity->getDefaultVariation(), $view_mode);
+    foreach ($rendered_fields as $field_name => $rendered_field) {
+      $build['product_variation_' . $field_name] = $rendered_field;
+    }
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_view().
+ */
+function commerce_product_commerce_product_variation_view(array &$build, \Drupal\Core\Entity\EntityInterface $entity, \Drupal\Core\Entity\Display\EntityViewDisplayInterface $display, $view_mode) {
+  $stop = null;
+}
+
+/**
+ * Implements hook_entity_display_build_alter().
+ */
+function commerce_product_entity_display_build_alter(&$build, $context) {
+  if ($context['entity'] instanceof \Drupal\commerce_product\Entity\ProductVariationInterface) {
+    foreach ($build as $key => &$item) {
+      $field_class = \Drupal::service('commerce_product.variation_field_renderer')->getAjaxReplacementClass($key, $context['entity']->getProductId());
+      $item['#attributes']['class'][] = $field_class;
+      $item['#ajax_replace_class'] = $field_class;
+    }
+  }
 }
 
 /**
--- a/modules/product/src/Entity/ProductType.php
+++ b/modules/product/src/Entity/ProductType.php
@@ -87,6 +87,13 @@ class ProductType extends ConfigEntityBundleBase implements ProductTypeInterface
   protected $variationType;
 
   /**
+   * Indicates if variation fields should be injected.
+   *
+   * @var bool
+   */
+  protected $injectVariationFields = TRUE;
+
+  /**
    * {@inheritdoc}
    */
   public function getDescription() {
@@ -113,6 +120,21 @@ class ProductType extends ConfigEntityBundleBase implements ProductTypeInterface
    */
   public function setVariationTypeId($variation_type_id) {
     $this->variationType = $variation_type_id;
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function shouldInjectVariationFields() {
+    return $this->injectVariationFields;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setInjectVariationFields($inject = TRUE) {
+    $this->injectVariationFields = (bool) $inject;
     return $this;
   }
 
--- a/modules/product/src/Entity/ProductVariation.php
+++ b/modules/product/src/Entity/ProductVariation.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\commerce_product\Entity;
 
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Entity\ContentEntityBase;
 use Drupal\Core\Entity\EntityChangedTrait;
 use Drupal\Core\Entity\EntityStorageInterface;
@@ -257,6 +258,18 @@ class ProductVariation extends ContentEntityBase implements ProductVariationInte
     $attribute_field_manager = \Drupal::service('commerce_product.attribute_field_manager');
     $field_map = $attribute_field_manager->getFieldMap($this->bundle());
     return array_column($field_map, 'field_name');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTagsToInvalidate() {
+    $tags = parent::getCacheTagsToInvalidate();
+    // Invalidate the variations view builder and product caches.
+    return Cache::mergeTags($tags, [
+      'commerce_product:' . $this->getProductId(),
+      'commerce_product_variation_view',
+    ]);
   }
 
   /**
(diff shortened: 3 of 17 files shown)


modules/product/src/Entity/ProductType.php
```diff
@@ -116,4 +123,19 @@ public function setVariationTypeId($variation_type_id) {
     return $this;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function hasInjectedVariationFields() {
```
bojanz: How about shouldInjectVariationFields / setInjectVariationFields() ?

modules/product/src/Entity/ProductVariation.php
```diff
@@ -262,6 +263,19 @@ protected function getAttributeFieldNames() {
...
    */
+  public function getCacheTagsToInvalidate() {
+    $tags = parent::getCacheTagsToInvalidate();
+    // Invalidate the variations view builder and product caches.
+    return Cache::mergeTags($tags, [
+      'commerce_product:' . $this->getProductId(),
+      'commerce_product_variation_view',
+    ]);
+  }
+
```
bojanz: Double newline.

modules/product/src/Tests/ProductVariationFieldInjectionTest.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace Drupal\commerce_product\Tests;
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
```
bojanz: Needs one newline before use, one less newline after use ;)

modules/product/tests/src/Kernel/VariationFieldInjectionTest.php
```diff
@@ -0,0 +1,139 @@
...
+use Drupal\commerce_product\Entity\Product;
+use Drupal\commerce_product\Entity\ProductVariation;
+use Drupal\commerce_product\Entity\ProductVariationType;
+use Drupal\Core\Field\FieldDefinitionInterface;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\KernelTests\KernelTestBase;
+
+/**
+ * Tests the attribute field manager.
```
bojanz: "Tests the attribute field manager." is a lie :)

modules/product/commerce_product.module
```diff
@@ -71,6 +71,24 @@ function commerce_product_theme() {
 }
 
 /**
+ * Implements hook_ENTITY_TYPE_view().
+ */
+function commerce_product_commerce_product_view(array &$build, \Drupal\Core\Entity\EntityInterface $entity, \Drupal\Core\Entity\Display\EntityViewDisplayInterface $display, $view_mode) {
```
agoradesign: Code style: import the namespaces instead of fully qualifiying here

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+    ]);
+  }
+
+  /**
+   * The allowed base field definitions for injection.
+   *
+   * @return array
+   *   An array of base field names.
+   */
+  protected function allowedBaseFields() {
```
bojanz: getAllowedBaseFields().

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+   *
+   * @param string $field_name
+   *   The field name.
+   * @param \Drupal\commerce_product\Entity\ProductVariationInterface $variation
+   *   The product variation.
+   *
+   * @return string
+   *   The CSS class.
+   */
+  protected function variationFieldCssClass($field_name, ProductVariationInterface $variation) {
```
bojanz: buildFieldClass(). Or inline it. It's a string join.

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+
+    $view_builder = $this->entityTypeManager->getViewBuilder('commerce_product_variation');
+
+    foreach ($field_definitions as $field_name => $field_definition) {
+      if ($view_display->getComponent($field_name)) {
+        $content = $view_builder->viewField($variation->$field_name, $view_mode);
+        $content['#attributes']['class'][] = $this->baseCssClass();
+
+        $content['#attributes']['class'][] = $this->variationFieldCssClass('variation_' . $field_name, $variation);
+        $content['#ajax_injection_class'] = $this->variationFieldCssClass('variation_' . $field_name, $variation);
```
agoradesign: very nitpicky and possibly a matter of taste, as the function call is a cheap one, but I'd prefer call the function only once, and re-use the variable instead

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+      $this->fieldDefinitions[$variation_type_id] = $definitions;
+    }
+
+    return $this->fieldDefinitions[$variation_type_id];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getRenderableFields(ProductVariationInterface $variation, $view_mode = 'default') {
```
bojanz: renderFields() perhaps? We'll also need a matching renderField($field_name, $variation, $view_mode) method, for the DS integration.

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+   * @param string $field_name
+   *   The field name.
+   * @param \Drupal\commerce_product\Entity\ProductVariationInterface $variation
+   *   The product variation.
+   *
+   * @return string
+   *   The CSS class.
+   */
+  protected function variationFieldCssClass($field_name, ProductVariationInterface $variation) {
+    return $this->baseCssClass() . '--' . $field_name . '__' . $variation->getProductId();
```
agoradesign: Not really a problem - is there a reason for mixing hyphens and underscores?

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+
+    return $injectable_fields;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addAjaxCommands(AjaxResponse $response, ProductVariationInterface $variation, $view_mode = 'default') {
```
bojanz: addAjaxCommands sounds too low level. What about replaceRenderedFields()?

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+   * The base CSS class all injected fields will have.
+   *
+   * This is also the base block class for injectable field classes that can be
+   * accessed via AJAX, follow BEM concepts.
+   *
+   * @return string
+   *   The CSS class.
+   */
+  protected function baseCssClass() {
+    return 'product--variation-field';
```
bojanz: Why not make this a constant on the class? Or just inline it? Doesn't feel worthy of its own method.

modules/product/src/VariationFieldInjection.php
agoradesign: I like the getter more, as it is not actually rendering the fileds, but collecting the renderable fields

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+use Drupal\Core\Ajax\ReplaceCommand;
+use Drupal\Core\Entity\EntityFieldManagerInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Field\BaseFieldDefinition;
+
+/**
+ * Default implementation of VariationFieldInjectionInterface.
+ */
+class VariationFieldInjection implements VariationFieldInjectionInterface {
```
bojanz: We definitely need Product in the class/interface names.

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+use Drupal\commerce_product\Entity\ProductVariationInterface;
+use Drupal\Core\Ajax\AjaxResponse;
+use Drupal\Core\Ajax\ReplaceCommand;
+use Drupal\Core\Entity\EntityFieldManagerInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Field\BaseFieldDefinition;
+
+/**
+ * Default implementation of VariationFieldInjectionInterface.
```
bojanz: I've started ommitting the docblock in these cases, cause "Default implementation of X" says absolutely nothing.

modules/product/src/VariationFieldInjection.php
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getRenderableFields(ProductVariationInterface $variation, $view_mode = 'default') {
+    $injectable_fields = [];
+
+    $field_definitions = $this->getFieldDefinitions($variation->bundle());
+
+    $view_display = $this->entityTypeManager
```
bojanz: Could we skip loading the view display? Shouldn't viewField() return something recognizable as empty if the field is hidden?

modules/product/src/VariationFieldInjection.php
agoradesign: In theory, the method provides more flexibility for extending the class and override only this part. In practice this example won't probably never happen. Basically, for me it doesn't matter. But if I had to decide, I would leave the method, how it is

modules/product/src/VariationFieldInjection.php
mglaman: I'll probably inline it. Idea was someone could override the service to change the class. Probably minor edge case.

agoradesign: Like I've already commented it at baseCssClass(): It's really an edge case. But why not keep and only rename it?

modules/product/commerce_product.module
```diff
@@ -71,6 +73,24 @@ function commerce_product_theme() {
...
+  /** @var \Drupal\commerce_product\Entity\ProductInterface $entity */
+
+  $product_type = ProductType::load($entity->bundle());
+  if ($product_type->shouldInjectVariationFields()) {
+    /** @var \Drupal\commerce_product\ProductVariationFieldRendererInterface $variation_field_injector */
+    $variation_field_injector = \Drupal::service('commerce_product.variation_field_renderer');
+
+    $injection_fields = $variation_field_injector->getRenderableFields($entity->getDefaultVariation(), $view_mode);
+    foreach ($injection_fields as $field_name => $injection_field) {
+      $build[$field_name] = $injection_field;
```
bojanz: We said we'd prefix the entries in $build with 'variation_' to prevent collisions and make it more obvious where the fields came from?

agoradesign: Strange, I could swear that I've already seen the prefix on review before!?!

bojanz:
Committed in 6f4ee9b797da9677cb6f312a80f41962e25df671.

PR #379, opened 2016-05-08
2721147 added small if statement

haringsrob:


bojanz:
You don't want to wrap an entire method in an IF statement, the resulting indentation makes code hard to read. Instead, return early (if there's no order)

bojanz:
Pushed a fix in 61ed1dbeae48ab4f61097257dd3b4ccdfa1b934e.

PR #380, opened 2016-05-09
2716241 Registration on login pane

haringsrob:
This is a proposal to have the registration form available on the login step.

I added an extra option to the pane config, so that the site builder can still choose not to show the form.

I also added the user_login hook, to redirect after the actual registration.

### DIFF
--- a/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
+++ b/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
@@ -261,18 +261,18 @@ class Login extends CheckoutPaneBase implements CheckoutPaneInterface, Container
         ],
       ],
     ];
-    $pane_form['register']['mail'] = array(
+    $pane_form['register']['mail'] = [
       '#type' => 'email',
       '#title' => $this->t('Email address'),
       '#description' => $this->t('A valid email address. All emails from the system will be sent to this address. The email address is not made public and will only be used if you wish to receive a new password or wish to receive certain news or notifications by email.'),
       '#required' => FALSE,
-    );
-    $pane_form['register']['pass'] = array(
+    ];
+    $pane_form['register']['pass'] = [
       '#type' => 'password_confirm',
       '#size' => 25,
       '#description' => $this->t('Provide a password for the new account in both fields.'),
       '#required' => FALSE,
-    );
+    ];
     $pane_form['register']['register'] = [
       '#type' => 'submit',
       '#value' => $this->t('Create account and continue'),
@@ -340,11 +340,12 @@ class Login extends CheckoutPaneBase implements CheckoutPaneInterface, Container
 
         // Advanced validation Make sure the account does not exist yet. And
         // that the username is unused/valid.
-        if ($this->entityTypeManager->getStorage('user')->loadByProperties(['mail' => $values['register']['mail']])) {
+        $user_storage = $this->entityTypeManager->getStorage('user');
+        if ($user_storage->loadByProperties(['mail' => $values['register']['mail']])) {
           $form_state->setErrorByName('mail', $this->t('A user is already registered with this email.'));
           return;
         }
-        if ($this->entityTypeManager->getStorage('user')->loadByProperties(['name' => $values['register']['mail']])) {
+        if ($user_storage->loadByProperties(['name' => $values['register']['mail']])) {
           $form_state->setErrorByName('mail', $this->t('A user is already registered with this username, please contact support to resolve this issue.'));
           return;
         }
@@ -389,9 +390,6 @@ class Login extends CheckoutPaneBase implements CheckoutPaneInterface, Container
         user_login_finalize($account);
         $this->order->setOwner($account);
         $this->credentialsCheckFlood->clearAccount($this->clientIp, $account->getAccountName());
-        break;
-
-      case 'continue':
         break;
 
       default:
--- a/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
+++ b/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
@@ -254,6 +254,12 @@ class Login extends CheckoutPaneBase implements CheckoutPaneInterface, Container
       '#type' => 'fieldset',
       '#title' => $this->t('Create new account'),
       '#access' => !$this->configuration['allow_guest_checkout'] && $this->configuration['show_registration_form'],
+      '#attributes' => [
+        'class' => [
+          'form-wrapper__login-option',
+          'form-wrapper__guest-checkout',
+        ],
+      ],
     ];
     $pane_form['register']['mail'] = array(
       '#type' => 'email',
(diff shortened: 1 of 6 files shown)


modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -76,17 +77,23 @@ class Login extends CheckoutPaneBase implements CheckoutPaneInterface, Container
...
    *   The request stack.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, CheckoutFlowInterface $checkout_flow, CredentialsCheckFloodInterface $credentials_check_flood, AccountInterface $current_user, EntityTypeManagerInterface $entity_type_manager, UserAuthInterface $user_auth, RequestStack $request_stack) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, CheckoutFlowInterface $checkout_flow, CredentialsCheckFloodInterface $credentials_check_flood, AccountInterface $current_user, EntityTypeManagerInterface $entity_type_manager, EntityManagerInterface $entity_manager, EntityFormBuilderInterface $entity_form_builder, UserAuthInterface $user_auth, RequestStack $request_stack) {
     parent::__construct($configuration, $plugin_id, $plugin_definition, $checkout_flow);
 
     $this->credentialsCheckFlood = $credentials_check_flood;
     $this->currentUser = $current_user;
     $this->entityTypeManager = $entity_type_manager;
+    $this->entityManager = $entity_manager;
```
bojanz: Why inject the deprecated entityManager when you already have the non-deprecated entityTypeManager?

bojanz:
Thanks for getting this rolling! The embedding of the entire existing registration form looks a bit messy, it would probably be better to copy over the AccountForm / RegistrationForm logic (but without the "needs admin approval" and the pass reset / preferred language stuff)

modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
haringsrob: I will change this, after I created this, I worked on an issue about the same in core.. But I forgot I used it here.

haringsrob:
Bojanz:

I wanted to do that, but I feel that that would not be the correct way to go. What if extra other modules modify the registration form? We would then make it impossible to use them? Or am I missing something?

haringsrob:
Also, If we keep it in one form, and we would add required fields, they would become part of the same js validation (tried this first), when you then would submit the login form, you would be blocked by the registration fields.

bojanz:
Both of these dilemmas are the same as for login.

You don't mark any field as required, then validate the fields if one of them was provided.
The context is sufficiently different that I'm not worried about the regular alter hooks not working (due to a different form ID).

haringsrob:
@bojanz, if we do modify, then I assume the following fields are enough (no timezone etc,...):
schermafbeelding 2016-05-10 om 07 34 00

haringsrob:
We do need an option to auto generate the username i think, as email + username is not user friendly.

haringsrob:
@bojanz, updated the PR, waiting for feedback.

modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -212,6 +236,29 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+      '#title' => $this->t('Create new account'),
+      '#access' => !$this->configuration['allow_guest_checkout'] && $this->configuration['show_registration_form'],
+    ];
+    $pane_form['register']['mail'] = array(
+      '#type' => 'email',
+      '#title' => $this->t('Email address'),
+      '#description' => $this->t('A valid email address. All emails from the system will be sent to this address. The email address is not made public and will only be used if you wish to receive a new password or wish to receive certain news or notifications by email.'),
+      '#required' => FALSE,
+    );
+    $pane_form['register']['pass'] = array(
```
bojanz: Long array usage.

modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -212,6 +236,29 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
       '#op' => 'continue',
     ];
 
+    $pane_form['register'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Create new account'),
+      '#access' => !$this->configuration['allow_guest_checkout'] && $this->configuration['show_registration_form'],
+    ];
+    $pane_form['register']['mail'] = array(
```
bojanz: Long array usage.

modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -220,53 +267,115 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+          return;
+        }
 
-    $uid = $this->userAuth->authenticate($username, $password);
-    if (!$uid) {
-      $this->credentialsCheckFlood->register($this->clientIp, $username);
-      $form_state->setErrorByName('name', $this->t('Unrecognized username or password.'));
+        // Advanced validation Make sure the account does not exist yet. And
+        // that the username is unused/valid.
+        if ($this->entityTypeManager->getStorage('user')->loadByProperties(['mail' => $values['register']['mail']])) {
```
bojanz: Why not get the $user_storage once then use it in both IFs?

modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -220,53 +267,115 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+      case 'register':
+        $storage = $this->entityTypeManager->getStorage('user');
+        /** @var \Drupal\user\UserInterface $account */
+        $account = $storage->load($form_state->get('logged_in_uid'));
+        user_login_finalize($account);
+        $this->order->setOwner($account);
+        $this->credentialsCheckFlood->clearAccount($this->clientIp, $account->getAccountName());
+        break;
+
+      case 'continue':
```
bojanz: Why have both the continue and the default cases, which do the same thing? Kill one

haringsrob:
I'm working on this.

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
```diff
@@ -96,4 +96,30 @@ public function testGuestOrderCheckout() {
...
+    $config->save();
+
+    $this->drupalLogout();
+    $this->drupalGet($this->product->toUrl()->toString());
+    $this->submitForm([], 'Add to cart');
+    $cart_link = $this->getSession()->getPage()->findLink('your cart');
+    $cart_link->click();
+    $this->submitForm([], 'Checkout');
+    $this->assertSession()->pageTextContains('Create new account');
+    $this->submitForm([
```
mglaman, line 110: Can we also test validation? Existing email, invalid email

haringsrob:
I think I have covered all the validations now.

PR #381, opened 2016-05-09
Issue #2721763: Undefined variable in commerce/modules/price/commerce…

nkgokul:
The reason we are getting this error we are getting one set of data for the revisions as well, which was creating a flood of notices on watchdog. This should fix the notices for now. 

We should also consider how to get this working with revisions as well.


bojanz:
Merged. No commerce entity types are currently revisionable, we'll need to revisit that before beta1.

PR #382, opened 2016-05-12
2723369 added maxlength - 11 for the system name

haringsrob:


ndf:
Hi Rob, I like descriptive machinenames, the longer the better.
How do you think about changing the prefix from "attribute_" to "attr_" ? It gives us 5 extra chars.

haringsrob:
Hi @ndf,

Could be done yes, but let's see what @bojanz or @mglaman  thinks about this.

When I can catch him on irc I'll ask.

ndf:
Nice, pull-request looks straight-forward to me!

mglaman:
How long are your attribute names you'd need them to be this long? Typically it would be color, length, size, etc. I don't see why we would need to use `attr` over `attribute`.

bojanz:
Yeah, I'm not convinced with 'attr_'. Let's go with the 22 char limitation for now.

Committed an improved fix in 70b2ecd85d380dd22236de198398081870577e76.

PR #385, opened 2016-05-13
Issue #2723697: ProductVariationAttributesWidget::ajaxRefresh() needs…

mglaman:
… to pass along the product view mode


modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -342,7 +342,7 @@ public static function ajaxRefresh(array $form, FormStateInterface $form_state)
     $variation = ProductVariation::load($form_state->get('selected_variation'));
     /** @var \Drupal\commerce_product\ProductVariationFieldRendererInterface $variation_field_renderer */
     $variation_field_renderer = \Drupal::service('commerce_product.variation_field_renderer');
-    $view_mode = 'default';
+    $view_mode = $form_state->getStorage()['form_display']->getMode();
```
bojanz: Ah good, didn't knwo the form display is already in form state.
Do we need to use getStorage() instead of just doing $form_state->get('form_display')?

mglaman: Yeah, unless we did `->get('storage')['form_display']`. `get` method is for direct child properties.

PR #388, opened 2016-05-25
Issue #2733061 by mglaman: Test the checkout as a guest

mglaman:


modules/checkout/tests/src/Functional/CheckoutOrderTest.php
```diff
@@ -0,0 +1,157 @@
...
+    $this->saveHtmlOutput();
+  }
+
+  /**
+   * Creates and sets the default store.
+   *
+   * @return \Drupal\commerce_store\Entity\StoreInterface
+   *    The store.
+   */
+  protected function createDefaultStore() {
```
mglaman: I'd like to get this committed, then move to an attribute in `Drupal\Tests\commerce_store\CreateStoreAttribute` or something

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
```diff
@@ -0,0 +1,157 @@
...
+    $store->save();
+    $store_storage->markAsDefault($store);
+
+    return $store;
+  }
+
+  /**
+   * Debugger method to save additional HTML output.
+   */
+  protected function saveHtmlOutput() {
```
mglaman: When using `--printer="\Drupal\Tests\Listeners\HtmlOutputPrinter"` option it writes verbose output. However not on fail. Ending the test with this helped debug errors. We can remove, however, it's been very helpful in writing BrowerTests based tests. It might belong in a CommerceBrowserTestBase if we make one.

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
```diff
@@ -0,0 +1,157 @@
...
+use Drupal\simpletest\BlockCreationTrait;
+use Drupal\Tests\BrowserTestBase;
+use Drupal\user\RoleInterface;
+
+/**
+ * Tests the checkout of an order.
+ *
+ * @group commerce
+ *
+ * @runTestsInSeparateProcesses
```
bojanz: @runTestsInSeparateProcesses
and
@preserveGlobalState disabled
are defaults, so you don't need to specify them.

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
bojanz: Let's document that and keep it then.

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
mglaman: I thought we had to explicitly add it for browser tests

PR #389, opened 2016-05-25
add tokens to commerce_product

gnoddep:
Add tokens to the commerce_product entity, so pathauto can use them to create URL aliases. The code is based on node.tokens.inc from the core node module.

### DIFF
--- a/modules/product/commerce_product.tokens.inc
+++ b/modules/product/commerce_product.tokens.inc
@@ -1,0 +1,198 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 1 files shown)


bojanz:
Please see https://www.drupal.org/node/2684875
We are working on fixing Token module to provide tokens for all entity types, including products.
That's the way to go.
I'll keep this PR open in the meantime to allow people to use it as a temporary solution.

bojanz:
https://www.drupal.org/node/2493567 has been committed to Token 8.x-1.x, these tokens are now provided automatically.

PR #391, opened 2016-05-25
Issue #2733763: CommerceBrowserTestBase that extends BrowserTestBase

mglaman:


tests/src/Functional/CommerceBrowserTestBaseTest.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Drupal\Tests\commerce\Functional;
+
+/**
+ * Tests CommerceBrowserTestBase functionality.
+ *
+ * @group commerce
+ */
+class CommerceBrowserTestBaseTest extends CommerceBrowserTestBase {
```
bojanz: This test feels a bit too meta for me. I know 1.x did it too, but it doesn't make much sense. If the code for creating a user or an entity fails, then literally every test will fail, so we'll know that the test base is broken.

tests/src/Functional/CommerceBrowserTestBase.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Drupal\Tests\commerce\Functional;
+
+use Drupal\Component\Render\FormattableMarkup;
+use Drupal\Tests\BrowserTestBase;
+
+/**
+ * Base class for Drupal Commerce (and contrib) to test with BrowserTest.
```
bojanz: "Provides a base class for Commerce functional tests."
(Drupal Commerce is just Commerce throughout the code, the contrib part is assumed)

tests/src/Functional/CommerceBrowserTestBaseTest.php
mglaman: All right, makes sense. Core did it too for base test, so I went "what the hell, why not"

PR #392, opened 2016-05-26
Issue #2733795: ProductTestTrait for products, variations, and attrib…

mglaman:
…ute testing

### DIFF
--- a/modules/product/src/ProductTestTrait.php
+++ b/modules/product/src/ProductTestTrait.php
@@ -1,0 +1,162 @@
+<?php
+
+namespace Drupal\commerce_product;
+
+use Drupal\commerce_product\Entity\ProductAttribute;
+use Drupal\commerce_product\Entity\ProductAttributeValue;
+use Drupal\commerce_product\Entity\ProductVariation;
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+
+/**
+ * Provides methods for working with products, variations, and attributes.
+ *
+ * This trait is meant to be used only by test classes.
+ *
+ * @todo Move to \Drupal\Tests\commerce_product post-SimpleTest.
+ */
+trait ProductTestTrait {
+
+  /**
+   * Creates a set of product variations.
+   *
+   * The $values_matrix parameter contains the price and attributes information.
+   *
+   * Example:
+   * [
+   *   [
+   *      'price' => 9.99
+   *      'attribute_color' => 2,
+   *      'attribute_size' => 10,
+   *   ],
+   *   [
+   *      'price' => 9.99
+   *      'attribute_color' => 2,
+   *      'attribute_size' => 6,
+   *   ],
+   * ]
+   *
+   * @param string $type
+   *   The variation type.
+   * @param array $values_matrix
+   *   Array of values.
+   * @param string $currency_code
+   *   The currency code.
+   *
+   * @return \Drupal\commerce_product\Entity\ProductVariationInterface[]
+   *   Array of created variations
+   */
+  protected function createProductVariations($type = 'default', $values_matrix = [], $currency_code = 'USD') {
+    $variations = [];
+    foreach ($values_matrix as $item) {
+      $variation_values = [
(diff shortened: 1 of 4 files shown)


modules/product/src/ProductTestTrait.php
```diff
@@ -0,0 +1,163 @@
...
+ * Provides methods for working with products, variations, and attributes.
+ *
+ * This trait is meant to be used only by test classes.
+ *
+ * @todo Move to \Drupal\Tests\commerce_product post-SimpleTest.
+ */
+trait ProductTestTrait {
+
+  /**
+   * Creates a set of product cariations.
```
bojanz: s/cariations/variations :)

modules/product/src/ProductTestTrait.php
```diff
@@ -0,0 +1,163 @@
...
+          $variation_values['price']['amount'] = $value;
+        }
+        else {
+          $variation_values[$key] = $value;
+        }
+      }
+
+      $variation = ProductVariation::create($variation_values);
+      $variation->save();
+      $variations[] = ProductVariation::load($variation->id());
```
bojanz, line 72: We can now use the base ->createEntity() which does the reload for us (and documents why)

modules/product/src/ProductTestTrait.php
```diff
@@ -0,0 +1,163 @@
...
+   * @param string $currency_code
+   *   The currency code.
+   *
+   * @return \Drupal\commerce_product\Entity\ProductVariationInterface[]
+   *   Array of created variations
+   */
+  protected function createProductVariations($type = 'default', $values_matrix = [], $currency_code = 'USD') {
+    $variations = [];
+    foreach ($values_matrix as $item) {
+
```
bojanz: Extra newline.

modules/product/src/ProductTestTrait.php
```diff
@@ -0,0 +1,163 @@
...
+   *   The variation type.
+   * @param array $values_matrix
+   *   Array of values.
+   * @param string $currency_code
+   *   The currency code.
+   *
+   * @return \Drupal\commerce_product\Entity\ProductVariationInterface[]
+   *   Array of created variations
+   */
+  protected function createProductVariations($type = 'default', $values_matrix = [], $currency_code = 'USD') {
```
bojanz, line 50: Feels odd to have $currency_code, this belongs in $values_matrix.

It's inconsistent that we have createStore() and createAttributeValue(), but createProductVariations(). Let's do createProductVariation() and the implementing tests can do a foreach.

modules/product/src/ProductTestTrait.php
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  /**
+   * Creates an attribute field and set of attribute values.
+   *
+   * @param string $variation_type
+   *   The variation type.
+   * @param string $name
+   *   The attribute field name.
+   * @param array $options
```
bojanz: $options should be $values (cause they're attribute values).

modules/product/src/ProductTestTrait.php
```diff
@@ -0,0 +1,163 @@
...
+   *   Associative array of key name values. [red => Red].
+   * @param bool $test_field
+   *   Flag to create a test field on the attribute.
+   *
+   * @return \Drupal\commerce_product\Entity\ProductAttributeValueInterface[]
+   *   Array of attribute entities.
+   */
+  protected function createAttributeSet($variation_type, $name, array $options, $test_field = FALSE) {
```
bojanz: This should just be called createAttributeSet(), since it only creates a single attribute, not several.

modules/product/src/ProductTestTrait.php
mglaman, line 72: But this is a trait, was wondering what we should do.

modules/product/src/ProductTestTrait.php
mglaman, line 50: Makes sense, because we still need to run `addVariation` since there isn't a "addVariations" to merge in an array.

modules/product/src/ProductTestTrait.php
mglaman: But it is?

PR #395, opened 2016-05-28
Issue #2735737: Convert CurrencyTest to BrowserTest

mglaman:


modules/price/tests/src/Functional/CurrencyTest.php
```diff
@@ -1,66 +1,79 @@
...
     'commerce_store',
   ];
 
   /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    $this->placeBlock('local_tasks_block');
```
bojanz: Why doesn't CommerceBrowserTestBase place the blocks like CommerceTestBase does? Seems common enough

PR #397, opened 2016-06-12
#2742205 Port commerce_cart tests to BrowserTest

smmccabe:
Split out a JavascriptTestTrait which can probably be useful elsewhere
Squash this when you merge in, I did lots of little commits so it's a bit of a mess


smmccabe:
Cleaned up into 1 commit

src/JavascriptTestTrait.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Drupal\commerce;
```
mglaman, line 3: Perhaps this should live under Drupal\Tests\commerce\FunctionalJavascript since it will only be used by other modules in that kind of namespace? @bojanz what are your thoughts? Being over picky? I know I had some test base items in just `Drupal\Tests\<module>`. But since this trait is only to be used for functional Javascript tests, maybe we should put it there.

modules/cart/tests/src/FunctionalJavascript/AddToCartMultiAttribute.php
```diff
@@ -0,0 +1,158 @@
...
+use Drupal\commerce_product\Entity\ProductVariation;
+use Drupal\commerce_product\Entity\ProductVariationType;
+use Drupal\Tests\commerce_cart\CartBrowserTestBase;
+
+/**
+ * Tests the add to cart form.
+ *
+ * @group commerce
+ */
+class AddToCartMultiAttribute extends CartBrowserTestBase {
```
mglaman, line 16: +1 on splitting this into its own class which adds JavaScript driver.

src/JavascriptTestTrait.php
mglaman, line 3: @smmccabe on another note, I wanted to make this a core patch. Do you want the honors?

smmccabe, line 3: @mglaman like core core? I thought you just meant commerce core :P Is there an existing issue or anything for core?

smmccabe, line 3: @mglaman Yah I wasn't sure where to put that, after writing a bit more that didn't need javascript, I'm thinking your suggestion above does make the most sense. @bojanz give official blessing?

mglaman:
Waiting for Travis to complete on my fork. Did the clean up on my end.

mglaman:
Thanks @smmccabe! Committed in @16c3a1a44209b472d4f0a2f6e0eb29a7d799ad1f with changes.

PR #398, opened 2016-06-13
#2742201 Port TaxTypeTest to BrowserTest

smmccabe:


mglaman:
Can you `rebase -i` and force push to squash the commits

smmccabe:
Updated, used commit message from issue queue

modules/tax/tests/src/TaxBrowserTestBase.php
```diff
@@ -1,13 +1,13 @@
 <?php
 
-namespace Drupal\commerce_tax\Tests;
+namespace Drupal\Tests\commerce_tax;
```
bojanz: The new location should be tests/src/Functional, namespace Drupal\Tests\commerce_tax\Functional, since we're extending the functional test base.

mglaman: All PHPUnit based tests go in Drupal\Tests<module\Unit|Kernel|Functional|FunctionalJavascript

smmccabe: Well it's not a test itself, it is just the base for tests, like OrderBrowserBaseTest, which I copied the location from, is OrderBrowserBaseTest wrong as well?

mglaman: Then yeah, my other test is wrong, I actually thought about that last night.

smmccabe:
Updated and branch passes my personal travis, should be good to go.

PR #399, opened 2016-06-13
Issue #2734627 by agoradesign, mglaman: Product entities without an a…

mglaman:
…ctive variation cannot be rendered, causing a fatal error


mglaman:
@bojanz can I get your look over for fix and the test

PR #400, opened 2016-06-13
Issue #2734259 by Sagar Ramgade, mglaman: Cart page requires Node mod…

mglaman:
…ule for 'access content' permission

### DIFF
--- a/modules/cart/commerce_cart.permissions.yml
+++ b/modules/cart/commerce_cart.permissions.yml
@@ -1,0 +1,3 @@
+access cart:
+  title: 'Access cart'
+  description: 'View the cart page.'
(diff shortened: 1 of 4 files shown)


mglaman:
@bojanz wondering your thoughts on this fix.

bojanz:
Back for 1.x we talked about having our own permission for the cart. Might be a good time to introduce it?

mglaman:
Maybe. But I feel like people would expect users to "just get" access to the cart page. But the same happens for checkout, I suppose. 

I think it makes sense to implement a custom permission then, since we have no dependency on Node, unlike 1.x

mglaman:
@bojanz okay, rebased (was missing commerce_cart port to BrowserTest.) Added new permission to route, updated tests.

PR #404, opened 2016-06-14
Issue #2748697: Product View doesn't update when Variation Type Displ…

smmccabe:
If you change the display of a variation, say by turning the SKU display off, it will not reflect on the product page, even after a refresh. If you change and attribute and trigger the ajax, it will correct itself since the AJAX is not cached. 

The problem is the commerce_product_view tag is not invalidated because as far as Drupal is concerned, nothing on the "product" has changed. I am unsure if the variation should be invalidating the cache or if the product should be setting a different cache tag.

### DIFF
--- a/modules/product/commerce_product.module
+++ b/modules/product/commerce_product.module
@@ -5,6 +5,7 @@
  * Defines the Product entity and associated features.
  */
 
+use Drupal\Core\Cache\Cache;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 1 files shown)


modules/product/commerce_product.module
```diff
@@ -44,6 +44,17 @@ function commerce_product_entity_form_display_update(EntityFormDisplayInterface
 }
 
 /**
+ * Implements hook_ENTITY_TYPE_update().
+ */
+function commerce_product_entity_view_display_update(Drupal\Core\Entity\EntityInterface $entity) {
+  // The product view uses the variation view and needs to be cleared, which doesn't happen automatically
+  // because we're editing the variation, not the product
+  if ($entity->getConfigTarget() == 'commerce_product_variation.default.default') {
```
mglaman: Shouldn't we target any `commerce_product_variation.*.*`? As you can embed variation displays in multiple cases with different view modes.

smmccabe:
something like this? just matching the first part, or should i regex for the ._._? i guess maybe i could also just include the first .?

smmccabe:
after some thought, changed to 'commerce_product_variation." to be a little more specific. Should be ready to merge.

mglaman:
@bojanz looks good to me, can I get your eyes quick? This blocks all of the final test ports.

PR #408, opened 2016-06-17
Issue #2749287: Implement a Calculator class and use it in the Price value object

FrancisBaileyH:


bojanz:
Thanks! Do you mind filling out the docblocks? We need a description for each method, param, return value.

FrancisBaileyH:
@bojanz I've gone ahead and filled out the doc blocks.

FrancisBaileyH:
Fixed two docblocks with improper spacing.

modules/price/src/CurrencyMismatchException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Drupal\commerce_price;
+
+/**
+ * This exception is thrown when attempting to perform math using two different currencies.
+ * For example, adding a price in USD to another in EUR.
+ */
+class CurrencyMismatchException extends \RuntimeException {
```
mglaman: @bojanz should we have an Exception namespace for these?

FrancisBaileyH: That's an easy change so let me know if this is something you'd like implemented.

FrancisBaileyH:
@bojanz @mglaman any updates on this pr?

modules/price/src/Price.php
```diff
@@ -63,4 +64,214 @@ public function __toString() {
...
 
+  /**
+   * Convert a \Drupal\commerce_price\Price object to a different currency.
+   *
+   * @param string $currency
+   *   A currency code.
+   * @param string $rate
+   *   A currency rate corresponding to the currency code.
+   *
+   * @return Price
```
bojanz: Any object name given as a return or param hint must be fully qualified (have the full namespace).

modules/price/src/Price.php
```diff
@@ -63,4 +64,214 @@ public function __toString() {
...
+   *   A new \Drupal\commerce_price\Price instance.
```
bojanz: You don't need to use the full namespace in the descriptions.

modules/price/src/CurrencyMismatchException.php
bojanz: Yes, we should. Follows the example of other Exception classes in Commerce.

modules/price/src/InvalidArgumentException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Drupal\commerce_price;
+
+class InvalidArgumentException extends \InvalidArgumentException {
```
bojanz: We can remove this one. It doesn't add anything over the builtin \InvalidArgumentException.

FrancisBaileyH:
@bojanz okay I applied your suggested changes.

bojanz:
Committed in 071d9ba6ce028d4e3a260082e997ded81d842e47 after some bike shedding :)

PR #409, opened 2016-06-21
Issue: #2708455: Cart totals are not updated on updating line items

robincee:
https://www.drupal.org/node/2708455

### DIFF
--- a/modules/cart/src/CartManager.php
+++ b/modules/cart/src/CartManager.php
@@ -110,6 +110,7 @@ class CartManager implements CartManagerInterface {
       $new_quantity = $matching_line_item->getQuantity() + $quantity;
       $matching_line_item->setQuantity($new_quantity);
       $matching_line_item->save();
+      $needs_cart_save = TRUE;
     }
     else {
       $line_item->save();
@@ -135,6 +136,13 @@ class CartManager implements CartManagerInterface {
     $line_item->save();
     $event = new CartLineItemUpdateEvent($cart, $line_item, $original_line_item);
     $this->eventDispatcher->dispatch(CartEvents::CART_LINE_ITEM_UPDATE, $event);
+    // Let's *always* update the order if a line item is changed.
+    //
+    // The only value you can change by default is the amount via the cart view.
+    // Once this changes, a cart update is required for the total_price__amount
+    // db column. We do not, like the addLineItem() function, test for 
+    // quantities and a possible difference, we expect we need to update it.
+    $cart->save();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


modules/cart/src/CartManager.php
```diff
@@ -110,6 +110,7 @@ public function addLineItem(OrderInterface $cart, LineItemInterface $line_item,
       $new_quantity = $matching_line_item->getQuantity() + $quantity;
       $matching_line_item->setQuantity($new_quantity);
       $matching_line_item->save();
+      $needs_cart_save = TRUE;
```
mglaman, line 113: I don't see this boolean checked elsewhere?

modules/cart/src/CartManager.php
```diff
@@ -135,6 +136,13 @@ public function updateLineItem(OrderInterface $cart, LineItemInterface $line_ite
     $line_item->save();
     $event = new CartLineItemUpdateEvent($cart, $line_item, $original_line_item);
     $this->eventDispatcher->dispatch(CartEvents::CART_LINE_ITEM_UPDATE, $event);
+    // Let's *always* update the order if a line item is changed.
+    //
+    // The only value you can change by default is the amount via the cart view.
+    // Once this changes, a cart update is required for the total_price__amount
+    // db column. We do not, like the addLineItem() function, test for
+    // quantities and a possible difference, we expect we need to update it.
```
mglaman: Can be simplified to "We need to recalculate the order's total"

robincee: agreed, but figured: now it's clear for the review. I'll update it to "We need to recalculate the order's total".

modules/cart/src/CartManager.php
robincee, line 113: scroll down a bit. you should see this just below the else:

```
if ($needs_cart_save && $save_cart) {
  $cart->save();
}
```

https://github.com/drupalcommerce/commerce/blob/8.x-2.x/modules/cart/src/CartManager.php#L122

modules/cart/src/CartManager.php
```diff
@@ -135,6 +136,8 @@ public function updateLineItem(OrderInterface $cart, LineItemInterface $line_ite
     $line_item->save();
     $event = new CartLineItemUpdateEvent($cart, $line_item, $original_line_item);
     $this->eventDispatcher->dispatch(CartEvents::CART_LINE_ITEM_UPDATE, $event);
+    // We need to recalculate the order's total when updating line items.
+    $cart->save();
```
mglaman, line 140: Another thought, do we need to run save before passing to the event? Otherwise the `$cart` won't have an update order total.

I almost feel like we need a method on the order entity to refresh the order total without saving. @bojanz thoughts?

modules/cart/src/CartManager.php
mglaman, line 113: Whoops, sorry

bojanz:
Merged a different fix in #548

PR #410, opened 2016-06-21
Issue #2747911 Port UninstallTest to BrowserTest

czigor:


bojanz:
I fear that converting this test to kernel test will actually ruin its purpose, cause kernel tests don't perform actual module install / uninstall. The test was added to test Commerce not being reinstallable because some shipped config wasn't removed on uninstall. I am not sure kernel tests can catch that, we'd need to test that (or just perform these calls inside a functional test instead)

mglaman:
@bojanz agree; updated issue title to reflect porting it to BrowserTest

czigor:
New PR created.

PR #416, opened 2016-06-22
Issue 2724059: allow quantity in shopping cart block to be plural

PieterJanPut:


mglaman:
As commented in D.o issue: We need to test this via a Functional test :)

mglaman:
Closed in favor of #448, which has tests. Same commit credit.

PR #419, opened 2016-06-22
Issue #2753831 by mglaman: Add the ability to throw a redirect exception

mglaman:


mglaman:
Captain @bojanz, awaiting your mark.

commerce.services.yml
```diff
@@ -47,3 +47,9 @@ services:
     arguments: ['@commerce.country_context']
     tags:
       - { name: cache.context}
+
+  commerce.secure_redirect_response_subscriber:
```
bojanz: Should be "commerce.redirect_response_subscriber" to match the class name?

src/EventSubscriber/RedirectResponseSubscriber.php
```diff
@@ -0,0 +1,39 @@
...
+   *
+   * @param \Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent $event
+   *   The event to process.
+   */
+  public function onRedirectResponse(GetResponseForExceptionEvent $event) {
+    $exception = $event->getException();
+    if ($exception instanceof RedirectResponse) {
+      $event->setResponse($exception->getHttpResponseObject());
+    }
+
```
bojanz: Extra newline.

src/Response/RedirectResponse.php
```diff
@@ -0,0 +1,93 @@
...
+   * @param string $url
+   *    The URL to redirect to.
+   * @param int $status_code
+   *    The redirect status code.
+   * @param string[] $headers
+   *    Headers to pass with the redirect.
+   */
+  public function __construct($url, $status_code = 302, array $headers = []) {
+    if (!UrlHelper::isValid($url)) {
+      throw new \InvalidArgumentException('Invalid URL for secure redirect');
```
bojanz: Maybe just 'Invalid URL provided.'? Or "Invalid url provided to $class."? The backtrace would provide the rest.

src/Response/RedirectResponse.php
```diff
@@ -0,0 +1,93 @@
...
+
+namespace Drupal\commerce\Response;
+
+use Drupal\Component\Utility\UrlHelper;
+use Drupal\Core\Routing\TrustedRedirectResponse;
+
+/**
+ * Provides an exception that represents the need for an HTTP redirect.
+ */
+class RedirectResponse extends \LogicException {
```
bojanz: I wonder if we can name this class better. Drupal suffixes all Exceptions with Exception. And even our own subscriber is clarifying that the redirect response is actually an exception: "Check for RedirectResponse which extends Exception.".

Maybe we can call it NeedsRedirectException? Or just RedirectException?

src/Response/RedirectResponse.php
```diff
@@ -0,0 +1,93 @@
...
+    return $this->headers;
+  }
+
+  /**
+   * Get HTTP redirect response object.
+   *
+   * @return \Drupal\Core\Routing\TrustedRedirectResponse
+   *   The response object.
+   */
+  public function getHttpResponseObject() {
```
bojanz: If we rename the class, this method can be called just getResponse() or getRedirectResponse().

bojanz:
This looks great! Love the tests. Let's tweak the naming a bit and merge it.

bojanz:
We agreed on RedirectSubscriber / NeedsRedirectException / getRedirectResponse()

tests/modules/commerce_redirect_test/src/Form/RedirectTestForm.php
```diff
@@ -0,0 +1,35 @@
...
+   */
+  public function getFormId() {
+    return 'commerce_redirect_test_form';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    throw new NeedsRedirectException('https://www.drupal.org');
```
timplunkett: I'm 90% sure that `return new RedirectResponse('https://www.drupal.org');` will work here.

mglaman: Thanks @timplunkett! I updated our test to use that, and mention we do not need to use our method

src/Response/NeedsRedirectException.php
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * The headers to set.
+   *
+   * @var string[]
+   */
+  protected $headers;
+
+  /**
+   * Creates a new NeedsRedirectException object.
```
bojanz: Constructs a new... I know, I know, I'm terrible :)

src/Response/NeedsRedirectException.php
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * Get HTTP redirect response object.
+   *
+   * @return \Drupal\Core\Routing\TrustedRedirectResponse
+   *   The response object.
+   */
+  public function getRedirectResponse() {
+    return new TrustedRedirectResponse($this->getUrl(), $this->getStatusCode(), $this->getHeaders());
```
bojanz: The parent class already has a getResponse(), makes more sense to make the response object directly in the constructor.

src/Response/NeedsRedirectException.php
```diff
@@ -0,0 +1,96 @@
...
+    parent::__construct($this->getRedirectResponse());
+  }
+
+  /**
+   * Get the redirect URL.
+   *
+   * @return string
+   *   The URL.
+   */
+  public function getUrl() {
```
bojanz: Do we care about getUrl()  / getStatusCode() and getHeaders() ? Similar methods already exist on the response itself. Feels like the only thing we need is the custom constructor.

src/Response/NeedsRedirectException.php
timplunkett: If
`return new NeedsRedirectException($some_url);`
is just a wrapper for
`return new EnforcedResponseException(new TrustedRedirectResponse($some_url));`
then why bother with a new class?

tests/modules/commerce_redirect_test/src/Controller/RedirectTestController.php
```diff
@@ -0,0 +1,38 @@
...
+    return [
+      '#markup' => 'Exception not thrown for secure redirect',
+    ];
+  }
+
+  /**
+   * Throws a local redirect response.
+   */
+  public function throwLocal() {
+    throw new NeedsRedirectException('/commerce_redirect_test/no_throw');
```
timplunkett: In a controller, you would just do `return new TrustedRedirectResponse('/commerce_redirect_test/no_throw');`, no need for an exception. Use the core API.

src/Response/NeedsRedirectException.php
mglaman: 1. Developer experience - just need to provide one class instance, with a URL. Not accidentally pass RedirectResponse
2. ERE is discussed for deprecation

tests/modules/commerce_redirect_test/src/Controller/RedirectTestController.php
mglaman: Yes, but this is just testing our code.

src/Response/NeedsRedirectException.php
mglaman: No, I'll remove them, just left overs from before

src/Response/NeedsRedirectException.php
bojanz: return new EnforcedResponseException(new TrustedRedirectResponse($some_url)); is pretty terrible DX for a payment gateway. A shortcut makes it nicer.

tests/modules/commerce_redirect_test/src/Controller/RedirectTestController.php
bojanz: We don't need to test controllers at all, they have their ways of redirecting. The point of this PR is to allow immediate redirects in nested forms.

src/Response/NeedsRedirectException.php
timplunkett: Instead of a new class that they have to instantiate, I think I would have gone with something like this. Not _that_ much better...

```
trait RedirectHelperTrait {
  protected function forceRedirect($url) {
    return new EnforcedResponseException(new TrustedRedirectResponse($url));
  }
}
```

¯\_(ツ)_/¯

borisson:
This still applies and it looks great, rtbc++

tests/modules/commerce_redirect_test/src/Form/RedirectTestSubForm.php
```diff
@@ -0,0 +1,35 @@
...
+namespace Drupal\commerce_redirect_test\Form;
+
+use Drupal\commerce\Response\NeedsRedirectException;
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Tests throwing a redirect inside of a form.
+ */
+class RedirectTestSubForm extends FormBase {
```
bojanz, line 12: I think this form is unneeded, because it doesn't reflect how we do subforms (we don't use form builder). Better to do the exception in the parent.

tests/modules/commerce_redirect_test/src/Form/RedirectTestForm.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Drupal\commerce_redirect_test\Form;
+
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Tests throwing a redirect inside of a form.
+ */
+class RedirectTestForm extends FormBase {
```
bojanz, line 11: Now that we only have a single form class (+ routing file), we might as well merge the test module into commerce_test, cleaner that way.

src/Response/NeedsRedirectException.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Drupal\commerce\Response;
+
+use Drupal\Component\Utility\UrlHelper;
+use Drupal\Core\Form\EnforcedResponseException;
+use Drupal\Core\Routing\TrustedRedirectResponse;
+
+/**
+ * Provides an exception that represents the need for an HTTP redirect.
```
bojanz, line 10: We should explain why we have this (for easier redirection in nested forms)

bojanz:
Addressed my own feedback and committed in fe24b3eb0f9daa355e35bfdd9602b9d497f1f696.

PR #422, opened 2016-06-23
Issue #2754559 - PhpUnit test file-names/classes must have a Test() suffix for auto-discovery

ndf:
https://www.drupal.org/node/2754559

### DIFF
--- a/modules/cart/tests/src/FunctionalJavascript/AddToCartMultiAttributeTest.php
+++ b/modules/cart/tests/src/FunctionalJavascript/AddToCartMultiAttributeTest.php
@@ -13,7 +13,7 @@ use Drupal\Tests\commerce_cart\Functional\CartBrowserTestBase;
  *
  * @group commerce
  */
-class AddToCartMultiAttribute extends CartBrowserTestBase {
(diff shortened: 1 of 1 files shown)


ndf:
The travis-fail is actually good, because the changed test-class "AddToCartMultiAttributeTest.php" is now actually discovered (and it fails when running).

bojanz:
Great find!

ndf:
All runs ok locally, but Travis seems to have a different issue here.

Time: 37.54 minutes, Memory: 20.00Mb
OK (85 tests, 745 assertions)
  ERROR: No valid tests were specified.
1 fails

bojanz:
@ndf Pushed a fix for that: 11f986978e66fc9abab498ca620552786dc825d4

smmccabe:
Fix to the test looks correct to me, only thing cause a fail is a small code standards whitespace issue now.

bojanz:
Committed in 0d814034503473d906deba2787f74b5ba2f25c59. Fixed the CS violation, removed unneeded comments. Thanks!

PR #423, opened 2016-06-24
Issue #2735441: Port commerce_order_handler_area_order_total

geertvd:


modules/order/src/Plugin/views/area/OrderTotal.php
```diff
@@ -0,0 +1,95 @@
+<?php
+/**
+ * @file
+ * Contains \Drupal\commerce_order\Plugin\views\area\OrderTotal.
```
mglaman:
```
 2 | ERROR | [x] Namespaced classes, interfaces and traits should not
   |       |     begin with a file doc comment
   |       |     (Drupal.Commenting.FileComment.NamespaceNoFileDoc)
```

mglaman:
@bojanz looks good, just needs rebase/squash.

modules/order/commerce_order.views.inc
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @file
+ * Provide views data that isn't tied to any other module.
+ */
+
+/**
+ * Implements hook_views_data().
+ */
+function commerce_order_views_data() {
+  $data['views']['commerce_order_total'] = array(
```
bojanz: We don't use array( in code anymore, needs to be the short array syntax: [ ]

modules/order/src/Plugin/views/area/OrderTotal.php
```diff
@@ -0,0 +1,91 @@
...
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\views\Plugin\views\area\AreaPluginBase;
+use Drupal\views\Plugin\views\argument\NumericArgument;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines an order total area handler.
+ *
+ * Defines an order total area handler that shows the order total field with its
```
bojanz: "Defines an order total area handler that" is redudant, already written in the line above. Start with "Shows the order total...'

modules/order/src/Plugin/views/area/OrderTotal.php
```diff
@@ -0,0 +1,91 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildOptionsForm(&$form, FormStateInterface $form_state) {
+    parent::buildOptionsForm($form, $form_state);
+
+    $form['empty']['#description'] = $this->t("Even if selected, this area handler will never render if a valid order cannot be found in the View's arguments.");
```
bojanz, line 66: Do we need this description? Technically the entire view should not render if there is no valid argument?

geertvd:
> Do we need this description? Technically the entire view should not render if there is no valid argument?

I'd leave that one in, the area handler will still be available for placement in other views and could use an explanation why it won't be rendered.

Will take care of the other feedback later today

modules/order/commerce_order.views.inc
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/**
+ * @file
+ * Provide views data that isn't tied to any other module.
```
bojanz, line 5: That's a weird comment. Why not "Provides views data for the Order module." or "Provides views data for orders."?

mglaman:
Talked with @bojanz, going to address tweak, squash, and merge. Will update with the final commit hash.

mglaman:
Committed as @3014f23d171190bdb92ebf24fbf7adb579f833a8

PR #424, opened 2016-06-24
2755101 Improve access callback logic

haringsrob:


modules/checkout/tests/src/Functional/CheckoutOrderTest.php
```diff
@@ -66,6 +71,66 @@ protected function setUp() {
...
+      'line_items' => [$line_item],
+    ]);
+    $order->save();
+
+    // I should not have access as anonymous.
+    $this->drupalLogout();
+    $this->drupalGet('/checkout/' . $order->id());
+    $this->assertSession()->statusCodeEquals(403);
+
+    // I should have access as the user that owns the order.
```
mglaman, line 116: Should we test other users can't access the order, not just anon + owner?

haringsrob, line 116: Yea why not, it covers a potential security issue. I'll add it

haringsrob, line 116: I have updated the test.

bojanz:
Committed in 857f69da3f3bc89d443d1b5c8969249f4cd68dc5.

PR #426, opened 2016-06-24
Issue 2708455: Update order total on update of quantity line item

PieterJanPut:
Set $matching_line_item to TRUE when the quantity of the line item is changed.


mglaman:
We need tests on the cart page to show this works. This will require a Functional test, or an update to the existing one (new method.)

mglaman:
:/ Sorry, turns out there is #409, which accounts for line items updating.

PR #427, opened 2016-06-25
Issue 2710979: Added crud tests for checkout flows.

finnef:
Fix for https://www.drupal.org/node/2710979.

### DIFF
--- a/modules/checkout/tests/src/Functional/CheckoutFlowTest.php
+++ b/modules/checkout/tests/src/Functional/CheckoutFlowTest.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace Drupal\Tests\commerce_checkout\Functional;
+
+use Drupal\Tests\commerce\Functional\CommerceBrowserTestBase;
+use Drupal\commerce_checkout\Entity\CheckoutFlow;
+
+/**
+ * Tests the checkout flow UI.
+ *
+ * @group commerce
+ */
+class CheckoutFlowTest extends CommerceBrowserTestBase {
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  public static $modules = [
+    'commerce_checkout',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getAdministratorPermissions() {
+    return array_merge([
+      'administer checkout flows',
+    ], parent::getAdministratorPermissions());
+  }
+
+  /**
+   * Tests creating a checkout flow.
+   */
+  public function testCheckoutFlowCreation() {
+    $this->drupalGet('admin/commerce/config/checkout-flows');
+    $this->getSession()->getPage()->clickLink('Add checkout flow');
+    $edit = [
+      'label' => 'Test checkout flow',
+      'id' => 'test_checkout_flow',
+      'plugin' => 'multistep_default',
+    ];
+    $this->submitForm($edit, 'Save');
+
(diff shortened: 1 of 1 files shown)


mglaman:
@finnef I re-ran the TravisCI build. Looks like something went wonkers.

modules/checkout/tests/src/Functional/CheckoutFlowTest.php
```diff
@@ -0,0 +1,109 @@
...
+    $this->getSession()->getPage()->clickLink('Add checkout flow');
+    $edit = [
+      'label' => 'Test checkout flow',
+      'id' => 'test_checkout_flow',
+      'plugin' => 'multistep_default',
+    ];
+    $this->submitForm($edit, 'Save');
+
+    // Test UI shows the correct message.
+    $this->assertSession()->pageTextContains(t("Saved the @name checkout flow.", ['@name' => $edit['label']]));
```
bojanz, line 47: The message should be in single quotes.

modules/checkout/tests/src/Functional/CheckoutFlowTest.php
```diff
@@ -0,0 +1,109 @@
...
+    $edit = [
+      'label' => 'Name has changed',
+      'configuration[panes][billing_information][weight]' => 1,
+      'configuration[panes][contact_information][weight]' => 2,
+      'configuration[panes][review][step_id]' => '_disabled',
+    ];
+    $this->submitForm($edit, 'Save');
+
+    // Test UI shows the correct message.
+    $this->assertSession()->pageTextContains(t("Saved the @name checkout flow.", ['@name' => $edit['label']]));
```
bojanz, line 76: The message should be in single quotes.

modules/checkout/tests/src/Functional/CheckoutFlowTest.php
```diff
@@ -0,0 +1,109 @@
...
+    $checkout_flow = CheckoutFlow::load('test_checkout_flow');
+    $this->assertEquals('Test checkout flow', $checkout_flow->label());
+    $this->assertEquals('multistep_default', $checkout_flow->getPluginId());
+  }
+
+  /**
+   * Tests editing a checkout flow.
+   */
+  public function testCheckoutFlowEditing() {
+    // @todo Test default flow is used when checking out.
```
bojanz, line 59: We can remove this todo, such a test case would belong to a different test, this only only concerns itself with the editing of the checkout flow entity.

modules/checkout/tests/src/Functional/CheckoutFlowTest.php
```diff
@@ -0,0 +1,109 @@
...
+    $this->drupalGet('admin/commerce/config/checkout-flows');
+    $this->getSession()->getPage()->clickLink('Add checkout flow');
+    $edit = [
+      'label' => 'Test checkout flow',
+      'id' => 'test_checkout_flow',
+      'plugin' => 'multistep_default',
+    ];
+    $this->submitForm($edit, 'Save');
+
+    // Test UI shows the correct message.
```
bojanz, line 46: I like to remove comments like these, because they generally don't say anything new, the code is already obvious. Comments are best when they explain the "why" behind the code, not repeat what the code is saying.

bojanz:
Addressed the feedback points and committed as d696d457a581a4c44ae6743410df33f017cb3bc5.

In a followup we need to also test the editing of checkout pane settings.

PR #428, opened 2016-06-25
Issue #2755633 - failing test for variations that are not distinguishable by their attributes

ndf:
Put this PR here for reference.

It contains 4 tests.
testMultipleVariationsWithEqualAttributes() might be relevant for product-creation form


steveoliver:
Let's get the [coding standards errors](https://travis-ci.org/drupalcommerce/commerce/jobs/140228928#L978) handled then do an updated review on this.

mglaman:
@bojanz so this was before we had the select list dropdown. Do we need to make a parent widget afterall which switches the widget? or just close this issue as outdated and leave it to config/documentation?

bojanz:
This is now outdated. The variation selection code was rewritten. Dynamic widget changing is a different issue.

PR #430, opened 2016-06-30
Issue #2758861 by mglaman: Provide an OrderTypeResolver

mglaman:


modules/order/src/Resolver/ChainOrderTypeResolver.php
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * The resolvers.
+   *
+   * @var \Drupal\commerce_order\Resolver\OrderTypeResolverInterface[]
+   */
+  protected $resolvers = [];
+
+  /**
+   * Constructs a new ChainStoreResolver object.
```
bojanz: Not a ChainStoreResolver resolver :)

modules/order/src/Resolver/ChainOrderTypeResolverInterface.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Drupal\commerce_order\Resolver;
+
+/**
+ * Runs the added resolvers one by one until one of them returns the store.
+ *
+ * Each resolver in the chain can be another chain, which is why this interface
+ * extends the store resolver one.
```
bojanz: Comment still referes to stores.

PR #434, opened 2016-07-07
Fix PaymentGateway/Payment field handling

petebarnett:
Brings PaymentGateway bundle functionality into its own class.
Invokes `onBundleDelete` before the config entity is deleted, so that the plugin can be instantiated.
Fixes bugs with `hook_entity_bundle_delete` implementation and field api.

### DIFF
--- a/modules/payment/commerce_payment.module
+++ b/modules/payment/commerce_payment.module
@@ -99,9 +99,23 @@ function commerce_payment_entity_bundle_create($entity_type_id, $bundle) {
  */
 function commerce_payment_entity_bundle_delete($entity_type_id, $bundle) {
   if (in_array($entity_type_id, ['commerce_payment', 'commerce_payment_method'])) {
+    $delete_storage = TRUE;
+    if ($entity_type_id == 'commerce_payment') {
+      $storage = \Drupal::entityTypeManager()->getStorage('commerce_payment_gateway');
+      // Check if other payment gateways are using the same plugin
+      $usage_count = \Drupal::entityQuery('commerce_payment_gateway')
+        ->condition('plugin', $storage->load($bundle)->getPluginId())
+        ->condition('id', $bundle, '<>')
+        ->count()
+        ->execute();
+      // Avoid deleting field storage definitions
+      if ($usage_count) {
+        $delete_storage = FALSE;
+      }
+    }
     $entity_type = \Drupal::entityTypeManager()->getDefinition($entity_type_id);
     $definitions = commerce_payment_entity_bundle_field_info($entity_type, $bundle);
-    commerce_payment_delete_field_definitions($definitions);
+    commerce_payment_delete_field_definitions($definitions, $delete_storage);
   }
 }
 
@@ -125,12 +139,16 @@ function commerce_payment_create_field_definitions(array $definitions) {
  *
  * @param array $definitions
  *   The field definitions.
+ * @param bool $delete_storage
+ *   If set TRUE, field storage definition will be deleted also
  */
-function commerce_payment_delete_field_definitions(array $definitions) {
+function commerce_payment_delete_field_definitions(array $definitions, $delete_storage=FALSE) {
   $field_storage_definition_listener = \Drupal::service('field_storage_definition.listener');
   $field_definition_listener = \Drupal::service('field_definition.listener');
   foreach ($definitions as $definition) {
-    $field_storage_definition_listener->onFieldStorageDefinitionDelete($definition);
     $field_definition_listener->onFieldDefinitionDelete($definition);
+    if ($delete_storage) {
+      $field_storage_definition_listener->onFieldStorageDefinitionDelete($definition);
+    }
   }
 }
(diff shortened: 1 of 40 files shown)


modules/payment/commerce_payment.module
```diff
@@ -125,12 +139,16 @@ function commerce_payment_create_field_definitions(array $definitions) {
  *
  * @param array $definitions
  *   The field definitions.
+ * @param bool $delete_storage
+ *   If set TRUE, field storage definition will be deleted also
```
joshmiller83, line 143: But does it delete the payment entities? What are the ramifications for viewing past orders and payment entities?

modules/payment/src/Entity/PaymentGateway.php
```diff
@@ -159,4 +161,67 @@ protected function getPluginCollection() {
     return $this->pluginCollection;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function postSave(EntityStorageInterface $storage, $update = TRUE) {
```
joshmiller83, line 194: I feel like the order in the code should be: 
preSave
postSave
preDelete
postDelete

No?

modules/payment/commerce_payment.module
petebarnett, line 143: I guess it would reference a non-existent payment gateway
The scope of this PR was just to fix the field functionality associated with payments, but I agree that the above needs to be addressed at some point

modules/payment/src/Entity/PaymentGateway.php
```diff
@@ -159,4 +161,67 @@ protected function getPluginCollection() {
...
+    parent::postSave($storage, $update);
+
+    $bundle_of = $this->getEntityType()->getBundleOf();
+    if (!$update) {
+      \Drupal::getContainer()->get('entity_bundle.listener')->onBundleCreate($this->id(), $bundle_of);
+    }
+    else {
+      $entity_field_manager = \Drupal::getContainer()->get('entity_field.manager');
```
joshmiller83: Can't we do this static and coupled ::getContainer() using a decoupled dependency injection? Or am I crazy?

bojanz: We should be using \Drupal::service() here. Nobody uses \Drupal::getContainer() :)

modules/payment/src/Entity/PaymentGateway.php
```diff
@@ -159,4 +161,67 @@ protected function getPluginCollection() {
...
+
+    foreach ($entities as $entity) {
+      \Drupal::service('entity_bundle.listener')->onBundleDelete($entity->id(), $entity->getEntityType()->getBundleOf());
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function postDelete(EntityStorageInterface $storage, array $entities) {
```
bojanz, line 226: Why do we have an empty postDelete method?

modules/payment/commerce_payment.module
```diff
@@ -125,12 +139,16 @@ function commerce_payment_create_field_definitions(array $definitions) {
...
  */
-function commerce_payment_delete_field_definitions(array $definitions) {
+function commerce_payment_delete_field_definitions(array $definitions, $delete_storage=FALSE) {
```
bojanz: Missing space around '=' for $delete_storage

modules/payment/src/Entity/PaymentGateway.php
petebarnett, line 226: It's not empty, just calls parent... The only thing the original parent did was invoke the hook, which we discussed is no longer required.

modules/payment/commerce_payment.module
```diff
@@ -125,12 +139,16 @@ function commerce_payment_create_field_definitions(array $definitions) {
...
  *   The field definitions.
+ * @param bool $delete_storage
+ *   If set TRUE, field storage definition will be deleted also
  */
-function commerce_payment_delete_field_definitions(array $definitions) {
+function commerce_payment_delete_field_definitions(array $definitions, $delete_storage = FALSE) {
   $field_storage_definition_listener = \Drupal::service('field_storage_definition.listener');
   $field_definition_listener = \Drupal::service('field_definition.listener');
   foreach ($definitions as $definition) {
-    $field_storage_definition_listener->onFieldStorageDefinitionDelete($definition);
```
petebarnett: Ignoring the big change, there is a subtle bug-fix in here as calling `onFieldStorageDefinitionDelete` before `onFieldDefinitionDelete` would result in database errors as Drupal tries to update tables that no longer exist.

bojanz:
I've abstracted out the field handling into a generic API: https://github.com/drupalcommerce/commerce/commit/e09ca1770018dfea1593340e489a36878e3cd7d0
Will be used by Promotions as well, and later moved to Entity API.

After doing that, and looking into this PR, I realized that the gateway approach for bundles that we have here is too magical, so I introduced PaymentType plugins, linked from the gateway instead. 

Thus, I am closing this PR, and will credit you on one of the followup commits, to make up for it.

PR #435, opened 2016-07-10
Add description for currency form fields

levmyshkin:
Hi, I was confused when I was needed to fill Currency code and Currency numeric code fields without any description. I think it will be better to add some information about Currency code and ISO 4217.


bojanz:
In my defence, it's not a form that 99% of people will ever use, since the form is only intended for imaginary/custom currencies, others should just be imported. I should actually see if there's a way to emphasize that more, perhaps with text on top of the form, in addition to your changes.

bojanz:
A variation of this was committed in #462. I do think that there's no good description for the currency code, we already show "XXX" as the placeholder example, indicating the desired format. The numeric code field shows a link to ISO.

PR #439, opened 2016-07-15
Issue #2763701: Create an API for plugin-based entity bundles

mglaman:
@bojanz so this can't be merged because its an amended commit to yours. But this adds a test and fixes some issues. It's two cherry-picks from my payment branch and then the test.

### DIFF
--- a/commerce.module
+++ b/commerce.module
@@ -5,9 +5,10 @@
  * Defines common functionality for all Commerce modules.
  */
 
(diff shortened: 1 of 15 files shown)


bojanz:
Committed in https://github.com/drupalcommerce/commerce/commit/7e8b28b329c389d3521214fc74e56fc8eee4eaed

Tweaked a few details along the way.

PR #445, opened 2016-07-21
Issue #2770747 by mglaman: Implement offer actions

mglaman:
Provides PromotionOffer plugins.

### DIFF
--- a/modules/order/src/Entity/LineItemInterface.php
+++ b/modules/order/src/Entity/LineItemInterface.php
@@ -2,14 +2,14 @@
 
 namespace Drupal\commerce_order\Entity;
 
-use Drupal\commerce_order\AdjustableInterface;
+use Drupal\commerce_order\EntityAdjustableInterface;
 use Drupal\Core\Entity\EntityChangedInterface;
 use Drupal\Core\Entity\ContentEntityInterface;
 
 /**
  * Defines the interface for line items.
  */
-interface LineItemInterface extends AdjustableInterface, ContentEntityInterface, EntityChangedInterface {
+interface LineItemInterface extends ContentEntityInterface, EntityAdjustableInterface, EntityChangedInterface {
 
   /**
    * Gets the parent order.
--- a/modules/order/src/Entity/OrderInterface.php
+++ b/modules/order/src/Entity/OrderInterface.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\commerce_order\Entity;
 
-use Drupal\commerce_order\AdjustableInterface;
+use Drupal\commerce_order\EntityAdjustableInterface;
 use Drupal\commerce_store\Entity\EntityStoreInterface;
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityChangedInterface;
@@ -12,7 +12,7 @@ use Drupal\profile\Entity\ProfileInterface;
 /**
  * Defines the interface for orders.
  */
-interface OrderInterface extends AdjustableInterface, ContentEntityInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoreInterface {
+interface OrderInterface extends ContentEntityInterface, EntityAdjustableInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoreInterface {
 
   /**
    * Gets the order number.
--- a/modules/order/src/EntityAdjustableInterface.php
+++ b/modules/order/src/EntityAdjustableInterface.php
@@ -1,14 +1,15 @@
 <?php
 
 namespace Drupal\commerce_order;
+use Drupal\Core\Entity\EntityInterface;
 
 /**
  * Defines an interface for objects that contain adjustments.
  *
- * @see \Drupal\commerce_order\Entity\OrderInterface
- * @see \Drupal\commerce_order\Entity\LineItemInterface
+ * @see \Drupal\commerce_order\Entity\OrderInterfaceEntity
+ * @see \Drupal\commerce_order\Entity\LineItemInterfaceEntity
  */
-interface AdjustableInterface {
+interface EntityAdjustableInterface extends EntityInterface {
 
   /**
    * Gets the adjustments.
--- a/modules/promotion/src/Annotation/PromotionOffer.php
+++ b/modules/promotion/src/Annotation/PromotionOffer.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Drupal\commerce_promotion\Annotation;
+
+use Drupal\Component\Annotation\Plugin;
+
+/**
+ * Defines the promotion offer plugin annotation object.
+ *
+ * Plugin namespace: Plugin\Commerce\PromotionOffer.
+ *
+ * @Annotation
+ */
+class PromotionOffer extends Plugin {
+
+  /**
+   * The plugin ID.
+   *
+   * @var string
+   */
+  public $id;
+
+  /**
+   * The label of the plugin.
+   *
+   * @var \Drupal\Core\Annotation\Translation
+   *
+   * @ingroup plugin_translatable
+   */
+  public $label;
+
+  /**
+   * An array of context definitions describing the context used by the plugin.
+   *
+   * The array is keyed by context names.
+   *
+   * @var \Drupal\Core\Annotation\ContextDefinition[]
+   */
+  public $context = [];
+
+  /**
+   * The target entity type this action applies to.
+   *
+   * For example, this should be 'commerce_order' or 'commerce_line_item'.
+   *
+   * @var string
+   */
+  public $target_entity_type;
+
+  /**
+   * The category under which the offer should listed in the UI.
+   *
+   * @var \Drupal\Core\Annotation\Translation
+   *
+   * @ingroup plugin_translatable
+   */
+  public $category;
+
+}
--- a/modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
+++ b/modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer;
+
+/**
+ * Provides a 'Order: Percentage off' condition.
+ *
+ * @PromotionOffer(
+ *   id = "commerce_promotion_order_percentage_off",
+ *   label = @Translation("Percentage off"),
+ *   target_entity_type = "commerce_order",
+ * )
+ */
+class OrderPercentageOff extends PercentageOffBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
+    $order = $this->getTargetEntity();
+    $price_amount = $order->getTotalPrice()->multiply($this->getAmount());
+    $this->applyAdjustment($order, $price_amount);
+  }
+
+}
--- a/modules/promotion/src/Plugin/Commerce/PromotionOffer/PercentageOffBase.php
+++ b/modules/promotion/src/Plugin/Commerce/PromotionOffer/PercentageOffBase.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer;
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides the base class for percentage off offers.
+ */
+abstract class PercentageOffBase extends PromotionOfferBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return [
+      'amount' => 0,
+    ] + parent::defaultConfiguration();
+  }
+
+  /**
+   * Gets the percentage amount, as a decimal, negated.
+   *
+   * @return string
+   *   The amount.
+   */
+  public function getAmount() {
+    return (string) $this->configuration['amount'];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form += parent::buildConfigurationForm($form, $form_state);
+
+    $form['amount'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Percentage'),
+      '#default_value' => $this->configuration['amount'] * 100,
+      '#maxlength' => 255,
+      '#required' => TRUE,
+      '#step' => 0.1,
+      '#min' => 0,
+      '#max' => 100,
+      '#length' => 4,
+      '#field_suffix' => t('%'),
+    ];
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validateConfigurationForm(array &$form, FormStateInterface $form_state) {
+    $values = $form_state->getValue($form['#parents']);
+    if (empty($values['target_plugin_configuration']['amount'])) {
+      $form_state->setError($form, $this->t('Percentage amount cannot be empty.'));
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
+    $values = $form_state->getValue($form['#parents']);
+    $this->configuration['amount'] = (string) ($values['amount'] / 100);
+    parent::submitConfigurationForm($form, $form_state);
+  }
+
+}
--- a/modules/promotion/src/Plugin/Commerce/PromotionOffer/ProductPercentageOff.php
+++ b/modules/promotion/src/Plugin/Commerce/PromotionOffer/ProductPercentageOff.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer;
+
+/**
+ * Provides a 'Product: Percentage off' condition.
+ *
+ * @PromotionOffer(
+ *   id = "commerce_promotion_product_percentage_off",
+ *   label = @Translation("Percentage off"),
+ *   target_entity_type = "commerce_line_item",
+ * )
+ */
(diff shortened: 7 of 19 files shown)


mglaman:
#446 should tests!

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferInterface.php
```diff
@@ -0,0 +1,50 @@
...
+use Drupal\Component\Plugin\ConfigurablePluginInterface;
+use Drupal\Component\Plugin\PluginInspectionInterface;
+use Drupal\Core\Executable\ExecutableInterface;
+use Drupal\Core\Plugin\ContextAwarePluginInterface;
+use Drupal\Core\Plugin\PluginFormInterface;
+
+/**
+ * Defines an interface for Offer plugins.
+ */
+interface PromotionOfferInterface extends ExecutableInterface, PluginFormInterface, ConfigurablePluginInterface, PluginInspectionInterface, ContextAwarePluginInterface {
```
mglaman, line 16: I'm not sure if it makes sense to extend core's `ExecutableInterface`. It just provides the execute method, but maybe we should provide our own method here that is `apply` or `run`, or something. And it can return a `bool` value so we know if it applied or not.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferBase.php
```diff
@@ -0,0 +1,106 @@
...
+    return $this->getContextValue($this->appliesTo());
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applyAdjustment(AdjustableInterface $item, Price $amount) {
+    $item->addAdjustment(new Adjustment([
+      'type' => 'promotion',
+      'label' => $this->pluginDefinition['label'],
```
mglaman: This is the plugin definition, we need to add a "Label" to the base configuration form. Or, if its on root promotion form make sure it sets it.

Actually, that's what should be done in https://www.drupal.org/node/2770731

mglaman:
PHPCS and it broke the FunctionalJavascript tests, which need to be expanded anyways to set an offer.

modules/promotion/src/Annotation/PromotionOffer.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+namespace Drupal\commerce_promotion\Annotation;
+
+use Drupal\Component\Annotation\Plugin;
+
+/**
+ * Defines a Promotion Offer item annotation object.
```
bojanz: "Defines the promotion offer plugin annotation object."

modules/promotion/src/Annotation/PromotionOffer.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Plugin namespace: Plugin\Commerce\PromotionOffer.
+ *
+ * @Annotation
+ */
+class PromotionOffer extends Plugin {
+
+
```
bojanz: Double newline.

modules/promotion/src/Annotation/PromotionOffer.php
```diff
@@ -0,0 +1,67 @@
...
+   * @ingroup plugin_translatable
+   */
+  public $label;
+
+  /**
+   * The name of the module providing the type.
+   *
+   * @var string
+   */
+  public $module;
```
bojanz: Each plugin has a 'provider' set, so this should be unneeded?

modules/promotion/src/Annotation/PromotionOffer.php
```diff
@@ -0,0 +1,67 @@
...
+   * The array is keyed by context names.
+   *
+   * @var \Drupal\Core\Annotation\ContextDefinition[]
+   */
+  public $context = [];
+
+  /**
+   * The target this action applies to.
+   *
+   * For example, this should be 'order' or 'line_item'.
```
bojanz: Feels weird for the possible targets to be up in the air. Maybe we make this $target_entity_type?

modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    if ($this->configuration['amount'] != 0) {
+      /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
+      $order = $this->getTargetEntity();
+
+      $calculated_amount = -($order->total_price->amount * $this->getAmount());
```
bojanz: Might make sense to merge the Calculator class first, then rebase on top of it.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferBase.php
bojanz: Let's default it to t('Discount'), that's a safe option other systems (and discount 1.x) use, then and add a todo for #2770731.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferBase.php
```diff
@@ -0,0 +1,106 @@
...
+      'amount' => $amount,
+      'source_id' => $this->getPluginId(),
```
bojanz, line 51: The source would be the ID of the promotion entity, not the plugin ID.

modules/promotion/src/PromotionOfferManager.php
```diff
@@ -0,0 +1,91 @@
...
+    // If the plugin did not specify a category, process that.
+    if (empty($definition['category'])) {
+      if ($target == PromotionOfferInterface::ORDER) {
+        $definition['category'] = $this->t('Order');
+      }
+      elseif ($target == PromotionOfferInterface::LINE_ITEM) {
+        $definition['category'] = $this->t('Product');
+      }
+      else {
+        $definition['category'] = $this->t('Other');
```
bojanz: Is there such a thing as "Other"?

src/Plugin/Field/FieldType/PluginItemDeriver.php
```diff
@@ -18,6 +18,7 @@ public function getDerivativeDefinitions($base_plugin_definition) {
     $supported = [
       'condition' => $this->t('Conditions'),
       'action' => $this->t('Action'),
+      'commerce_promotion_offer' => $this->t('Promotion Offer'),
```
bojanz: We don't capitalize the second word in labels usually, no?

modules/promotion/src/Plugin/Commerce/PromotionOffer/ProductPercentageOff.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer;
+use Drupal\commerce_price\Price;
```
bojanz: Missing newline after namespace.

modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer;
```
bojanz: Missing newline after namespace.

modules/promotion/src/Annotation/PromotionOffer.php
mglaman: Because it made the annotation seem long and ugly. `targetEntityType` over `target`. But we can change it to being more verbose

modules/promotion/src/PromotionOfferManager.php
mglaman: Shipment? I had no idea what to do here because we can't throw an exception because what if people add promotions targeting other entities. Maybe I just need to load the entity definition and get the entity's label.

modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
mglaman: Fixed, this line just didn't change.

modules/promotion/src/Annotation/PromotionOffer.php
mglaman: Changing to targetEntityType, since we need an actual entity type plugin ID.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferInterface.php
mglaman, line 16: This is stupid, because that has to do with conditions and is a different PR to be made.

mglaman:
It looks like core doesn't support fields that use shared storage and have serialized properties. So we'll need to move Offer to field storage item instead of base field.

sqlcontententitystorage_php_-_commerce2x_-____drupal_sites_commerce2x_

modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
```diff
@@ -0,0 +1,28 @@
...
+ *   targetEntityType = "commerce_order",
+ * )
+ */
+class OrderPercentageOff extends PercentageOffBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    if ($this->configuration['amount'] != 0) {
```
bojanz: Why would the settings form allow a 0 value to be accepted?

modules/promotion/src/Annotation/PromotionOffer.php
```diff
@@ -0,0 +1,59 @@
...
+  public $context = [];
+
+  /**
+   * The target entity type this action applies to.
+   *
+   * For example, this should be 'commerce_order' or 'commerce_line_item'.
+   *
+   * @var string
+   */
+  public $targetEntityType;
```
bojanz: Annotation keys are snake_case, not camelCase. So it should be target_entity_type

modules/promotion/src/Plugin/Commerce/PromotionOffer/PercentageOffBase.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer;
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides base class for percentage off offers.
```
bojanz: Provides _the_ base class for percentage off offers.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PercentageOffBase.php
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Gets the percentage amount, as a decimal, negated.
+   *
+   * @return string
+   *   The amount.
+   */
+  public function getAmount() {
+    return (string) (-$this->configuration['amount']);
```
bojanz: Should the negation be done here, or in the code that creates the adjustment? Isn't it saner to say "This class provides a 19% off promotion" instead of "This class provides a -19% off promotion"?

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferBase.php
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function validateConfigurationForm(array &$form, FormStateInterface $form_state) {
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
```
bojanz: I think we usually do {} for empty methods?

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferInterface.php
```diff
@@ -0,0 +1,50 @@
...
+   * @return \Drupal\Core\Entity\EntityInterface
+   *   The target entity.
+   */
+  public function getTargetEntity();
+
+  /**
+   * Applies the promotion offer's adjustment to an item.
+   *
+   * @param \Drupal\commerce_order\AdjustableInterface $item
+   *   The adjustable item.
```
bojanz: This is odd. Why is it $item and not $entity?

modules/promotion/src/PromotionOfferManager.php
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * Constructor for OfferManager objects.
```
bojanz: Constructs a new PromotionOfferManager object.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PercentageOffBase.php
mglaman: Well \Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\PromotionOfferBase::applyAdjustment already has a Price passed to it to pass to the Adjustment.

and in \Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\ProductPercentageOff::execute we invoke Multiple. I guess in the execute we can flip \Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\PercentageOffBase::getAmount.

Just felt like better DX.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferInterface.php
mglaman: Because we don't specify that AdjustableInterface extends any entity interface, and could therefore be anything

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferBase.php
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applyAdjustment(AdjustableInterface $item, Price $amount) {
+    $item->addAdjustment(new Adjustment([
+      'type' => 'promotion',
+      // @todo Change to label from UI when added in #2770731.
+      'label' => t('Discount'),
+      'amount' => $amount->multiply('-1'),
```
mglaman, line 50: @bojanz this was my solution. so the promotion price is always inversed when creating the adjustment. 

So an offer will always return a price, but this ensures it gets deducted.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferInterface.php
bojanz: Perhaps that's our mistake.

mglaman: Should it extend EntityInterface, then? I mean in reality anyone could have an object that collects adjustments for some use case or another without it being a content entity.

modules/promotion/src/Plugin/Commerce/PromotionOffer/PromotionOfferInterface.php
```diff
@@ -0,0 +1,50 @@
...
+   *
+   * @return \Drupal\Core\Entity\EntityInterface
+   *   The target entity.
+   */
+  public function getTargetEntity();
+
+  /**
+   * Applies the promotion offer's adjustment to an item.
+   *
+   * @param \Drupal\commerce_order\EntityAdjustableInterface $item
```
bojanz: Now that it's EntityAdjustableInterface, we can rename $item to $entity :)

modules/order/src/EntityAdjustableInterface.php
```diff
@@ -5,10 +5,10 @@
 /**
  * Defines an interface for objects that contain adjustments.
  *
- * @see \Drupal\commerce_order\Entity\OrderInterface
- * @see \Drupal\commerce_order\Entity\LineItemInterface
+ * @see \Drupal\commerce_order\Entity\OrderInterfaceEntity
+ * @see \Drupal\commerce_order\Entity\LineItemInterfaceEntity
  */
-interface AdjustableInterface {
+interface EntityAdjustableInterface {
```
bojanz: extends EntityInterface?

modules/order/src/Entity/OrderInterface.php
```diff
@@ -12,7 +12,7 @@
 /**
  * Defines the interface for orders.
  */
-interface OrderInterface extends AdjustableInterface, ContentEntityInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoreInterface {
+interface OrderInterface extends EntityAdjustableInterface, ContentEntityInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoreInterface {
```
bojanz: Now that the name has changed, let's move it after ContentEntityInterface

modules/order/src/Entity/LineItemInterface.php
```diff
@@ -2,14 +2,14 @@
...
-use Drupal\commerce_order\AdjustableInterface;
+use Drupal\commerce_order\EntityAdjustableInterface;
 use Drupal\Core\Entity\EntityChangedInterface;
 use Drupal\Core\Entity\ContentEntityInterface;
 
 /**
  * Defines the interface for line items.
  */
-interface LineItemInterface extends AdjustableInterface, ContentEntityInterface, EntityChangedInterface {
+interface LineItemInterface extends EntityAdjustableInterface, ContentEntityInterface, EntityChangedInterface {
```
bojanz: Now that the name has changed, let's move it after ContentEntityInterface

mglaman: Diff didn't make this disappear, but moved.

PR #447, opened 2016-07-23
Adjustments

mglaman:
@bojanz this is going to need some work. I've tried to leave plenty of notes and todo comments.

However, there is some strong test coverage along the way.

### DIFF
--- a/modules/order/src/AdjustableInterface.php
+++ b/modules/order/src/AdjustableInterface.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+/**
+ * Defines an interface for objects that contain adjustments.
+ *
+ * @see \Drupal\commerce_order\Entity\OrderInterface
+ * @see \Drupal\commerce_order\Entity\LineItemInterface
+ */
+interface AdjustableInterface {
+
+  /**
+   * Gets the order's adjustments.
+   *
+   * @return \Drupal\commerce_order\Adjustment[]
+   *   The adjustments.
+   */
+  public function getAdjustments();
+
+  /**
+   * Sets the adjustments for an order.
+   *
+   * @param \Drupal\commerce_order\Adjustment[] $adjustments
+   *   The adjustments.
+   *
+   * @return $this
+   */
+  public function setAdjustments(array $adjustments);
+
+  /**
+   * Adds an adjustment to the order.
+   *
+   * @param \Drupal\commerce_order\Adjustment $adjustment
+   *   The adjustment.
+   *
+   * @return $this
+   */
+  public function addAdjustment(Adjustment $adjustment);
+
+  /**
+   * Removes an adjustment.
+   *
+   * The adjustment is removed by targeting the source identifier.
+   *
+   * @param \Drupal\commerce_order\Adjustment $adjustment
+   *   The adjustment to remove.
+   *
+   * @return $this
+   */
+  public function removeAdjustment(Adjustment $adjustment);
+
+}
--- a/modules/order/src/Adjustment.php
+++ b/modules/order/src/Adjustment.php
@@ -1,0 +1,108 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+use Drupal\commerce_price\Price;
+
+/**
+ * Represents an adjustment.
+ */
+final class Adjustment {
+
+  /**
+   * The adjustment type.
+   *
+   * @var string
+   */
+  protected $type;
+
+  /**
+   * The adjustment label.
+   *
+   * @var string
+   */
+  protected $label;
+
+  /**
+   * The price value object.
+   *
+   * @var \Drupal\commerce_price\Price
+   */
+  protected $amount;
+
+  /**
+   * The source identifier of the adjustment.
+   *
+   * Points to the source object, if known. For example, a promotion entity for
+   * a discount adjustment.
+   *
+   * @var string
+   */
+  protected $sourceId;
+
+  /**
+   * Constructs a new Adjustment instance.
+   *
+   * @param array $definition
+   *   The definition.
+   */
+  public function __construct(array $definition) {
+    foreach (['type', 'label', 'amount'] as $required_property) {
+      if (empty($definition[$required_property])) {
+        throw new \InvalidArgumentException(sprintf('Missing required property %s.', $required_property));
+      }
+    }
+    if (!$definition['amount'] instanceof Price) {
+      throw new \InvalidArgumentException(sprintf('Property amount should be an instance of %s.', Price::class));
+    }
+
+    $this->type = $definition['type'];
+    $this->label = $definition['label'];
+    $this->amount = $definition['amount'];
+
+    if (!empty($definition['source_id'])) {
+      $this->sourceId = $definition['source_id'];
+    }
+  }
+
+  /**
+   * Gets the adjustment type.
+   *
+   * @return string
+   *   The adjustment type.
+   */
+  public function getType() {
+    return $this->type;
+  }
+
+  /**
+   * Gets the adjustment label.
+   *
+   * @return string
+   *   The adjustment label.
+   */
+  public function getLabel() {
+    return $this->label;
+  }
+
+  /**
+   * Get the source identifier.
+   *
+   * @return string
+   *   The source identifier.
+   */
+  public function getSourceId() {
+    return $this->sourceId;
+  }
+
+  /**
+   * Gets the Price value object for the adjustment.
+   *
+   * @return \Drupal\commerce_price\Price
+   *   The Price value object.
+   */
+  public function getAmount() {
+    return $this->amount;
+  }
+
+}
--- a/modules/order/src/Entity/LineItem.php
+++ b/modules/order/src/Entity/LineItem.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\commerce_order\Entity;
 
+use Drupal\commerce_order\Adjustment;
 use Drupal\Core\Entity\ContentEntityBase;
 use Drupal\Core\Entity\EntityChangedTrait;
 use Drupal\Core\Entity\EntityStorageInterface;
@@ -158,11 +159,59 @@ class LineItem extends ContentEntityBase implements LineItemInterface {
   /**
    * {@inheritdoc}
    */
+  public function getAdjustments() {
+    return $this->get('adjustments')->getAdjustments();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setAdjustments(array $adjustments) {
+    $this->set('adjustments', $adjustments);
+    $this->recalculateTotalPrice();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addAdjustment(Adjustment $adjustment) {
+    $this->get('adjustments')->appendItem($adjustment);
+    $this->recalculateTotalPrice();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function removeAdjustment(Adjustment $adjustment) {
+    $this->get('adjustments')->removeAdjustment($adjustment);
+    $this->recalculateTotalPrice();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function preSave(EntityStorageInterface $storage) {
     parent::preSave($storage);
+    $this->recalculateTotalPrice();
+  }
 
+  /**
+   * Recalculates the line item total price.
+   *
+   * This will update the total price based on line item and adjustment prices.
+   */
+  protected function recalculateTotalPrice() {
+    // @todo Rework this once pricing is finished.
     $this->total_price->amount = $this->unit_price->amount * $this->quantity->value;
     $this->total_price->currency_code = $this->unit_price->currency_code;
+
+    foreach ($this->getAdjustments() as $adjustment) {
+      // @todo The price field should have an ::addPrice method.
+      $this->total_price->amount += $adjustment->getAmount()->getDecimalAmount();
+    }
   }
 
   /**
@@ -226,6 +275,14 @@ class LineItem extends ContentEntityBase implements LineItemInterface {
       ->setDisplayConfigurable('form', TRUE)
       ->setDisplayConfigurable('view', TRUE);
 
+    $fields['adjustments'] = BaseFieldDefinition::create('commerce_adjustment')
+      ->setLabel(t('Adjustments'))
+      ->setDescription(t('Adjustments to the order total.'))
+      ->setRequired(FALSE)
+      ->setCardinality(BaseFieldDefinition::CARDINALITY_UNLIMITED)
+      ->setDisplayConfigurable('form', FALSE)
+      ->setDisplayConfigurable('view', TRUE);
+
     $fields['total_price'] = BaseFieldDefinition::create('commerce_price')
       ->setLabel(t('Total price'))
       ->setDescription(t('The total price of the line item.'))
@@ -271,7 +328,7 @@ class LineItem extends ContentEntityBase implements LineItemInterface {
     else {
       // This line item type won't reference a purchasable entity. The field
       // can't be removed here, or converted to a configurable one, so it's
-      // hidden instead. See https://www.drupal.org/node/2346347#comment-10254087.
+      // hidden instead. https://www.drupal.org/node/2346347#comment-10254087.
       $fields['purchased_entity']->setRequired(FALSE);
       $fields['purchased_entity']->setDisplayOptions('form', [
         'type' => 'hidden',
--- a/modules/order/src/Entity/Order.php
+++ b/modules/order/src/Entity/Order.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\commerce_order\Entity;
 
+use Drupal\commerce_order\Adjustment;
 use Drupal\commerce_store\Entity\StoreInterface;
 use Drupal\Core\Entity\ContentEntityBase;
 use Drupal\Core\Entity\EntityChangedTrait;
@@ -87,11 +88,25 @@ class Order extends ContentEntityBase implements OrderInterface {
     }
 
     // Recalculate the total.
+    $this->recalculateTotalPrice();
+  }
+
+  /**
+   * Recalculates the line item total price.
+   *
+   * This will update the total price based on line item and adjustment prices.
+   */
+  protected function recalculateTotalPrice() {
     // @todo Rework this once pricing is finished.
     $this->total_price->amount = 0;
     foreach ($this->getLineItems() as $line_item) {
       $this->total_price->amount += $line_item->total_price->amount;
       $this->total_price->currency_code = $line_item->total_price->currency_code;
+    }
+
+    foreach ($this->getAdjustments() as $adjustment) {
+      // @todo The price field should have an ::addPrice method.
+      $this->total_price->amount += $adjustment->getAmount()->getDecimalAmount();
     }
   }
 
@@ -270,6 +285,7 @@ class Order extends ContentEntityBase implements OrderInterface {
    */
   public function setLineItems(array $line_items) {
     $this->set('line_items', $line_items);
+    $this->recalculateTotalPrice();
     return $this;
   }
 
@@ -286,6 +302,7 @@ class Order extends ContentEntityBase implements OrderInterface {
   public function addLineItem(LineItemInterface $line_item) {
     if (!$this->hasLineItem($line_item)) {
       $this->get('line_items')->appendItem($line_item);
+      $this->recalculateTotalPrice();
     }
     return $this;
   }
@@ -389,6 +406,40 @@ class Order extends ContentEntityBase implements OrderInterface {
   /**
    * {@inheritdoc}
    */
+  public function getAdjustments() {
+    return $this->get('adjustments')->getAdjustments();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setAdjustments(array $adjustments) {
+    $this->set('adjustments', $adjustments);
+    $this->recalculateTotalPrice();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addAdjustment(Adjustment $adjustment) {
+    $this->get('adjustments')->appendItem($adjustment);
+    $this->recalculateTotalPrice();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function removeAdjustment(Adjustment $adjustment) {
+    $this->get('adjustments')->removeAdjustment($adjustment);
+    $this->recalculateTotalPrice();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
     $fields = parent::baseFieldDefinitions($entity_type);
 
@@ -468,6 +519,18 @@ class Order extends ContentEntityBase implements OrderInterface {
       ->setDisplayConfigurable('form', TRUE)
       ->setDisplayConfigurable('view', TRUE)
       ->setSetting('workflow_callback', ['\Drupal\commerce_order\Entity\Order', 'getWorkflowId']);
+
+    $fields['adjustments'] = BaseFieldDefinition::create('commerce_adjustment')
+      ->setLabel(t('Adjustments'))
+      ->setDescription(t('Adjustments to the order total.'))
+      ->setRequired(FALSE)
+      ->setCardinality(BaseFieldDefinition::CARDINALITY_UNLIMITED)
+      ->setDisplayOptions('form', [
+        'type' => 'commerce_adjustment_default',
+        'weight' => 0,
+      ])
+      ->setDisplayConfigurable('form', FALSE)
+      ->setDisplayConfigurable('view', FALSE);
 
     $fields['total_price'] = BaseFieldDefinition::create('commerce_price')
       ->setLabel(t('Total price'))
--- a/modules/order/src/Entity/OrderInterface.php
+++ b/modules/order/src/Entity/OrderInterface.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\commerce_order\Entity;
 
+use Drupal\commerce_order\AdjustableInterface;
 use Drupal\commerce_store\Entity\EntityStoreInterface;
 use Drupal\Core\Entity\EntityChangedInterface;
 use Drupal\Core\Entity\EntityInterface;
@@ -11,7 +12,7 @@ use Drupal\profile\Entity\ProfileInterface;
 /**
  * Defines the interface for orders.
  */
-interface OrderInterface extends EntityStoreInterface, EntityChangedInterface, EntityInterface, EntityOwnerInterface {
+interface OrderInterface extends EntityStoreInterface, EntityChangedInterface, EntityInterface, EntityOwnerInterface, AdjustableInterface {
 
   /**
    * Gets the order number.
--- a/modules/order/src/Form/OrderTypeForm.php
+++ b/modules/order/src/Form/OrderTypeForm.php
@@ -6,6 +6,9 @@ use Drupal\Core\Entity\BundleEntityFormBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 
+/**
+ * Provides an order type form.
+ */
 class OrderTypeForm extends BundleEntityFormBase {
 
   /**
--- a/modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
+++ b/modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Drupal\commerce_order\Plugin\Field\FieldType;
+
+use Drupal\commerce_order\Adjustment;
+use Drupal\Core\Field\FieldItemBase;
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\TypedData\DataDefinition;
+
+/**
+ * Plugin implementation of the 'commerce_adjustment' field type.
+ *
+ * @FieldType(
+ *   id = "commerce_adjustment",
+ *   label = @Translation("Adjustment"),
+ *   description = @Translation("Stores adjustments to the parent entity's price."),
+ *   list_class = "\Drupal\commerce_order\Plugin\Field\FieldType\AdjustmentItemList",
+ *   default_widget = "commerce_adjustment_default",
+ * )
+ */
+class AdjustmentItem extends FieldItemBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function propertyDefinitions(FieldStorageDefinitionInterface $field_definition) {
+    $properties['value'] = DataDefinition::create('any')
+      ->setLabel(t('Value'))
+      ->setRequired(TRUE);
+
+    return $properties;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isEmpty() {
+    return $this->value === NULL || !$this->value instanceof Adjustment;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setValue($values, $notify = TRUE) {
+    if (!$values instanceof Adjustment) {
+      $values = NULL;
+    }
+    parent::setValue($values, $notify);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function schema(FieldStorageDefinitionInterface $field_definition) {
+    return [
+      'columns' => [
+        'value' => [
+          'description' => 'The adjustment value object.',
+          'type' => 'blob',
+          'not null' => TRUE,
+          'serialize' => TRUE,
+        ],
+      ],
+    ];
+  }
+
+}
--- a/modules/order/src/Plugin/Field/FieldType/AdjustmentItemList.php
+++ b/modules/order/src/Plugin/Field/FieldType/AdjustmentItemList.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Drupal\commerce_order\Plugin\Field\FieldType;
+
+use Drupal\commerce_order\Adjustment;
+use Drupal\Core\Field\FieldItemList;
+
+/**
+ * Represents a list of adjustment item field values.
+ */
+class AdjustmentItemList extends FieldItemList implements AdjustmentItemListInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAdjustments() {
+    $adjustments = [];
+    /** @var \Drupal\commerce_order\Plugin\Field\FieldType\AdjustmentItem $field_item */
+    foreach ($this->list as $key => $field_item) {
+      $adjustments[$key] = $field_item->value;
+    }
+
+    return $adjustments;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function removeAdjustment(Adjustment $adjustment) {
+    /** @var \Drupal\commerce_order\Plugin\Field\FieldType\AdjustmentItem $field_item */
+    foreach ($this->list as $key => $field_item) {
+      if ($field_item->value === $adjustment) {
+        $this->removeItem($key);
+      }
+    }
+  }
+
+}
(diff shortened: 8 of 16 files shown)


modules/order/src/Entity/OrderInterface.php
```diff
@@ -222,4 +223,23 @@ public function getBillingProfileId();
...
 
+  /**
+   * Adds an adjustment to the order.
+   *
+   * @param \Drupal\commerce_order\Adjustment $adjustment
+   *    The adjustment.
+   *
+   * @return $this
+   *
+   * @todo: How do we remove or clear them?
```
mglaman: We never talked about this. We'll need some sort of "Remove" button in the UI. But how about programmatically?

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,124 @@
...
+
+use Drupal\commerce_order\Adjustment;
+use Drupal\commerce_price\Plugin\Field\FieldType\PriceItem;
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\TypedData\DataDefinition;
+
+/**
+ * Plugin implementation of the 'commerce_price' field type.
+ *
+ * @todo See if bojanz throws bricks at me for extended PriceItem.
```
mglaman: 😄 See next line for reasoning.

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
```diff
@@ -0,0 +1,99 @@
...
+namespace Drupal\commerce_order\Plugin\Field\FieldWidget;
+
+use Drupal\commerce_price\Plugin\Field\FieldWidget\PriceDefaultWidget;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Plugin implementation of the 'commerce_order_adjustment_default' widget.
+ *
+ * @todo Needs a functional test.
```
mglaman: Need Functional test for widget/UI

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
```diff
@@ -0,0 +1,99 @@
...
+    ];
+
+    return $element;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function massageFormValues(array $values, array $form, FormStateInterface $form_state) {
+    foreach ($values as $key => $value) {
```
mglaman: This is ugly. I wonder if we can just hack `#parents` in form. But it gets job done for now.

modules/price/src/Price.php
```diff
@@ -37,7 +37,7 @@ public function __construct($amount, $currency_code) {
   /**
    * Gets the decimal amount.
    *
-   * @return string
+   * @return float
```
mglaman: Type hint fix, `+=` was saying wrong string concat operator.

modules/order/src/Entity/Order.php
```diff
@@ -87,12 +88,28 @@ public function preSave(EntityStorageInterface $storage) {
...
       $this->total_price->amount += $line_item->total_price->amount;
       $this->total_price->currency_code = $line_item->total_price->currency_code;
     }
+
+    /** @var \Drupal\commerce_order\Plugin\Field\FieldType\AdjustmentItem $adjustment */
+    // @todo FieldItemList implementation which just gets Adjustment that
+    // returns Adjustment objects directly.
+    foreach ($this->adjustments as $adjustment) {
+      $adjustment_price = $adjustment->toPrice();
+      // @todo The price field should have an ::addPrice method.
```
mglaman, line 108: What are thoughts on adding a method to the price field that consumes a price object and adds its to the amount value (if currency codes match.)

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,126 @@
...
+   *   The definition.
+   */
+  public function __construct(array $definition) {
+    foreach (['type', 'label', 'amount', 'currency_code', 'source_id'] as $required_property) {
+      if (empty($definition[$required_property])) {
+        throw new \InvalidArgumentException(sprintf('Missing required property %s.', $required_property));
+      }
+    }
+
+    $this->type = $definition['type'];
```
rszrama, line 59: This could be a pattern we're using elsewhere, but I'm curious why we aren't setting these values above in the foreach. e.g. $this->{$required_property} = $definition[$required_property];

Perhaps because we also want to initialize optional default properties in other classes?

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,126 @@
...
+    return $this->label;
+  }
+
+  /**
+   * Gets the decimal amount.
+   *
+   * @return string
+   *   The decimal amount.
+   */
+  public function getDecimalAmount() {
```
rszrama: Problem with just getAmount()? Seems odd to have a mismatch in the function name and property name. Was more useful in 1.x where we had the concept of integer price field amounts and decimal amounts.

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * The currency code.
+   *
+   * @var string
+   */
+  protected $currencyCode;
+
+  /**
+   * The source identifier of the adjustment.
```
rszrama, line 34: What's a source identifier?

modules/order/src/Entity/Order.php
rszrama, line 108: Pretty sure we had a similar function in 1.x, and if not we definitely needed one. ; D

Any reason not to automatically convert currency, tho?

modules/order/src/Entity/Order.php
```diff
@@ -389,6 +406,14 @@ public function setData($data) {
   /**
    * {@inheritdoc}
    */
+  public function addAdjustment(Adjustment $adjustment) {
+    $this->get('adjustments')->appendItem($adjustment);
+    return $this;
```
rszrama, line 437: Should we recalculateTotal() after appending the adjustment?

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,124 @@
...
+    ];
+
+    return $schema;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setValue($values, $notify = TRUE) {
+    // Allow callers to pass am Adjustment object as the field item value.
```
rszrama: am != an

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function schema(FieldStorageDefinitionInterface $field_definition) {
+    $schema = parent::schema($field_definition);
+
+    $schema['columns']['type'] = [
+      'description' => 'The decimal amount.',
```
rszrama: copy/pasta in this description and the label description.

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,124 @@
...
+      'currency_code' => $this->currency_code,
+      'source_id' => $this->source_id,
+    ]);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function toPrice() {
+    return $this->toAdjustment()->toPrice();
```
rszrama: PriceItems don't already have toPrice()?

modules/order/commerce_order.module
```diff
@@ -157,6 +157,49 @@ function commerce_order_add_line_items_field($order_type) {
...
 /**
+ * Adds the default adjustments field to an order type.
+ *
+ * Adjustments can't be a base field because the Views integration is broken.
+ * Instead, it is created as a configurable field for each order type.
+ *
+ * @param \Drupal\commerce_order\Entity\OrderTypeInterface $order_type
+ *   The order type.
+ */
+function commerce_order_add_adjustments_field($order_type) {
```
bojanz: Let's go with a base field. We can survive not having Views integration here for the longest, and the core patch is very close (we should push it to completion ourselves)

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+namespace Drupal\commerce_order;
+use Drupal\commerce_price\Price;
```
bojanz, line 5: Missing newline above use (so that phpcs let's us go)

modules/order/src/Adjustment.php
bojanz, line 34: Sounds like we need an expanded comment. Maybe something like "Points to the source object, if known. For example, a promotion entity for a discount adjustment".

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,126 @@
...
+  protected $sourceId;
+
+  /**
+   * Constructs a new Adjustment instance.
+   *
+   * @param array $definition
+   *   The definition.
+   */
+  public function __construct(array $definition) {
+    foreach (['type', 'label', 'amount', 'currency_code', 'source_id'] as $required_property) {
```
bojanz: The source_id isn't required, we said we wanted support for arbitrary adjustments.

modules/order/src/Adjustment.php
bojanz, line 59: I prefer it to be more explicit TBH. $this->{$required_property} is a bit of magic that will also set any misspelled or unknown keys.

modules/order/src/Entity/Order.php
bojanz, line 437: Indeed. We also want getAdjustments() / setAdjustments / removeAdjustment, right?

modules/order/src/Entity/OrderInterface.php
bojanz: removeAdjustment(). We have that pattern for plenty of other objects.

modules/order/src/Entity/OrderInterface.php
```diff
@@ -222,4 +223,23 @@ public function getBillingProfileId();
...
+   */
+  public function addAdjustment(Adjustment $adjustment);
+
+  /**
+   * Causes the order to recalculate its total.
+   *
+   * This will update the total price based on line item and adjustment prices.
+   */
+  public function calculateTotal();
```
bojanz: Doesn't feel like this should be a public interface method to me. It's more of an internal helper.

modules/order/src/Adjustment.php
bojanz: It's getDecimalAmount() on Price, Matt copied it to Adjustment for consistency. There we went with getDecimalAmount() to differentiate between amount the number and amount the object (e.g no $amount->getAmount()). So we probably need to leave this for now.

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
bojanz: There is nothing wrong with having specific columns for the adjustment, but I think it's more clear for AdjustmentItem to stand on its own, without people having to lookup PriceItem. We are talking about repeating 20 lines of code.

modules/price/src/Price.php
bojanz: Well, it's definitely a string cause that's what bcmath consumes. We want to start using the Calculator class instead of operators anyway. (as soon as we merge the Calculator class, that is. Wanted to see our needs evolve first)

modules/order/tests/src/Unit/AdjustmentTest.php
```diff
@@ -0,0 +1,67 @@
...
+      'currency_code' => 'USD',
+      'source_id' => 'promotion:1',
+    ];
+
+    $adjustment = new Adjustment($definition);
+    $this->assertEquals('discount', $adjustment->getType());
+    $this->assertEquals('10% off', $adjustment->getLabel());
+    $this->assertEquals(-1.00, $adjustment->getDecimalAmount());
+    $this->assertEquals('USD', $adjustment->getCurrencyCode());
+    $this->assertEquals('promotion:1', $adjustment->getSourceId());
```
bojanz: It's probably enough for the source id to be just 'one', the source object type is adjustment type specific. If you have a 'discount' adjustment, you know you want to load a promotion entity.

modules/order/src/Entity/Order.php
bojanz, line 108: We already have a PR open for that, #408

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,126 @@
...
+    return $this->sourceId;
+  }
+
+  /**
+   * Gets the Price value object for the adjustment.
+   *
+   * @return \Drupal\commerce_price\Price
+   *   The Price value object.
+   */
+  public function toPrice() {
```
bojanz: Feels weird to have this. If we need to use it to calculate adjustments, then maybe we have a trait with the add / subtract / multiple / divide methods and use it for both Price and Adjustment. TBD

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
bojanz: We don't want this method in any case. The adjustment should be all we need.

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
```diff
@@ -0,0 +1,99 @@
...
+        '_none' => $this->t('- Select -'),
+        // @todo Promotion
+        // We want a specific way to add a promotion to an order.
+        // Not sure how this would look. Do we have an AdjustmentType plugin
+        // Which provides the configuration form? Or is that insane.
+        // Or.. is a promotion a reference field all on its own and not
+        // shown here?
+        // 'discount' => $this->t('Apply promotion')
+        'order_adjustment' => $this->t('Apply order adjustment'),
+        // @todo: How do we expose line item level adjustments on order UI.
```
bojanz: There are two options here. 
One is to have $line_item->adjustments, storing the line item adjustments directly on the line item. That requires the adjustment UI to be directly on the OrderEditForm, where it gathers the adjustments from the line items and the order, allows them to be edited.
The second option is to have all adjustments on the order, and track to what they apply (I believe that's what Spree does). That would allow the UI to be in the widget.

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+namespace Drupal\commerce_order;
```
bojanz: Time to create a commerce_pricing module? We'll want pricing to work even without orders?

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,124 @@
...
+use Drupal\commerce_price\Plugin\Field\FieldType\PriceItem;
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\TypedData\DataDefinition;
+
+/**
+ * Plugin implementation of the 'commerce_price' field type.
+ *
+ * @todo See if bojanz throws bricks at me for extended PriceItem.
+ *
+ * @dearBojan: instead of just serializing a Adjustment object, put its
```
bojanz: Having individual columns is definitely more proper. The reason why I wanted to avoid this is because core has no way to maintain the schema, once we choose the columns, we're kinda stuck with them (or we need big migration code). Plus, we need to duplicate validation code between the AdjustmentItem and the Adjustment class. Meanwhile, I'm unsure we have any use of those columns in Views. We'll end up using custom formatters for the whole field anyway.

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
mglaman: Cool, I'll do the copy and paste. It seemed like a greatidea until it made the widget code look stupid.

modules/order/src/Adjustment.php
mglaman, line 59: Core has a mapping convention like this, for some config entities. I didn't bother since so small, and explicit made it more clear.

modules/order/src/Entity/Order.php
mglaman, line 437: Adding the same pattern to addLineItem, too (since same deal)

modules/order/src/Adjustment.php
rszrama: Then maybe we rename the property to $decimalAmount for clarity's sake?

modules/order/src/Adjustment.php
mglaman: That's what I was thinking, which is where the trait mentioned https://github.com/drupalcommerce/commerce/pull/447#discussion-diff-71970775 could go.

modules/order/src/Adjustment.php
mglaman, line 34: Added that secondary description.

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
mglaman: Before reading into some other things, initial thought:

We should have `$line_item->adjustments`, I just did not add it yet. But I feel that these should be applied through the promotion system and other code.

I agree with specifying the product adjustment on the order, then tracking where it applies. The problems I foresee are:
1. How do we store this "target", just as `target_id`? But that feels confusing when its just for the order. But could also work for shipments (possibly, maybe.) We can document the property as "the target for the adjustment, if not the order" and keep it optional
2. In order total calculation, we'll have to grab all adjustments that target product and take it into account.
3. How do we convey that this adjusts the _unit price_ and not the _total price_. I know offhand I might confuse it as adjusting the total price vs unit price.

modules/order/commerce_order.module
```diff
@@ -156,6 +156,7 @@ function commerce_order_add_line_items_field($order_type) {
   }
 }
 
+
```
mglaman: Extra newline

modules/order/src/AdjustableInterface.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+/**
+ * Defines an interface for objects that interact with Adjustments.
+ *
+ * @todo Provide a trait? But that'd assume the field is always 'adjustments'
```
mglaman: @bojanz thoughts? If we add commerce_pricing, we can put the interface an an AdjustableTrait, cuts down copy pasta a bit. Especially once we have shipments.

modules/order/src/Adjustment.php
mglaman: I figure this way it'd be easier to pass it to a calculator, instead of the adjustment doing the math

Edit: Adjustment knows it needs to go out and be calculated, it shouldn't know what its adjusting (amount wise)

modules/order/src/Entity/LineItem.php
```diff
@@ -271,7 +331,7 @@ public static function bundleFieldDefinitions(EntityTypeInterface $entity_type,
     else {
       // This line item type won't reference a purchasable entity. The field
       // can't be removed here, or converted to a configurable one, so it's
-      // hidden instead. See https://www.drupal.org/node/2346347#comment-10254087.
+      // hidden instead. https://www.drupal.org/node/2346347#comment-10254087.
```
mglaman: ¯\_(ツ)_/¯ coding standards wouldn't let me commit.

modules/order/src/Entity/Order.php
```diff
@@ -469,6 +519,18 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setDisplayConfigurable('view', TRUE)
       ->setSetting('workflow_callback', ['\Drupal\commerce_order\Entity\Order', 'getWorkflowId']);
 
+    $fields['adjustments'] = BaseFieldDefinition::create('commerce_order_adjustment')
+      ->setLabel(t('Adjustments'))
+      ->setDescription(t('Adjustments to the order total.'))
+      ->setRequired(FALSE)
+      ->setCardinality(BaseFieldDefinition::CARDINALITY_UNLIMITED)
+      ->setDisplayOptions('form', [
```
mglaman: Crud, we actually _don't_ want to expose the UI for this, line item adjustments are added via backend processes - like manual code or promotions sytem.

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
mglaman: @bojanz what if type was actually a class/interface name. 

So we'd have the following
- `\Drupal\..\OrderInterface::class` => 'Apply order adjustment'
- `\Drupal\..\LineItemInterface::class` => 'Apply product adjustment'

We'd then know the actual item it applies to instead of some magical string identifier. With that, `source_id` literally is the ID, and we could make an `adjustable` field which is the target ID. With order's it would just be the order ID and/or ignored.

modules/order/src/Adjustment.php
mglaman: Makes sense, however we should discuss this in general with the Price object; keep the naming consistent.

modules/order/src/AdjustableInterface.php
mglaman: I went to try this. The awkwardness isn't the field, since we could provide a method which provides the definition. But more so needing to calculate a total in the set methods. Felt weird to provide an `abstract protect function calculateTotals()` in the trait.

modules/order/src/Entity/LineItem.php
```diff
@@ -158,11 +159,58 @@ public function setData($data) {
...
+    $this->set('adjustments', $adjustments);
+    $this->calculateTotal();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function removeAdjustment($source_id) {
+    // @todo Implement, discuss if source ID is right approach. (it's not.)
```
bojanz: It's not. removeAdjustment(Adjustment $adjustment)

modules/order/src/AdjustableInterface.php
bojanz: Yeah, the need for calculateTotals on the order makes me thing we should not be abstracting this. At least not yet.

modules/order/src/Adjustment.php
bojanz: The calculator class works only with amounts, not price objects. So it doesn't matter here. We could replace 'amount' and 'currency_code' wit a single 'amount' that is a Price object, if that makes more sense? Would make it closer to line items.

modules/order/src/Entity/LineItem.php
mglaman: So `foreach` and then `===` and remove if match

modules/order/src/Adjustment.php
mglaman: I like that. It helps reinforce the use of our value objects

modules/order/src/AdjustableInterface.php
mglaman: Well I had to add the same methods on line items for add, set, and remove

mglaman:
@bojanz so with this we have the ability to adjust the price of an order and individual line items. We also have a rudimentary UI for adding adjustments to an order. 

Do you think we could merge this, and then make a follow-up for allowing the UI to support line item level adjustments?

EDIT: Actually, we need to provide a formatter oh some sort. Probably a table of "Title: Amount"

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,62 @@
...
+
+use Drupal\commerce_order\Adjustment;
+use Drupal\Core\Field\FieldItemBase;
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\TypedData\MapDataDefinition;
+
+/**
+ * Plugin implementation of the 'commerce_price' field type.
+ *
+ * @dearBojan: instead of just serializing a Adjustment object, put its
```
bojanz: The dear bojan part is now outdated. The name of the field type is wrong.

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,62 @@
...
+ *              definition items as the schema, and it can return. This way
+ *              we could actually expose the data to Views and others.
+ *
+ * @FieldType(
+ *   id = "commerce_order_adjustment",
```
bojanz: Shouldn't the plugins be just commerce_adjustment, not commerce_order_adjustment? We are going to use them for line items too, no?

modules/order/src/Plugin/DataType/Adjustment.php
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Provides an adjustment data type.
+ *
+ * @DataType(
+ *   id = "adjustment",
+ *   label = @Translation("Adjustment")
+ * )
+ */
+class Adjustment extends TypedData {
```
bojanz: Why do we need this? can't we do the same trick in the FieldItem class?

modules/order/src/Plugin/Field/FieldFormatter/AdjustmentDefaultFormatter.php
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @FieldFormatter(
+ *   id = "commerce_order_adjustment_default_formatter",
+ *   label = @Translation("Order adjustments"),
+ *   field_types = {
+ *     "commerce_order_adjustment",
+ *   },
+ * )
+ */
+class AdjustmentDefaultFormatter extends FormatterBase implements ContainerFactoryPluginInterface {
```
bojanz: Not sure I see a point in this class. Remember that a list of adjustments does not represent the order total summary (cause that's the line item adjustments + the order adjustments)

mglaman:
There's an errror: Drupal\Core\Field\FieldException: Attempt to create field storage adjustments which is reserved by entity type commerce_order.

modules/order/src/Plugin/DataType/Adjustment.php
mglaman: It's not just the trick, for setting the value but also the storage of the item

modules/order/src/AdjustableInterface.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+/**
+ * Defines an interface for objects that interact with Adjustments.
+ *
+ * @see \Drupal\commerce_order\Entity\Order
```
bojanz: Would it make more sense to point to OrderInterface / LineItemInterface?

modules/order/src/AdjustableInterface.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+/**
+ * Defines an interface for objects that interact with Adjustments.
```
bojanz: interact or contain? also, adjustments probably don't need to be capitalized

modules/order/src/AdjustableInterface.php
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Adds an adjustment to the order.
+   *
+   * @param \Drupal\commerce_order\Adjustment $adjustment
+   *   The adjustment.
+   *
+   * @return $this
+   */
+  public function addAdjustment(Adjustment $adjustment);
```
bojanz: The correct order should probably be get / set / add / remove, no? We also usually have a has method as well, dunno if needed here.

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,108 @@
...
+   * @var string
+   */
+  protected $label;
+
+  /**
+   * The price value object.
+   *
+   * @var \Drupal\commerce_price\Price
+   */
+  protected $price;
```
bojanz: Wouldn't $price make a lot more sense as $amount? That's what we use on payments

modules/order/src/Entity/LineItem.php
```diff
@@ -158,11 +159,59 @@ public function setData($data) {
...
+  /**
+   * {@inheritdoc}
+   */
   public function preSave(EntityStorageInterface $storage) {
     parent::preSave($storage);
+    $this->calculateTotal();
+  }
 
+  /**
+   * Causes the line item to recalculate its total.
```
bojanz: Causes the... is weird. Why not just "Recalculates the line item total price"? Also, if the getter is getTotalPrice, shouldn't this be recalculateTotalPrice()?

modules/order/src/Form/OrderTypeForm.php
```diff
@@ -6,6 +6,9 @@
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 
+/**
+ * Form for order types.
```
bojanz: This change is not needed, and the comment doesn't match what we usually do ("Provides an order type form.").

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,67 @@
...
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\TypedData\DataDefinition;
+
+/**
+ * Plugin implementation of the 'commerce_adjustment' field type.
+ *
+ * @FieldType(
+ *   id = "commerce_adjustment",
+ *   label = @Translation("Adjustment"),
+ *   description = @Translation("Stores an adjustment which contains a decimal amount and a three letter currency code and other metadata."),
```
bojanz: Doesit matter what the adjustment contains? "Stores adjustments to the parent entity's price."

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -0,0 +1,67 @@
...
+ * )
+ */
+class AdjustmentItem extends FieldItemBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function propertyDefinitions(FieldStorageDefinitionInterface $field_definition) {
+    $properties['value'] = DataDefinition::create('any')
+      ->setLabel(t('Adjustment value object.'))
```
bojanz: "Value" is probably good enough as a label? The column is definitely not storing an object, if we're to be precise.

modules/order/src/Plugin/Field/FieldType/AdjustmentItemList.php
```diff
@@ -0,0 +1,39 @@
...
+ * Represents a list of adjustment item field values.
+ */
+class AdjustmentItemList extends FieldItemList implements AdjustmentItemListInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAdjustments() {
+    $adjustments = [];
+
```
bojanz: Unneeded newline.

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
```diff
@@ -0,0 +1,218 @@
...
+use Drupal\Core\Field\WidgetBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Plugin implementation of 'commerce_adjustment_default_widget'.
+ *
+ * @FieldWidget(
+ *   id = "commerce_adjustment_default_widget",
```
bojanz: The id should be "commerce_adjustment_default", we never add "_widget" to widget ids.

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
```diff
@@ -0,0 +1,218 @@
...
+        ],
+      ],
+    ];
+    $element['definition']['label'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Label'),
+      '#default_value' => ($adjustment) ? $adjustment->getLabel() : '',
+    ];
+
+    $element['definition']['amount'] = [
```
bojanz: 8.x-2.x now has a commerce_price form element you can use.

modules/order/src/AdjustableInterface.php
mglaman: Yeah, the example is outdated.

modules/order/src/AdjustableInterface.php
mglaman: Yeah, contained.

modules/order/src/Adjustment.php
mglaman: I feel like it was, then @rszrama made a comment about amount because its actually a price object

modules/order/src/Form/OrderTypeForm.php
mglaman: I think I had to add it because of phpcs

modules/order/src/Entity/LineItem.php
mglaman: Fixed it for order, too.

bojanz:
Do you really want to continue? (y/n): y
exception 'Drupal\Component\Plugin\Exception\PluginNotFoundException'[error]
with message 'The "commerce_adjustment_default_widget" plugin does
not exist.' in

modules/order/src/Adjustment.php
bojanz: if Price $amount is confusing, then that means we need to rename the Price object :)
Feels odd to have an adjustment price cause an adjustment is not something you buy.

modules/order/src/Plugin/Field/FieldWidget/AdjustmentDefaultWidget.php
```diff
@@ -0,0 +1,97 @@
...
+  public function massageFormValues(array $values, array $form, FormStateInterface $form_state) {
+    foreach ($values as $key => $value) {
+      if ($value['type'] == '_none') {
+        continue;
+      }
+
+      $values[$key] = new Adjustment([
+        'type' => $value['type'],
+        'label' => $value['definition']['label'],
+        'price' => new Price($value['definition']['amount'], $value['definition']['currency_code']),
```
bojanz: This line looks unconverted, where would it get the currency_code from?
$element['definition']['amount']['#value'] will give you the full price object (unsure why it's not in $form_state->getValue(), a bug I've yet to catch.)

mglaman: Oh shoot, I need to update that, then.

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * The adjustment label.
+   *
+   * @var string
+   */
+  protected $label;
+
+  /**
+   * The price value object.
```
bojanz, line 27: The adjustment amount.

modules/order/src/Adjustment.php
```diff
@@ -0,0 +1,108 @@
...
+      }
+    }
+    if (!$definition['amount'] instanceof Price) {
+      throw new \InvalidArgumentException(sprintf('Property amount should be an instance of %s.', Price::class));
+    }
+
+    $this->type = $definition['type'];
+    $this->label = $definition['label'];
+    $this->amount = $definition['amount'];
+
```
bojanz, line 62: Why the newline?

modules/order/src/AdjustableInterface.php
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Defines an interface for objects that contain adjustments.
+ *
+ * @see \Drupal\commerce_order\Entity\OrderInterface
+ * @see \Drupal\commerce_order\Entity\LineItemInterface
+ */
+interface AdjustableInterface {
+
+  /**
+   * Gets the order's adjustments.
```
bojanz, line 14: D'oh, all docblocks mention order.

modules/order/src/AdjustableInterface.php
```diff
@@ -0,0 +1,53 @@
...
+   *   The adjustment.
+   *
+   * @return $this
+   */
+  public function addAdjustment(Adjustment $adjustment);
+
+  /**
+   * Removes an adjustment.
+   *
+   * The adjustment is removed by targeting the source identifier.
```
bojanz, line 44: No it's not.

modules/order/src/Adjustment.php
mglaman, line 62: I like newlines.
\n\n\n\n\n\n\n\n\nn\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n

bojanz:
Committed in 83e6660a24d846bfe56e55de4a111f3f2c2a3556.

PR #448, opened 2016-07-25
Issue #2724059 by mglaman, PieterJanPut, bojanz: Cart block quantity …

mglaman:
…text is not translatable

### DIFF
--- a/modules/cart/src/Plugin/Block/CartBlock.php
+++ b/modules/cart/src/Plugin/Block/CartBlock.php
@@ -7,6 +7,7 @@ use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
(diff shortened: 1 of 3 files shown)


modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -74,7 +75,7 @@ public static function create(ContainerInterface $container, array $configuratio
    */
   public function defaultConfiguration() {
     return [
-      'item_text' => $this->t('items'),
+      'item_text' => 'item' . PluralTranslatableMarkup::DELIMITER . 'items',
```
mglaman, line 78: I'm unsure if this is to be passed into translation here or not.

mglaman, line 78: Went to, and got code violation warnings for concatenating inside of a translate function. There's no core pattern here [that I can see] to follow.

mglaman, line 78: @bojanz I asked around, there is no standard way, especially with config. I think this is good to go.

bojanz:
No longer relevant. See https://www.drupal.org/node/2724059#comment-11538119

PR #450, opened 2016-07-25
Issue #2762995 by mglaman, dylanf: Add a description/message field

mglaman:


modules/promotion/src/Entity/Promotion.php
```diff
@@ -240,6 +255,21 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setDisplayConfigurable('view', TRUE)
       ->setDisplayConfigurable('form', TRUE);
 
+    $fields['description'] = BaseFieldDefinition::create('string_long')
+      ->setLabel(t('Description'))
+      ->setDescription(t('The description of the promotion. This is additional information, beyond the name, which can be displayed to the customer.'))
```
mglaman: @bojanz description look "good enough"?

borisson:
I think this looks great.

modules/promotion/src/Entity/Promotion.php
bojanz: "The description of the promotion." is redundant, it doesn't say anything that the "Description" field name doesn't already say. We remove that, it leaves us with "This is additional information, beyond the name, which can be displayed to the customer.". Maybe we can reword that as "Used to describe the promotion to the customer."?  Let's bikeshed on IRC.

mglaman: So Woo, Spree, etc don't have a helper label. "Additional information about the promotion to show to the customer"

mglaman:
```
An explanation of the promotion. The customer will be able to see this description at checkout.
```

@bojanz ^ yeah?

mglaman:
Per @bojanz disccusion "Additional information about the promotion to show to the customer"

PR #451, opened 2016-07-29
Issue #2656896 Product revisions

czigor:
https://www.drupal.org/node/2656896
First stab, untested. Not worth looking at it yet.

### DIFF
--- a/modules/product/src/Entity/Product.php
+++ b/modules/product/src/Entity/Product.php
@@ -45,8 +45,11 @@ use Drupal\Core\Field\BaseFieldDefinition;
  *   translatable = TRUE,
  *   base_table = "commerce_product",
  *   data_table = "commerce_product_field_data",
+ *   revision_table = "commerce_product_revision",
+ *   revision_data_table = "commerce_product_field_revision",
  *   entity_keys = {
  *     "id" = "product_id",
+ *     "revision" = "product_vid",
  *     "bundle" = "type",
  *     "label" = "title",
  *     "langcode" = "langcode",
@@ -61,6 +64,8 @@ use Drupal\Core\Field\BaseFieldDefinition;
  *     "delete-form" = "/product/{commerce_product}/delete",
  *     "delete-multiple-form" = "/admin/commerce/products/delete",
  *     "collection" = "/admin/commerce/products"
+ *     "version-history" = "/product/{commerce_product}/revisions",
+ *     "revision" = "/product/{commerce_product}/revisions/{commerce_product_revision}/view",
  *   },
  *   bundle_entity_type = "commerce_product_type",
  *   field_ui_base_route = "entity.commerce_product_type.edit_form",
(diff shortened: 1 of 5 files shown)


modules/product/src/Entity/Product.php
```diff
@@ -285,6 +290,43 @@ protected function ensureTranslations(array $entities) {
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function preSaveRevision(EntityStorageInterface $storage, \stdClass $record) {
+    parent::preSaveRevision($storage, $record);
+
+    if (!$this->isNewRevision() && isset($this->original) && (!isset($record->revision_log) || $record->revision_log === '')) {
+      // If we are updating an existing node without adding a new revision, we
```
borisson: s/node/product/

modules/product/src/Entity/ProductType.php
```diff
@@ -74,6 +75,13 @@ class ProductType extends ConfigEntityBundleBase implements ProductTypeInterface
   protected $description;
 
   /**
+   * Default value of the 'Create new revision' checkbox of this node type.
```
borisson: s/node/product/

modules/product/src/Form/ProductForm.php
```diff
@@ -84,6 +97,41 @@ public function form(array $form, FormStateInterface $form_state) {
...
+      '#type' => 'details',
+      '#group' => 'advanced',
+      '#title' => t('Revision information'),
+      // Open by default when "Create new revision" is checked.
+      '#open' => $product->isNewRevision(),
+      '#attributes' => array(
+        'class' => array('node-form-revision-information'),
+      ),
+      '#attached' => array(
+        'library' => array('node/drupal.node'),
```
borisson: If we really need this, we should depend on node from the product module. As far as I can tell, we only need one part of the javascript related to this library, so we should probably copy only that part over into a new library?

modules/product/src/Form/ProductForm.php
```diff
@@ -84,6 +97,41 @@ public function form(array $form, FormStateInterface $form_state) {
...
+      ),
+      '#weight' => 20,
+      '#optional' => TRUE,
+    );
+
+    $form['revision'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Create new revision'),
+      '#default_value' => $product->type->entity->isNewRevision(),
+      '#access' => $current_user->hasPermission('administer nodes'),
```
borisson: s/nodes/products/

modules/product/src/Form/ProductForm.php
```diff
@@ -270,6 +318,32 @@ protected function actions(array $form, FormStateInterface $form_state) {
 
   /**
    * {@inheritdoc}
+   *
+   * Updates the node object by processing the submitted values.
+   *
+   * This function can be called by a "Next" button of a wizard to update the
+   * form state's entity with the current step's values before proceeding to the
+   * next step.
+   */
```
borisson, line 320: It's not allowed to have both `{@inheritdoc}` and documentation in docblock, so we should either remove this comment or copy the relevant portion of the parent's docblock over.

modules/product/src/Form/ProductForm.php
```diff
@@ -270,6 +318,32 @@ protected function actions(array $form, FormStateInterface $form_state) {
...
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    // Build the node object from the submitted values.
```
borisson: s/node/product/

czigor:
Ready for review now.

modules/product/src/Entity/ProductType.php
```diff
@@ -90,6 +98,20 @@ class ProductType extends ConfigEntityBundleBase implements ProductTypeInterface
...
    * {@inheritdoc}
    */
+  public function isNewRevision() {
+    return $this->new_revision;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setNewRevision($new_revision) {
```
andypost, line 108: no external interfaces changed so where this method from?

bojanz:
The issue (http://drupal.org/node/2656896) has a newer patch, let's continue there.

PR #453, opened 2016-07-30
Issue 2772253: Add link forgotten password in error message in Login checkout pane

aristova:
Issue
https://www.drupal.org/node/2772253

### DIFF
--- a/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
+++ b/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
@@ -243,7 +243,7 @@ class Login extends CheckoutPaneBase implements CheckoutPaneInterface, Container
     $username = $values['returning_customer']['name'];
     $password = trim($values['returning_customer']['password']);
     // Setting url for "forgotten password" link in the error message.
-    $route_parameters = !empty($username) ? array('name' => $username) : array();
+    $route_parameters = !empty($username) ? ['name' => $username] : [];
     $route = Url::fromRoute('user.pass', [], ['query' => $route_parameters])->toString();
     if (empty($username) || empty($password)) {
       $form_state->setErrorByName('name', $this->t('Unrecognized username or password. <a href=":url">Have you forgotten your password?</a>', [':url' => $route]));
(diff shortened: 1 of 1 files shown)


modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -243,8 +242,11 @@ public function validatePaneForm(array &$pane_form, FormStateInterface $form_sta
     $values = $form_state->getValue($pane_form['#parents']);
     $username = $values['returning_customer']['name'];
     $password = trim($values['returning_customer']['password']);
+    // Setting url for "forgotten password" link in the error message.
+    $route_parameters = !empty($username) ? array('name' => $username) : array();
```
borisson: With all other changes lines in this PR are short array syntax, but this line uses long syntax, let's change that as well.

modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -243,8 +242,11 @@ public function validatePaneForm(array &$pane_form, FormStateInterface $form_sta
     $values = $form_state->getValue($pane_form['#parents']);
     $username = $values['returning_customer']['name'];
     $password = trim($values['returning_customer']['password']);
+    // Setting url for "forgotten password" link in the error message.
+    $route_parameters = !empty($username) ? ['name' => $username] : [];
```
bojanz, line 246: $route_parameters should be $query. The real route parameters are the empty array passed to Url.

modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -243,8 +242,11 @@ public function validatePaneForm(array &$pane_form, FormStateInterface $form_sta
...
+    $route = Url::fromRoute('user.pass', [], ['query' => $route_parameters])->toString();
```
bojanz, line 247: $route should be $password_url, since we're not getting a route.

bojanz:
Addressed my own feedback, committed in 283dd88ec7b91fdedbf2913b681e6036370bc31f. Thanks!

PR #457, opened 2016-08-01
Limit payment method list builder to user in route

bradjones1:
Unlike most other list builders, which are administrative, this one is context-specific to the route.

### DIFF
--- a/modules/payment/src/PaymentMethodListBuilder.php
+++ b/modules/payment/src/PaymentMethodListBuilder.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Drupal\commerce_payment;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityListBuilder;
+
+/**
+ * Defines the list builder for payment methods.
+ */
+class PaymentMethodListBuilder extends EntityListBuilder {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $entitiesKey = 'methods';
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'commerce_payment_methods';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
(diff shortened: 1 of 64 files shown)


bradjones1:
https://www.drupal.org/node/2776417

modules/payment/src/PaymentMethodListBuilder.php
```diff
@@ -22,6 +23,16 @@ public function getFormId() {
     return 'commerce_payment_methods';
   }
 
+  protected function getEntityIds() {
```
bojanz, line 28: The parent method also has support for paging, we should recopy that code:
    // Only add the pager if a limit is specified.
    if ($this->limit) {
      $query->pager($this->limit);
    }

We also need a docblock with an inheritdoc inside.

modules/payment/src/PaymentMethodListBuilder.php
```diff
@@ -22,6 +23,16 @@ public function getFormId() {
...
+    /** @var RouteMatchInterface $route */
```
borisson: Not sure how strict @bojanz is with coding standards, but this should be the FQN: 
`/** @var \Drupal\Core\Routing\RouteMatchInterface $route */`

modules/payment/src/PaymentMethodListBuilder.php
```diff
@@ -4,6 +4,7 @@
 
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityListBuilder;
+use Drupal\Core\Routing\RouteMatchInterface;
```
borisson: Is this use still needed now that the `@var` no longer uses the imported interface?

borisson:
This looks great now.

bojanz:
Squashed this into payment in preparation for merge. Gave you commit credit on a different commit in return: fdc891093fe4006698b4ab8c6e467a7f85f183aa

PR #458, opened 2016-08-03
Test for Credit card type class

borisson:


bojanz:
Committed a finished version.

PR #459, opened 2016-08-03
Add test for CreditCard class

borisson:


modules/payment/tests/src/Unit/CreditCardTest.php
```diff
@@ -0,0 +1,112 @@
...
+class CreditCardTest extends \PHPUnit_Framework_TestCase {
+
+  /**
+   * @covers ::getTypes
+   */
+  public function testGetTypes() {
+    $types = CreditCard::getTypes();
+    $this->assertInternalType('array', $types);
+    foreach ($types as $key => $cardType) {
+      $this->assertInstanceOf(CreditCardType::class, $cardType);
```
bojanz: Accidental camelCase. Let's just use $type instead of $cardType.

modules/payment/src/CreditCard.php
```diff
@@ -42,48 +44,48 @@ public static function getTypes() {
     $definitions = [
       'visa' => [
         'id' => 'visa',
-        'label' => t('Visa'),
+        'label' => new TranslatableMarkup('Visa'),
```
bojanz, line 47: Makes sense. Using t() was my mistake.

modules/payment/tests/src/Unit/CreditCardTest.php
```diff
@@ -0,0 +1,112 @@
...
+    $this->assertInstanceOf(CreditCardType::class, $visa);
+    $this->assertEquals('visa', $visa->getId());
+  }
+
+  /**
+   * @covers ::detectType
+   * @covers ::matchPrefix
+   * @dataProvider cardsProvider
+   */
+  public function testDetectType($number, $cardType) {
```
bojanz: $cardType should be $type.

modules/payment/tests/src/Unit/CreditCardTest.php
```diff
@@ -0,0 +1,112 @@
...
+    // Card number has to be an integer.
+    $card = CreditCard::detectType('4111 1111 1111 1111');
```
bojanz: Feels odd to test the non-integer number for every $number / $type combination.
I suggest passing a $valid variable from the data provider, then extending the data provider to pass a few invalid values (non-integer, number too long, right prefix but wrong number, unrecognized type).

bojanz:
Finished and committed db2f146a309bd8d0111e7334e6d7827948b4409a

PR #462, opened 2016-08-04
Currency description

borisson:
Fixes the comment in https://github.com/drupalcommerce/commerce/pull/435

### DIFF
--- a/modules/price/src/Form/CurrencyForm.php
+++ b/modules/price/src/Form/CurrencyForm.php
@@ -74,7 +74,7 @@ class CurrencyForm extends EntityForm {
     $form['numericCode'] = [
       '#type' => 'textfield',
       '#title' => $this->t('Numeric code'),
-      '#description' => $this->t('The three digit code, as defined by <a href="!url" target="_blank">ISO 4217</a>.', ['!url' => $iso_4217_url]),
+      '#description' => $this->t('The three digit code, as defined by <a href=":url" target="_blank">ISO 4217</a>.', [':url' => $iso_4217_url]),
       '#default_value' => $currency->getNumericCode(),
       '#element_validate' => ['::validateNumericCode'],
       '#pattern' => '[\d]{3}',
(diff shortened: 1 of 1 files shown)


modules/price/src/Form/CurrencyForm.php
```diff
@@ -43,6 +44,12 @@ public function form(array $form, FormStateInterface $form_state) {
     $form = parent::form($form, $form_state);
     $currency = $this->entity;
 
+    $import_url = Url::fromRoute('entity.commerce_currency.import')->toString();
```
bojanz, line 47: Perfect!

modules/price/src/Form/CurrencyForm.php
```diff
@@ -53,6 +60,7 @@ public function form(array $form, FormStateInterface $form_state) {
     $form['currencyCode'] = [
       '#type' => 'textfield',
       '#title' => $this->t('Currency code'),
+      '#description' => $this->t('The uppercase alphabetic currency code. For example USD.'),
```
bojanz: Do we need "The uppercase alphabetic currency code" if we have a clear example that communicates the whole thing?

Example needs interpunction:
"For example: USD." 
(unsure if "For example: $example." is more common than "For example, $example.")

modules/price/src/Form/CurrencyForm.php
```diff
@@ -65,6 +73,9 @@ public function form(array $form, FormStateInterface $form_state) {
     $form['numericCode'] = [
       '#type' => 'textfield',
       '#title' => $this->t('Numeric code'),
+      '#description' => $this->t('The numeric currency code. According to ISO 4217 this code consists of three digits
```
bojanz: "The numeric currency code." is redundant, just repeats the element title. 
Maybe just "The three digit code as defined by ISO 4217."?

bojanz:
Exception: User error: Invalid placeholder (!url) in string: This form is only intended to be used to create custom currencies. Real-world currencies <a href="!url">should be imported</a>

PR #463, opened 2016-08-04
Add tests for the checkout flow UI

borisson:


modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowWithPanesBase.php
```diff
@@ -360,7 +360,7 @@ protected function buildPaneRow(CheckoutPaneInterface $pane, array &$form, FormS
         '#attributes' => [
           'class' => ['pane-configuration-edit-form'],
         ],
-        '#element_validate' => [get_class($this), 'validatePaneConfigurationForm'],
+        '#element_validate' => [static::class . '::validatePaneConfigurationForm'],
```
bojanz: Why is this needed? We use get_class($this) all over the place

PR #464, opened 2016-08-11
Issue 2782489: Fix return  in buildPaneForm().

aristova:
Method Drupal\commerce_checkout\Plugin\Commerce\CheckoutPane\CompletionMessage:buildPaneForm
receives array $pane_form with some values ('#parents', '#access'...), but returns array with just one key-value pair ( '#markup' => 'text of message').
I think it's a bug, because we can't use method isVisible(), for example.


bojanz:
Good catch, no code should override the provided form array.

PR #402, opened 2016-06-13
Issue #2718835 by skyredwang, mglaman: Make ProductAttribute getValue…

mglaman:
… sorted by weights

PR #465, opened 2016-08-11
Issue #2718835 by mglaman, skyredwang: Make ProductAttribute getValue sorted by weights

bojanz:
Continuation of #402.

Fixed the CS errors and this:

```
   public function getValues() {
     $storage = $this->entityTypeManager()->getStorage('commerce_product_attribute_value');
-    return $storage->loadByAttribute(['attribute' => $this->id]);
+    return $storage->loadByAttribute($this->id());
   }
```

Rewrote the #value fix in ProductAttributeForm to be more clean.

### DIFF
--- a/modules/product/tests/src/FunctionalJavascript/ProductAttributeJavascriptTest.php
+++ b/modules/product/tests/src/FunctionalJavascript/ProductAttributeJavascriptTest.php
@@ -2,7 +2,6 @@
 
 namespace Drupal\Tests\commerce_product\FunctionalJavascript;
 
-use Drupal\commerce_product\Entity\ProductAttribute;
 use Drupal\Tests\commerce\FunctionalJavascript\JavascriptTestTrait;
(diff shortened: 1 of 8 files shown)


modules/product/tests/src/FunctionalJavascript/ProductAttributeJavascriptTest.php
```diff
@@ -69,26 +63,47 @@ public function testProductAttributeValues() {
...
+  /**
+   * Helper methods to get an attributes value labels, in order by weight.
+   *
+   * @param string $name
+   *   The attribute name.
+   *
+   * @return \Drupal\commerce_product\Entity\ProductAttributeValueInterface[]
+   *   Array of attribute value labels, in order of weight.
+   */
+  protected function reloadAttributeValues($name) {
```
bojanz: This method does not do what the docs say it does.

bojanz:
Final version up, can be merged when green.

PR #468, opened 2016-08-17
Make it possible to add a line item to cart that doesn't have a product

googletorp:
This add flexibility, since the line item doesn't require a PurchasableEntityInterface to be set on the line item, the Event for adding it to cart shouldn't require it as well.

Fix getEntity method on CartEntityAddEvent


bojanz:
Replaced by #470.

PR #470, opened 2016-08-18
Don't dispatch CartEntityAddEvent if there is no entity

googletorp:


mglaman:
Can we get a kernel test for this? We'll need a test module which listens on the event.

bojanz:
This was committed in b26425d9218cae00cd281640000b28af48169b6a

PR #475, opened 2016-08-27
2770757 implement promotion conditions

mglaman:


mglaman:
Branched off of #445

modules/promotion/src/Plugin/Commerce/PromotionCondition/OrderAmountComparison.php
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form += parent::buildConfigurationForm($form, $form_state);
+
+    $form['amount'] = [
+      '#type' => 'commerce_price',
```
mglaman: Need to extend FunctionalJavascript test for this.

modules/promotion/src/Plugin/Commerce/PromotionCondition/OrderAmountComparison.php
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function evaluate() {
+    /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
+    $order = $this->getTargetEntity();
+    /** @var \Drupal\commerce_price\Price $total_price */
+    $total_price = $order->getTotalPrice();
+    /** @var \Drupal\commerce_price\Price $comparison_price */
+    $comparison_price = $this->configuration['amount'];
```
mglaman: Need to extend FunctionalJavascript to make sure this works for retrieving value. Does in code.

modules/promotion/src/Plugin/Commerce/PromotionCondition/PromotionConditionBase.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionCondition;
+
+use Drupal\Core\Condition\ConditionPluginBase;
+
+abstract class PromotionConditionBase extends ConditionPluginBase implements PromotionConditionInterface {
```
mglaman, line 10: Needs docblock.

modules/promotion/src/Plugin/Commerce/PromotionCondition/PromotionConditionInterface.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionCondition;
+
+use Drupal\Core\Condition\ConditionInterface;
+
+interface PromotionConditionInterface extends ConditionInterface {
```
mglaman, line 10: Needs docblock

modules/promotion/src/PromotionConditionManager.php
```diff
@@ -0,0 +1,88 @@
...
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Executable\ExecutableManagerInterface;
+use Drupal\Core\Executable\ExecutableInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Plugin\CategorizingPluginManagerTrait;
+use Drupal\Core\Plugin\Context\ContextAwarePluginManagerTrait;
+use Drupal\Core\Plugin\DefaultPluginManager;
+
+
+class PromotionConditionManager extends DefaultPluginManager implements ExecutableManagerInterface, CategorizingPluginManagerInterface {
```
mglaman, line 23: Needs docblock

modules/promotion/src/Entity/Promotion.php
```diff
@@ -320,6 +334,15 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
         'weight' => 3,
       ]);
 
+    $fields['conditions'] = BaseFieldDefinition::create('commerce_plugin_item:commerce_promotion_condition')
+      ->setLabel(t('Offer'))
```
bojanz: Wong label.

modules/promotion/src/Entity/Promotion.php
```diff
@@ -243,9 +243,23 @@ public function apply(EntityInterface $entity) {
     $entity_type_id = $entity->getEntityTypeId();
     // @todo should whatever invokes this method be providing the context?
     $context = new Context(new ContextDefinition('entity:' . $entity_type_id), $entity);
+
+    // Execute each plugin, this is an AND operation.
+    // @todo support OR operations.
+    /** @var \Drupal\commerce\Plugin\Field\FieldType\PluginItem $item */
+    foreach ($this->get('conditions') as $item) {
```
bojanz, line 250: I would expect the conditions to be evaluated in another method, like applies(), and I would expect apply() to always work. I don't think the API consumers would blindly try to apply promotions without checking if they apply first?

modules/promotion/src/Plugin/Commerce/PromotionCondition/PromotionConditionInterface.php
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Defines an interface for Condition plugins.
+ *
+ * Extends the core Condition interface to support Commerce Promotion specific
+ * requirements.
+ */
+interface PromotionConditionInterface extends ConditionInterface {
+
+  /**
+   * Returns what entity the promotion offer applies to.
```
bojanz: We use "Gets" instead of "Returns". And this returns the entity type, not the entity. Perhaps appliesTo() is not descriptive enough? getTargetEntityType() would not require me to read the docblock.

modules/promotion/src/Plugin/Commerce/PromotionCondition/OrderAmountComparison.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionCondition;
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides a 'Order: Percentage off' condition.
```
bojanz: Wrong docblock.

modules/promotion/src/Plugin/Commerce/PromotionCondition/OrderAmountComparison.php
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @PromotionCondition(
+ *   id = "commerce_promotion_order_amount_comparison",
+ *   label = @Translation("Order Amount Comparison"),
+ *   target_entity_type = "commerce_order",
+ * )
+ */
+class OrderAmountComparison extends PromotionConditionBase {
```
bojanz: Orders don't have an amount, they have a total price. So the class should be OrderTotalPrice, no?

modules/promotion/src/Entity/Promotion.php
mglaman: That makes sense.. so you could choose to explicitly apply a promotion regardless. `applies` and `apply` are two methods, validation should be done by end implementer.

smmccabe: If a promotion applies that doesn't match the rules, that is going to confuse the hell out of any admins though. If a API user wants to apply a promotion outside of conditions, wouldn't it be better for them to make a new promotion? Also any re-validation of the order based on a change would cause it to be dumped? either way it would be confusing.

mglaman: > If a promotion applies that doesn't match the rules, that is going to confuse the hell out of any admins though. I

That's a level they'd never see.

> If a API user wants to apply a promotion outside of conditions, wouldn't it be better for them to make a new promotion

Possibly. 

> Also any re-validation of the order based on a change would cause it to be dumped

That's all TBD, but true. I'm sure next validation round the conditions would fail and it'd be removed. Not our problem

smmccabe: Whats the point of splitting it up though then? aren't we just allowing an unnecessarily dangerous thing?

mglaman: True. What if we wanted to check if a promotion "just applies" We'd want to have `applies` then. Should `apply` also then wrap the `applies` method?

Either way, no one should be loading a promotion directly and running these. There will be a service you pass an entity to, and that entity is processed.

smmccabe: Hmm yah, there should be an applies stand alone, lots of use cases for that. I think apply should include it, but you're right, it is only for preventing someone from doing something stupid.

modules/order/src/Entity/LineItem.php
```diff
@@ -102,7 +102,7 @@ public function setTitle($title) {
    * {@inheritdoc}
    */
   public function getQuantity() {
-    return $this->get('quantity')->value;
+    return (string) $this->get('quantity')->value;
```
mglaman, line 105: @bojanz so should we change the field type to be string, or store it as decimal and keep this cast?

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
```diff
@@ -42,6 +42,10 @@ public function isEmpty() {
    * {@inheritdoc}
    */
   public function setValue($values, $notify = TRUE) {
+    if (is_array($values)) {
```
mglaman, line 45: This should probably be fixed in 8.x-2.x HEAD not just here.

mglaman, line 45: Relevant D.o issue: https://www.drupal.org/node/2791845

modules/price/src/Element/Price.php
```diff
@@ -36,7 +36,7 @@ public function getInfo() {
     $class = get_class($this);
     return [
       '#size' => 10,
-      '#maxlength' => 128,
+      '#maxlength' => NULL,
```
mglaman: Having maxlength causes utility::strlen to call which gets passed an array. @bojanz this should be fixed in 8.x-2.x HEAD right, first?

modules/price/src/Element/Price.php
```diff
@@ -61,9 +61,9 @@ public function getInfo() {
    * {@inheritdoc}
    */
   public static function valueCallback(&$element, $input, FormStateInterface $form_state) {
-    if ($input !== FALSE && $input !== NULL) {
+    if ($input !== FALSE && $input !== NULL && !empty($input['amount']) && !empty($input['currency_code'])) {
```
mglaman: This prevents null values from being passed in. We should fix this in 8.x-2.x HEAD right?

modules/price/src/Element/Price.php
```diff
@@ -61,9 +61,9 @@ public function getInfo() {
...
       // Convert empty string value to numeric value.
-      if ($input['amount'] === '') {
+      if ($input['amount'] == '') {
```
mglaman: Undo this change.

modules/order/src/Plugin/Field/FieldType/AdjustmentItem.php
mglaman: This is being addressed in #476

modules/price/src/Element/Price.php
mglaman: Addressed in https://github.com/drupalcommerce/commerce/pull/477

modules/price/src/Element/Price.php
mglaman: Addressed in https://github.com/drupalcommerce/commerce/pull/478

modules/order/src/Entity/LineItem.php
bojanz: It would be more proper to change a field type, but that will probably require providing a custom widget. We can go with the cast and open a followup issue.

modules/promotion/src/Entity/PromotionInterface.php
```diff
@@ -214,7 +214,18 @@ public function isEnabled();
...
 
   /**
-   * Applies the promotion to an entity.
+   * Checks if the promotion entity can be applied.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity.
+   *
+   * @return bool
+   *   Returns TRUE if promotion can be applied, or false if conditions failed.
```
bojanz: We don't need "Returns", that's redundant. Just "TRUE if the promotion can be applied, FALSE otherwise."

modules/promotion/src/Entity/PromotionInterface.php
```diff
@@ -214,7 +214,18 @@ public function isEnabled();
   public function setEnabled($enabled);
 
   /**
-   * Applies the promotion to an entity.
+   * Checks if the promotion entity can be applied.
```
bojanz: we usually use "whether" instead of "if", right?

modules/promotion/src/Plugin/Commerce/PromotionCondition/OrderTotalPriceComparison.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionCondition;
+
+use Drupal\commerce_price\Price;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides a 'Order: Total amount aomparison' condition.
```
bojanz: s/a/an
s/aomparison/comparison

modules/promotion/src/Plugin/Commerce/PromotionCondition/PromotionConditionInterface.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Drupal\commerce_promotion\Plugin\Commerce\PromotionCondition;
+
+use Drupal\Core\Condition\ConditionInterface;
+
+/**
+ * Defines an interface for Condition plugins.
+ *
+ * Extends the core Condition interface to support Commerce Promotion specific
```
bojanz: Probably don't need to say "Commerce"? It's implied. The whole sentence feels redundant.

modules/promotion/src/PromotionConditionManager.php
```diff
@@ -0,0 +1,90 @@
...
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Executable\ExecutableManagerInterface;
+use Drupal\Core\Executable\ExecutableInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Plugin\CategorizingPluginManagerTrait;
+use Drupal\Core\Plugin\Context\ContextAwarePluginManagerTrait;
+use Drupal\Core\Plugin\DefaultPluginManager;
+
+/**
+ * Provides the Condition plugin manager.
```
bojanz: Bad docblock.

```
/**
 * Manages discovery and instantiation of promotion condition plugins.
 *
 * @see \Drupal\commerce_promotion\Annotation\CommercePromotionCondition
 * @see plugin_api
 */
```

modules/promotion/src/Annotation/PromotionCondition.php
```diff
@@ -0,0 +1,25 @@
...
+use Drupal\Core\Condition\Annotation\Condition;
+
+/**
+ * Defines the promotion condition plugin annotation object.
+ *
+ * Plugin namespace: Plugin\Commerce\PromotionCondition.
+ *
+ * @Annotation
+ */
+class PromotionCondition extends Condition {
```
bojanz: Commerce annotations start with Commerce. So this should be CommercePromotionCondition. If the same problem exists with offers we can merge and open a followup.

modules/promotion/tests/src/Kernel/PromotionConditionTest.php
```diff
@@ -0,0 +1,159 @@
...
+use Drupal\commerce_store\StoreCreationTrait;
+use Drupal\KernelTests\KernelTestBase;
+
+/**
+ * Tests promotion conditions.
+ *
+ * @group commerce
+ */
+class PromotionConditionTest extends KernelTestBase {
+  use StoreCreationTrait;
```
bojanz, line 20: Missing newline before use.

modules/promotion/tests/src/Kernel/PromotionConditionTest.php
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * The default store.
+   *
+   * @var \Drupal\commerce_store\Entity\StoreInterface
+   */
+  protected $store;
+
+  /**
+   * The offer manager.
```
bojanz: Incorrect docblock (it's not an offer manager)

modules/promotion/tests/src/Kernel/PromotionConditionTest.php
```diff
@@ -0,0 +1,159 @@
...
+      'order_number' => '6',
+      'store_id' => $this->store,
+      'line_items' => [],
+    ]);
+  }
+
+  /**
+   * Tests the order amount comparison condition.
+   */
+  public function testOrderAmountComparison() {
```
bojanz: Still says Amount instead of Total (wasn't updated for the rename)

modules/promotion/src/PromotionConditionManager.php
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * Constructs a ConditionManager object.
```
bojanz: Constructs a new PromotionConditionManager object.

PR #477, opened 2016-08-31
Issue #2793205 by mglaman: Price element has #maxlength, causes error…

mglaman:
… in Unicode::strlen

### DIFF
--- a/modules/price/src/Element/Price.php
+++ b/modules/price/src/Element/Price.php
@@ -5,7 +5,7 @@ namespace Drupal\commerce_price\Element;
 use CommerceGuys\Intl\Formatter\NumberFormatterInterface;
 use Drupal\commerce_price\Price as PriceValue;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\Render\Element\Textfield;
+use Drupal\Core\Render\Element\FormElement;
 
 /**
  * Provides a price form element.
@@ -27,7 +27,7 @@ use Drupal\Core\Render\Element\Textfield;
  *
  * @FormElement("commerce_price")
  */
-class Price extends Textfield {
+class Price extends FormElement {
 
   /**
    * {@inheritdoc}
@@ -134,6 +134,9 @@ class Price extends Textfield {
       // separator to use. This is the same pattern Drupal core uses.
       '#placeholder' => $number_formatter->format('9.99'),
     ];
+    unset($element['#size']);
+    unset($element['#maxlength']);
+
     if (count($currency_codes) == 1) {
       $last_visible_element = 'amount';
       $currency_code = reset($currency_codes);


mglaman:
@bojanz when #maxlength is set it gets passed to Unicode::strlen, but the value of the element is an array. so ['amount' => '9.99', 'currency_code' => 'USD'] when can't be `strlen`'d

See https://github.com/drupalcommerce/commerce/pull/478 which exposes this fact.

mglaman:
Lets just put it here instead of on element definition

`'#maxlength' => $element['#maxlength'],`

modules/price/src/Element/Price.php
```diff
@@ -27,16 +27,16 @@
  *
  * @FormElement("commerce_price")
  */
-class Price extends Textfield {
+class Price extends FormElement {
```
mglaman, line 30: @bojanz I looked at `\Drupal\Core\Render\Element\Textfield` and we weren't gaining anything out of extending it.  We weren't using any of the parent class's process or pre-render methods.

modules/price/src/Element/Price.php
```diff
@@ -128,8 +128,8 @@ public static function processElement(array $element, FormStateInterface $form_s
       '#title' => $element['#title'],
       '#default_value' => $amount,
       '#required' => $element['#required'],
-      '#size' => $element['#size'],
-      '#maxlength' => $element['#maxlength'],
+      '#size' => $element['#amount_size'],
+      '#maxlength' => $element['#amount_maxlength'],
```
mglaman: Allows us to control size, maxlength in main element and not hardcode. But also not trigger the `Unicode::strlen()`

modules/price/src/Element/Price.php
bojanz: Sure.

modules/price/src/Element/Price.php
```diff
@@ -27,16 +27,16 @@
...
 
   /**
    * {@inheritdoc}
    */
   public function getInfo() {
     $class = get_class($this);
     return [
-      '#size' => 10,
-      '#maxlength' => 128,
+      '#amount_size' => 10,
```
bojanz: Can't we not still use #size and #maxlength and unset it after transfering it to the subelement? I believe that's what other complex form elements do?

mglaman: I didn't see any other patterns like that, here's the check in question `\Drupal\Core\Form\FormValidator::performRequiredValidation`.

I'll just unset after we pass to amount

modules/price/src/Element/Price.php
```diff
@@ -134,6 +134,9 @@ public static function processElement(array $element, FormStateInterface $form_s
       // separator to use. This is the same pattern Drupal core uses.
       '#placeholder' => $number_formatter->format('9.99'),
     ];
+    unset($element['#size']);
```
mglaman, line 137: @bojanz done

PR #478, opened 2016-08-31
Issue #2793211 by mglaman: Price element valueCallback needs to ensur…

mglaman:
…e amount and currency code are not empty

### DIFF
--- a/modules/price/src/Element/Price.php
+++ b/modules/price/src/Element/Price.php
@@ -61,7 +61,8 @@ class Price extends FormElement {
    * {@inheritdoc}
    */
   public static function valueCallback(&$element, $input, FormStateInterface $form_state) {
-    if ($input !== FALSE && $input !== NULL) {
+    // Ensure we have all possible values before creating price object.
+    if ($input !== FALSE && $input !== NULL && isset($input['amount']) && isset($input['currency_code'])) {
       // Convert empty string value to numeric value.
       if ($input['amount'] === '') {
         $input['amount'] = '0';


mglaman:
Egh, @bojanz this is failing because of https://github.com/drupalcommerce/commerce/pull/477

EDIT: The lack of ^ being merged.

modules/price/src/Element/Price.php
```diff
@@ -61,7 +61,7 @@ public function getInfo() {
    * {@inheritdoc}
    */
   public static function valueCallback(&$element, $input, FormStateInterface $form_state) {
-    if ($input !== FALSE && $input !== NULL) {
+    if ($input !== FALSE && $input !== NULL && !empty($input['amount']) && !empty($input['currency_code'])) {
       // Convert empty string value to numeric value.
       if ($input['amount'] === '') {
```
bojanz, line 67: This conversion no longer makes sense if we pass on empty amounts? And won't this ignore a 0 EUR price, which is valid?

modules/price/src/Element/Price.php
```diff
@@ -61,7 +61,8 @@ public function getInfo() {
    * {@inheritdoc}
    */
   public static function valueCallback(&$element, $input, FormStateInterface $form_state) {
-    if ($input !== FALSE && $input !== NULL) {
+    // Ensure we have all possible values before creating price object.
+    if ($input !== FALSE && $input !== NULL && isset($input['amount']) && !empty($input['currency_code'])) {
```
mglaman: @bojanz changed it to an `isset` for amount, kept `!empty` for currency code. This should be good, yeah?

PR #480, opened 2016-09-03
Issue #2755529 by ndf: Product variants bulk creation

drugan:
See more on the issue page: [https://www.drupal.org/node/2755529](https://www.drupal.org/node/2755529)

### DIFF
--- a/modules/product/src/Entity/ProductVariation.php
+++ b/modules/product/src/Entity/ProductVariation.php
@@ -11,6 +11,8 @@ use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Url;
 use Drupal\user\UserInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Component\Utility\Html;
 
 /**
  * Defines the product variation entity class.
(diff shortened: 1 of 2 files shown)


modules/product/src/Entity/ProductVariation.php
```diff
@@ -500,4 +502,227 @@ public static function getCurrentUserId() {
...
+   * @param array $form
+   *
+   *   An array form for commerce_product.
+   *
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *
+   *   The form state of the commerce_product form with at least one variation
+   *   created.
+   */
+  public static function createAllVariations(array $form, FormStateInterface $form_state) {
```
mglaman, line 531: Perhaps bulk variation creation should go into its own service? So it could be run without just IEF?

drugan: Answered here: [https://www.drupal.org/node/2755529#comment-11577389](https://www.drupal.org/node/2755529#comment-11577389)

drugan: Travis didn't accepted the build because of 4 errors:
1. @see tag line:521 need to be moved under @param tags.
2. $param tags line:564 and line:567 need to be swapped.
3. The same on line:612 and line:615.
4. Line:683 need two additional spaces indented.

Sorry about that.

PR #481, opened 2016-09-04
Issue #2794909 by drugan: Change HTML step property of the quantity f…

drugan:
See more on the issue page: [https://www.drupal.org/node/2794909](https://www.drupal.org/node/2794909)


bojanz:
Continued in #507

PR #483, opened 2016-09-06
Changed adjustment amount to required only if there is selected any adjustment.

facine:


joshmiller83:
Without patch:

image

With patch (that applies cleanly):

image

😎

joshmiller83:
Copied over to the following Drupal issue ...

https://www.drupal.org/node/2798757

@facine ... perhaps you could comment there so @bojanz can easily tag you for the drupal.org contribution.

mglaman:
@joshmiller83 they make it easier to assign credit now w/ autocomplete

joshmiller83:
Unfortunately, though, the "states" doesn't seem to actually check for required fields. With patch, a price adjustment with label and no number yields the following fatal PHP error:

The provided value "" is not a numeric value." at /modules/contrib/commerce/modules/price/src/Calculator.php line 196, referer: /admin/commerce/orders/2/edit?destination=/admin/commerce/orders/wishlists

mglaman:
@joshmiller83 so that's a different bug and might be because branch is behind. That shold turn into 0.00

bojanz:
Merging for now, but it's obvious we need to rewrite this widget to remove #states usage (replace it with #ajax)

PR #484, opened 2016-09-07
Issue #2796005 by mglaman: Chain resolvers do not support tagged serv…

mglaman:
…ice priority

### DIFF
--- a/modules/order/commerce_order.services.yml
+++ b/modules/order/commerce_order.services.yml
@@ -8,7 +8,7 @@ services:
     class: Drupal\commerce_order\Resolver\DefaultOrderTypeResolver
     arguments: ['@entity_type.manager']
     tags:
-      - { name: commerce_order.order_type_resolver, priority: 10 }
+      - { name: commerce_order.order_type_resolver, priority: -100 }
 
   commerce_order.order_refresh:
     class: Drupal\commerce_order\OrderRefresh
--- a/modules/order/src/Resolver/ChainOrderTypeResolver.php
+++ b/modules/order/src/Resolver/ChainOrderTypeResolver.php
@@ -29,8 +29,8 @@ class ChainOrderTypeResolver implements ChainOrderTypeResolverInterface {
   /**
    * {@inheritdoc}
    */
-  public function addResolver(OrderTypeResolverInterface $resolver) {
-    $this->resolvers[] = $resolver;
+  public function addResolver(OrderTypeResolverInterface $resolver, $priority) {
+    $this->resolvers[$priority] = $resolver;
   }
 
   /**
--- a/tests/src/Unit/Resolver/ChainCountryResolverTest.php
+++ b/tests/src/Unit/Resolver/ChainCountryResolverTest.php
@@ -3,6 +3,7 @@
 namespace Drupal\Tests\commerce\Unit\Resolver;
 
 use Drupal\commerce\Resolver\ChainCountryResolver;
+use Drupal\Core\DependencyInjection\ContainerBuilder;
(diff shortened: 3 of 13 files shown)


bojanz:
Defaulting priority to 0 means that any call to addResolver() that doesn't have a priority specified will overwrite the first resolver in the list.

mglaman:
That's what I get for copying core.

mglaman:
@bojanz updated, removed default parameter value for $priority.

modules/order/src/Resolver/ChainOrderTypeResolver.php
```diff
@@ -29,8 +29,8 @@ public function __construct(array $resolvers = []) {
   /**
    * {@inheritdoc}
    */
-  public function addResolver(OrderTypeResolverInterface $resolver) {
-    $this->resolvers[] = $resolver;
+  public function addResolver(OrderTypeResolverInterface $resolver, $priority) {
+    $this->resolvers[$priority] = $resolver;
```
mglaman, line 33: This needs to ksort after add, otherwise priority doesn't mean anything.

mglaman:
@bojanz okay, `$priority = 0` removed. Also added `ksort($this-resolvers)` after adding one, ensures priority is respected. Found that as last minute bug when implementing this patch on project.

modules/order/commerce_order.services.yml
```diff
@@ -8,4 +8,4 @@ services:
     class: Drupal\commerce_order\Resolver\DefaultOrderTypeResolver
     arguments: ['@entity_type.manager']
     tags:
-      - { name: commerce_order.order_type_resolver, priority: 10 }
+      - { name: commerce_order.order_type_resolver, priority: -100 }
```
mglaman, line 11: Followed the pattern in https://github.com/drupalcommerce/commerce/pull/366 now that I know how priority works O:)

tests/src/Unit/Resolver/ChainCountryResolverTest.php
```diff
@@ -48,9 +48,9 @@ public function testResolver() {
     $third_resolver->expects($this->never())
       ->method('resolve');
 
-    $this->chainCountryResolver->addResolver($first_resolver);
-    $this->chainCountryResolver->addResolver($second_resolver);
-    $this->chainCountryResolver->addResolver($third_resolver);
+    $this->chainCountryResolver->addResolver($first_resolver, 0);
```
mglaman: The priority should be done as `-100`, `100`, and `200`. We should rebuild the container and properly test the priority sorting done when the container is rebuilt.

tests/src/Unit/Resolver/ChainLocaleResolverTest.php
```diff
@@ -48,9 +48,9 @@ public function testResolver() {
     $third_resolver->expects($this->never())
       ->method('resolve');
 
-    $this->chainLocaleResolver->addResolver($first_resolver);
-    $this->chainLocaleResolver->addResolver($second_resolver);
-    $this->chainLocaleResolver->addResolver($third_resolver);
+    $this->chainLocaleResolver->addResolver($first_resolver, 0);
```
mglaman: The priority should be done as `-100`, `100`, and `200`. We should rebuild the container and properly test the priority sorting done when the container is rebuilt.

mglaman (commented):
Need to update tests to properly reflect priority values and how the container is built.

mglaman:
@bojanz want to review this? I've updated the tests to also test priority. I didn't want to mock the container 100%, so I borrowed how it sorts tagged services and mocked that.

PR #485, opened 2016-09-08
Remove usage of composer_manager in TravisCI

mglaman:
Add repository as local composer path. Use `composer require` to add
self.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,11 +6,11 @@
     "license": "GPL-2.0+",
     "require": {
         "drupal/core": "~8.1",
-        "drupal/address": "8.1.x-dev",
-        "drupal/entity": "8.1.x-dev",
-        "drupal/inline_entity_form": "8.1.x-dev",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -6,11 +6,11 @@
     "license": "GPL-2.0+",
     "require": {
         "drupal/core": "~8.1",
-        "drupal/address": "8.1.x-dev",
-        "drupal/entity": "8.1.x-dev",
-        "drupal/inline_entity_form": "8.1.x-dev",
-        "drupal/profile": "8.1.x-dev",
-        "drupal/state_machine": "8.1.x-dev",
+        "drupal/address": "1.x-dev",
```
mglaman, line 9: @bojanz so `composer require` was borking over this. Which I have experienced myself, too. Changing made tests pass with the new `composer require drupal/commerce`

bojanz: D'oh, we forgot to update those. Good catch

PR #488, opened 2016-09-13
Fix value used in createPaymentMethod must be array

googletorp:
$values is $values['payment_details'] with latest stable core. The effect is a notice and sending NULL to createPaymentMethod which can lead to fatal depending on PHP config/version since it must be an array.


bojanz:
Needs more investigation. The original line is correct, $form['payment_details'] exists, so the bug might be elsewhere (bad #parents?)

bojanz:
Many things have changed since september, including today's complete rewrite of the PaymentInformation pane. It's probably safe to close this. Reopen if you have additional info on where the bug could be coming from.

PR #489, opened 2016-09-13
Order payment method should be selectable and default

googletorp:
Right now, if a user has saved a payment method on an order (not reusable) and somehow comes back to the payment information pane (maybe he decided to buy some more stuff or something), he will be prompted to create a new payment method (or use an existing reusable).

The payment method actually used on an order should take precedence over existing reusable ones and always be default.


steveoliver:
Good idea. Needs re-roll.

mglaman:
We should also expand the tests to show that going back has the selected payment option.

bojanz:
Added tests, clarified naming, expanded fix. Committed in 9f451cddd49e56ac06aca433d0fbca0786a36af8.

PR #490, opened 2016-09-14
Issue #2798697 by plopesc: Payment Gateway & Checkout Flow form throws exception

plopesc:
Conflict with Wokbench moderation.
See https://www.drupal.org/node/2798697


mglaman (approved):
+1 meets what was discussed, I like the usage of `CommercePluginEntityFormBase`. We can re-use this for contrib which follow our patterns

PR #491, opened 2016-09-15
#2800223 Create an order refresh process

mglaman:
- [x] needsRefresh($order) would check the last refresh date ($order->changed?) and use the refresh settings to make a decision. Can always be TRUE for the initial implementation.

refresh($order) does the following:
- [x] Update line item titles from the purchased entity titles (and test.)
- [x] Resolve the unit prices (using the price resolver from PR #366) (and test.)
- [X] Clear all adjustments
- [X] Run the order processors.
- [X] Availability processor
- [x] Move and test order refresh settings
- [X] Save the order.


modules/order/commerce_order.services.yml
```diff
@@ -9,3 +9,15 @@ services:
     arguments: ['@entity_type.manager']
     tags:
       - { name: commerce_order.order_type_resolver, priority: 10 }
+
+  commerce_order.order_refresh:
+    class: Drupal\commerce_order\OrderRefresh
+    arguments: ['@entity_type.manager']
+    tags:
+      - { name: service_collector, call: addProcessor, tag: commerce_order.order_refresh_processor }
```
bojanz, line 17: Core tends to put call after tag. o:)

modules/order/src/AvailabilityOrderRefreshProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Drupal\commerce_order;
+
+use Drupal\commerce\AvailabilityManagerInterface;
+use Drupal\commerce_order\Entity\OrderInterface;
+
+/**
+ * Provides an order refresh processor for availability.
+ */
+class AvailabilityOrderRefreshProcessor implements OrderRefreshProcessorInterface {
```
bojanz: AvailabilityOrderProcessor, OrderProcessorInterface. The refresh part is redundant.

modules/order/src/AvailabilityOrderRefreshProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+    $this->availabilityManager = $availability_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function refresh(OrderInterface $order) {
+    foreach ($order->getLineItems() as $line_item) {
+      $available = $this->availabilityManager->check($line_item->getPurchasedEntity(), $line_item->getQuantity());
+
```
bojanz: Do we want to set a message when the removal happens? I think 1.x did?

modules/order/src/OrderRefresh.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+
+/**
+ * Default implementation for order refresh.
```
bojanz, line 13: I tend to leave out docblocks that don't add anything of value nowadays (like the "default implementation" one). Keep it if you disagree.

modules/order/src/OrderRefresh.php
```diff
@@ -0,0 +1,79 @@
...
+   * The order refresh processors.
+   *
+   * @var \Drupal\commerce_order\OrderRefreshProcessorInterface[]
+   */
+  protected $processors;
+
+  protected static $refreshedOrders = [];
+
+  /**
+   * Creates a new OrderRefresh object.
```
bojanz: Constructs a new. Verb matches method name.

modules/order/src/OrderRefresh.php
```diff
@@ -0,0 +1,79 @@
...
+    $order->setAdjustments([]);
+    foreach ($order->getLineItems() as $line_item) {
+      $line_item->setAdjustments([]);
+      $line_item->setTitle($line_item->getPurchasedEntity()->getLineItemTitle());
+      // @todo re-resolve unit price.
+      $line_item->save();
+    }
+
+    foreach ($this->processors as $processor) {
+      $processor->refresh($order);
```
bojanz: $processor->process($order) would make more sense IMO.

modules/order/src/OrderRefreshInterface.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+
+/**
+ * Service that runs refresh processes on an order.
```
bojanz: Everything's a service, no point in mentioning that. "Refreshes draft orders."? Should decide where we want to do the description of the full process, in the class docblock or the method docblock.

modules/order/src/OrderStorage.php
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * The order refresh process.
+   *
+   * @var \Drupal\commerce_order\OrderRefreshInterface
+   */
+  protected $orderRefresh;
+
+  /**
+   * Constructs a new CommerceContentEntityStorage object.
```
bojanz: Wrong class name in the docblock.

modules/order/src/OrderStorage.php
```diff
@@ -0,0 +1,77 @@
...
+      $container->get('event_dispatcher'),
+      $container->get('commerce_order.order_refresh')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function postLoad(array &$entities) {
+    foreach ($entities as $entity) {
```
bojanz, line 68: Wondering if it would have made more sense to do this in hook_commerce_order_load(). Each overriden storage is something we'll need to undo at some point, to allow the workflow modules to do their own swapping.

bojanz:
We also agreed to move the cart refresh settings into the order module, since we want the order refresh to run for any draft order, not just carts.

bojanz (changes requested):

modules/order/src/OrderStorage.php
mglaman: Feels ugly to have it in a typical hook. Would it be odd to just put in the entity's `::postLoad` method that the storage is expected to invoke?

modules/order/commerce_order.services.yml
mglaman: I copied how our other services.yml are, I'll update but we should update the others?

modules/order/src/AvailabilityOrderRefreshProcessor.php
mglaman: What would the message be? Also I don't know where in 1.x we'd have something like this. 

In http://cgit.drupalcode.org/commerce/tree/modules/cart/commerce_cart.module#n1114 we removed line items with `null` price values, but this was silent as well.

modules/order/src/OrderRefresh.php
mglaman: I only left it to match the Drupal coding standards.

Edit: otherwise I agree. It's useless if we have interface to document it.

modules/order/src/OrderRefresh.php
```diff
@@ -0,0 +1,117 @@
...
+  public function refresh(OrderInterface $order) {
+    // Do not attempt to refresh an order that is already refreshing.
+    if (!isset(self::$refreshingOrders[$order->id()])) {
+      self::$refreshingOrders[$order->id()] = TRUE;
+
+      $order->setAdjustments([]);
+      foreach ($order->getLineItems() as $line_item) {
+        $line_item->setAdjustments([]);
+        $line_item->setTitle($line_item->getPurchasedEntity()->getLineItemTitle());
+        $line_item->setUnitPrice($this->chainPriceResolver->resolve($line_item->getPurchasedEntity(), $line_item->getQuantity()));
```
bojanz: It's not readable for the resolve() call to be nested like this. One statement per line.

bojanz: It's not readable for the resolve() call to be nested like this. One statement per line.

mglaman:
locally all pass but CartRefreshTest because I didn't ove it. but not for Travis

modules/order/src/AvailabilityOrderRefreshProcessor.php
bojanz: Looks like you're right, I guess I imagined the message.

modules/cart/src/Form/AddToCartForm.php
```diff
@@ -179,7 +179,8 @@ public function buildEntity(array $form, FormStateInterface $form_state) {
     $entity = parent::buildEntity($form, $form_state);
     // Now that the purchased entity is set, populate the title and price.
     $entity->setTitle($entity->getPurchasedEntity()->getLineItemTitle());
-    $entity->setUnitPrice($this->chainPriceResolver->resolve($entity->getPurchasedEntity()));
+    $price = $this->chainPriceResolver->resolve($entity->getPurchasedEntity());
```
bojanz: Is resolving the price at this point necessary? I thought this was the equivalent of creating a line item, in which case we'd be able to simply do $entity->setUnitPrice($entity->getPurchasedEntity()->getPrice());

modules/order/src/AvailabilityOrderProcessor.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+use Drupal\commerce\AvailabilityManagerInterface;
+use Drupal\commerce_order\Entity\OrderInterface;
+
+/**
+ * Provides an order refresh processor for availability.
```
bojanz: "Provides an order processor that removes entities that are no longer available."?

modules/order/src/AvailabilityOrderProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * The availability manager.
+   *
+   * @var \Drupal\commerce\AvailabilityManagerInterface
+   */
+  protected $availabilityManager;
+
+  /**
+   * Creates a new AvailabilityOrderRefreshProcessor object.
```
bojanz: Constructs a new AvailabilityOrderProcessor object.

modules/order/src/Entity/OrderTypeInterface.php
```diff
@@ -29,4 +32,38 @@ public function getWorkflowId();
...
+   * @return string
+   *   The refresh mode.
+   */
+  public function getRefreshMode();
+
+  /**
+   * Sets the refresh mode for the order type.
+   *
+   * @param string $refresh_mode
+   *   The refresh mode.
```
bojanz: Missing @return $this

modules/order/src/Entity/OrderTypeInterface.php
```diff
@@ -29,4 +32,38 @@ public function getWorkflowId();
...
+   * @return int
+   *   The frequency, in seconds.
+   */
+  public function getRefreshFrequency();
+
+  /**
+   * Sets the refresh frequency for the order type.
+   *
+   * @param int $refresh_frequency
+   *   The frequency, in seconds.
```
bojanz: Missing @return $this

modules/order/src/OrderProcessorInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+
+/**
+ * Defines an interface for processing an order.
```
bojanz: Maybe something like:
"Defines an interface for order processors.

Order processors modify/handle/rebuild orders during the refresh process."

modules/order/tests/src/Kernel/OrderRefreshTest.php
```diff
@@ -0,0 +1,267 @@
...
+    $line_item = $this->reloadEntity($line_item);
+
+    $this->assertEquals($this->variation2->getPrice(), $line_item->getUnitPrice());
+  }
+
+  /**
+   * Tests the order refresh process, with the availability processor.
+   */
+  public function testAvailabilityOrderRefreshProcessor() {
+
```
bojanz: Unneeded newline at the beginning of the method.

modules/order/tests/src/Kernel/OrderRefreshTest.php
```diff
@@ -0,0 +1,267 @@
...
+    $another_line_item = $this->lineItemStorage->createFromPurchasableEntity($this->variation2, [
+      'unit_price' => new Price('3.00', 'USD'),
+      'quantity' => '2',
+
+    ]);
+    $another_line_item->save();
+    $another_line_item = $this->reloadEntity($another_line_item);
+
+    $this->order->setLineItems([$line_item, $another_line_item])->save();
+
```
bojanz: Unneeded newline between the method call and the result checking.

modules/order/tests/src/Kernel/OrderRefreshTest.php
```diff
@@ -0,0 +1,267 @@
...
+    $this->assertEquals(2, count($this->order->getLineItems()));
+
+    $this->container->get('commerce_order.order_refresh')->refresh($this->order);
+
```
bojanz: Unneeded newline between the method call and the result checking.

modules/order/tests/src/Kernel/OrderRefreshTest.php
```diff
@@ -0,0 +1,267 @@
...
+   */
+  public function testUnitPriceRefresh() {
+    /** @var \Drupal\commerce_order\Entity\LineItemInterface $line_item */
+    $line_item = $this->lineItemStorage->createFromPurchasableEntity($this->variation2);
+    $line_item->save();
+    $line_item = $this->reloadEntity($line_item);
+
+    $this->order->addLineItem($line_item);
+    $this->order->save();
+    $this->container->get('commerce_order.order_refresh')->refresh($this->order);
```
bojanz: Why do we use $this->container->get('commerce_order.order_refresh') all over the place in an OrderRefreshTest? Set a $this->orderRefresh at setUp().

bojanz (changes requested):

modules/order/src/OrderStorage.php
bojanz: That sounds even worse TBH.

mglaman: Yeah, that's why I didn't put it there

modules/cart/src/Form/AddToCartForm.php
mglaman: Wait, it might be. What if people are reacting to the Add to Cart event and want the resolved price?

modules/order/src/OrderRefresh.php
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function needsRefresh(OrderInterface $order) {
+    /** @var \Drupal\commerce_order\Entity\OrderTypeInterface $order_type */
+    $order_type = $this->orderTypeStorage->load($order->bundle());
+
+    // Check the order's changed time against the refresh frequency.
+    // We use time() since the REQUEST constant can become stale. This is
```
bojanz: You mean REQUEST_TIME instead of REQUEST, right?

bojanz: You mean REQUEST_TIME instead of REQUEST, right?

modules/order/src/OrderRefresh.php
bojanz: This still hasn't been addressed.

modules/order/tests/src/Functional/OrderRefreshSettingsTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Drupal\Tests\commerce_order\Functional;
+
+use Drupal\Tests\commerce\Functional\CommerceBrowserTestBase;
+use Drupal\commerce_order\Entity\OrderType;
+
+/**
+ * Tests the cart refresh.
```
bojanz: Tests the order refresh settings.

(Why isn't this test a part of the OrderTypeTest?)

mglaman: Because direct refactor of directory and namespace from cart module. I'm going to just merge them

modules/order/src/OrderRefresh.php
```diff
@@ -0,0 +1,119 @@
...
+    // Do not attempt to refresh an order that is already refreshing.
+    if (!isset(self::$refreshingOrders[$order->id()])) {
+      self::$refreshingOrders[$order->id()] = TRUE;
+
+      $order->setAdjustments([]);
+      foreach ($order->getLineItems() as $line_item) {
+        $line_item->setAdjustments([]);
+        $line_item->setTitle($line_item->getPurchasedEntity()->getLineItemTitle());
+        $unit_price = $this->chainPriceResolver
+          ->resolve($line_item->getPurchasedEntity(), $line_item->getQuantity());
```
bojanz: We should not be breaking the ->resolve call into its own line.

PR #493, opened 2016-09-15
Issue #2799375 by mglaman: AdjustmentItemList::removeAdjustment isn't…

mglaman:
… removing adjustment


bojanz:
Do we have confirmation that this doesn't result in false positives due to the loose comparison?

mglaman:
@bojanz so here's the deal. It might be my implementation. The `===` only works _if it is the exact same object_.

For example, the tests fail with `===` and creating a match object to try and remove.

```
    // Create new adjustment object, matching existing. Ensure it gets removed.
    $adjustment = new Adjustment([
      'type' => 'custom',
      'label' => '10% off',
      'amount' => new Price('-1.00', 'USD'),
    ]);
    $order->removeAdjustment($adjustment);
```

Whereas it passed in test when we have `$adjustments[0]`.

So I guess if we want to decide: you must retrieve the object, compare it, and choose to remove it. Or be able to reconstruct the adjustment and say "remove something like this".

PR #495, opened 2016-09-19
Add cache tags to PriceDefaultFormatter.

facine:
We need to expire the cache if change the formatter settings.

### DIFF
--- a/modules/price/src/Plugin/Field/FieldFormatter/PriceDefaultFormatter.php
+++ b/modules/price/src/Plugin/Field/FieldFormatter/PriceDefaultFormatter.php
@@ -132,6 +132,10 @@ class PriceDefaultFormatter extends FormatterBase implements ContainerFactoryPlu
     }
     $currencies = $this->currencyStorage->loadMultiple($currency_codes);
 
+    $cache_tags = [
+      'config:' . $this->fieldDefinition->getTargetEntityTypeId() . '.' . $this->fieldDefinition->getTargetBundle() . '.' . $this->viewMode,
+    ];
+
     $elements = [];
     foreach ($items as $delta => $item) {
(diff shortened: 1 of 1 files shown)


modules/price/src/Plugin/Field/FieldFormatter/PriceDefaultFormatter.php
```diff
@@ -142,6 +142,9 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
             'languages:' . LanguageInterface::TYPE_INTERFACE,
             'country',
           ],
+          'tags' => [
+            'config:' . $items[$delta]->getFieldDefinition()->getTargetEntityTypeId() . '.' . $items[$delta]->getFieldDefinition()->getTargetBundle() . '.' . $this->viewMode,
```
bojanz: $this->fieldDefinition should be the same as $items[$delta]->getFieldDefinition(), right? That one is shorter.
Looks like this tag doesn't change between deltas, which means we can calculate it above the loop (define a $cache_tag above $elements = [];)

facine:
You're in the right! Fixed!

Thanks!

swentel:
When changing entity displays, everything with 'rendered' is cleared anyway, so you should be fine. Most default formatters in core don't have any special cache tags, except then for files/image formatters maybe. So unless you're doing something funky, I think you're fine with the current formatters.

(And even if we'd need a cache tag on the field, then we should rather add this to core)

facine:
By anyway when I try this locally don't work, now works fine without this patch...

Sorry!

Berdir:
Yeah. this shouldn't be needed. the cache tag also doesn't seem valid, it is missing the prefix  "core.entity_view_display.", this doesn't actually do something right now.

PR #501, opened 2016-09-20
Add commerce prefix to all commerce related permissions.

facine:
I think to avoid possible problems with other modules, we should add the prefix: commerce.


bojanz:
No point in doing this, all permissions are going to be removed once https://www.drupal.org/node/2801031 happens.

facine:
Close this PR!

PR #502, opened 2016-09-20
Issue #2802097 by mglaman, mitrpaka: Checkout as a guest - cart not e…

mglaman:
…mptied (and no new cart created) after order completetion

### DIFF
--- a/modules/cart/commerce_cart.module
+++ b/modules/cart/commerce_cart.module
@@ -9,6 +9,7 @@ use Drupal\commerce_order\Entity\OrderInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\state_machine\Plugin\Workflow\WorkflowState;
 
 /**
  * Implements hook_theme().
@@ -99,6 +100,20 @@ function commerce_cart_user_login($account) {
 function commerce_cart_commerce_order_delete(OrderInterface $order) {
   if (!empty($order->cart)) {
     \Drupal::service('commerce_cart.cart_session')->deleteCartId($order->id());
+  }
+}
+
+/**
+ * Implements hook_ENTITY_TYPE_update().
+ */
+function commerce_cart_commerce_order_presave(OrderInterface $order) {
+  if (!$order->isNew() && $order->cart->value == TRUE) {
+    $place_transition = $order->getState()->getWorkflow()->getTransition('place');
+    $from_states = $place_transition->getFromStates();
+    $to_state = $place_transition->getToState()->getId();
+    if (in_array($order->original->state->value, array_keys($from_states)) && $order->state->value == $to_state) {
+      \Drupal::service('commerce_cart.cart_provider')->finalizeCart($order);
+    }
   }
 }
 
--- a/modules/cart/src/Plugin/Block/CartBlock.php
+++ b/modules/cart/src/Plugin/Block/CartBlock.php
@@ -4,6 +4,7 @@ namespace Drupal\commerce_cart\Plugin\Block;
 
 use Drupal\commerce_cart\CartProviderInterface;
 use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
@@ -109,11 +110,17 @@ class CartBlock extends BlockBase implements ContainerFactoryPluginInterface {
    *   A render array.
    */
   public function build() {
+    $cachable_metadata = new CacheableMetadata();
+    $cachable_metadata->addCacheContexts(['user', 'session']);
+
     /** @var \Drupal\commerce_order\Entity\OrderInterface[] $carts */
     $carts = $this->cartProvider->getCarts();
     $carts = array_filter($carts, function ($cart) {
       /** @var \Drupal\commerce_order\Entity\OrderInterface $cart */
-      return $cart->hasLineItems();
+      // There is a chance the cart may have converted from a draft order, but
+      // is still in session. Such as just completing check out. So we verify
+      // that the cart is still a cart.
+      return $cart->hasLineItems() && $cart->cart->value;
     });
 
     $count = 0;
@@ -124,6 +131,7 @@ class CartBlock extends BlockBase implements ContainerFactoryPluginInterface {
         foreach ($cart->getLineItems() as $line_item) {
           $count += (int) $line_item->getQuantity();
         }
+        $cachable_metadata->addCacheableDependency($cart);
       }
     }
 
@@ -149,6 +157,11 @@ class CartBlock extends BlockBase implements ContainerFactoryPluginInterface {
       '#url' => Url::fromRoute('commerce_cart.page')->toString(),
       '#content' => $cart_views,
       '#links' => $links,
+      '#cache' => [
+        'contexts' => $cachable_metadata->getCacheContexts(),
+        'tags' => $cachable_metadata->getCacheTags(),
+        'max-age' => $cachable_metadata->getCacheMaxAge(),
+      ],
     ];
   }
 
--- a/modules/cart/tests/src/Kernel/CartOrderPlacedTest.php
+++ b/modules/cart/tests/src/Kernel/CartOrderPlacedTest.php
@@ -1,0 +1,157 @@
+<?php
+
+namespace Drupal\Tests\commerce_cart\Kernel;
+
+use Drupal\commerce_store\StoreCreationTrait;
+use Drupal\Component\Render\FormattableMarkup;
+use Drupal\KernelTests\Core\Entity\EntityKernelTestBase;
+
+/**
+ * Tests the unsetting of the cart flag when order is placed.
+ */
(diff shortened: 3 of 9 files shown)


modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -109,6 +110,9 @@ public function blockSubmit($form, FormStateInterface $form_state) {
    *   A render array.
    */
   public function build() {
+    $cachable_metadata = new CacheableMetadata();
+    $cachable_metadata->addCacheContexts(['user']);
```
bojanz: You also need session for anonymous carts?

modules/cart/tests/src/Kernel/CartOrderPlacedTest.php
```diff
@@ -0,0 +1,168 @@
...
+namespace Drupal\Tests\commerce_cart\Kernel;
+
+use Drupal\commerce_store\StoreCreationTrait;
+use Drupal\Component\Render\FormattableMarkup;
+use Drupal\KernelTests\Core\Entity\EntityKernelTestBase;
+
+/**
+ * Tests the unsetting of the cart flag when order is placed.
+ */
+class CartOrderPlacedTest extends EntityKernelTestBase {
```
bojanz: I would also expect this test to confirm that the next add to cart uses a fresh cart object?

modules/cart/commerce_cart.module
```diff
@@ -103,6 +104,20 @@ function commerce_cart_commerce_order_delete(OrderInterface $order) {
 }
 
 /**
+ * Implements hook_ENTITY_TYPE_update().
+ */
+function commerce_cart_commerce_order_presave(OrderInterface $order) {
```
bojanz: It makes sense for this to be done for both carts and admin orders, but didn't you catch this bug by testing checkout? There's this in CheckoutFlowBase which should have handled it:
      if ($next_step_id == 'complete') {
        // @todo Do this earlier in the process.
        $this->order->cart = FALSE;
        // Place the order.
        $transition = $this->order->getState()->getWorkflow()->getTransition('place');
        $this->order->getState()->applyTransition($transition);
        // Notify other modules.
        $dispatch_checkout_complete = TRUE;
      }

modules/cart/src/Plugin/Block/CartBlock.php
mglaman: True, adding that.

modules/cart/tests/src/Kernel/CartOrderPlacedTest.php
mglaman: Good call. Added

```
    // We should be able to create a new cart and not get an exception.
    $new_cart_order = $this->container->get('commerce_cart.cart_provider')->createCart('default', $this->store);
    $this->assertNotEquals($cart_order->id(), $new_cart_order->id());
```

Getting duplicate cart exception.

modules/cart/commerce_cart.module
mglaman: Hm. So maybe it was unsetting the cart value, but the block was stale on just the checkout complete page?

mglaman: Or our hidden issue is 

```
\Drupal::service('commerce_cart.cart_session')->deleteCartId($order->id());
```

mglaman: I've removed the code from checkout. It's now just done here.

modules/cart/tests/src/Kernel/CartOrderPlacedTest.php
mglaman: This is resolved. We test creating new cart, which uncovered all those other little bugs.

modules/cart/tests/src/Kernel/CartOrderPlacedTest.php
```diff
@@ -0,0 +1,165 @@
...
+   * @var \Drupal\commerce_cart\CartManagerInterface
+   */
+  protected $cartManager;
+
+  /**
+   * The attribute field manager.
+   *
+   * @var \Drupal\commerce_product\ProductAttributeFieldManagerInterface
+   */
+  protected $attributeFieldManager;
```
mglaman: This isn't used, copy paste cruft.

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
```diff
@@ -316,8 +316,6 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       ]);
 
       if ($next_step_id == 'complete') {
-        // @todo Do this earlier in the process.
-        $this->order->cart = FALSE;
```
mglaman: Removed this, since the `place` transition will fire `commerce_cart` to do `finalizeCart`

mglaman (commented):
Added some testing around creating carts and what happens when finishing checkout. Issue ballooned a lot of little bugs.

modules/cart/src/CartManager.php
```diff
@@ -89,6 +109,19 @@ public function emptyCart(OrderInterface $cart, $save_cart = TRUE) {
   /**
    * {@inheritdoc}
    */
+  public function finalizeCart(OrderInterface $cart, $save_cart = TRUE) {
+    $cart->cart = FALSE;
+    $this->cartSession->deleteCartId($cart->id());
```
mglaman: This prevents anonymous checkout from working. Maybe we move the session removal out of this method, and just keep it in the checkout form submit. What do you think @bojanz 

access_denied___drupal_and_google_hangouts_-_matt_commerceguys_com

modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -109,11 +110,18 @@ public function blockSubmit($form, FormStateInterface $form_state) {
...
     /** @var \Drupal\commerce_order\Entity\OrderInterface[] $carts */
     $carts = $this->cartProvider->getCarts();
     $carts = array_filter($carts, function ($cart) {
       /** @var \Drupal\commerce_order\Entity\OrderInterface $cart */
-      return $cart->hasLineItems();
+      // There is a chance the cart may have converted from a draft order, but
+      // is still in session. Such as just completing check out. So we verify
+      // that the cart is still a cart.
+      // @todo find better place to put this, such as provider::isFinalized
+      return $cart->hasLineItems() && $cart->cart->value;
```
mglaman, line 123: Right now when a checkout completes the order ID is still in session, which allows us to support anonymous access to an order. However, the order is no longer a cart but causes it to be returned by the cart provider.

If we adjust the provider's `::getCartIds` or `::getCarts` we'll the same issue we would have if we deleted the ID from the session - no idea if the anonymous user owned the cart.

Right now our main bug is the fact the cart block shows this anonymous order in the cart block on checkout complete (and probably any other page until the session expires.) This check prevents this by validating the returned cart is still a cart.

@bojanz should we commit this with the `@todo` or find fix. Trying to outweigh fix vs. lingering bug

modules/cart/commerce_cart.module
```diff
@@ -103,6 +104,20 @@ function commerce_cart_commerce_order_delete(OrderInterface $order) {
...
+  if (!$order->isNew() && $order->cart->value == TRUE) {
+    $place_transition = $order->getState()->getWorkflow()->getTransition('place');
+    $from_states = $place_transition->getFromStates();
+    $to_state = $place_transition->getToState()->getId();
+    if (in_array($order->original->state->value, array_keys($from_states)) && $order->state->value == $to_state) {
+      \Drupal::service('commerce_cart.cart_manager')->finalizeCart($order);
```
bojanz: We agreed to move the finalizeCart method from CartManager to CartProvider, that way we don't need to inject the CartProvider into the CartManager.

modules/cart/tests/src/Kernel/CartOrderPlacedTest.php
bojanz: Still there

modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -109,11 +110,18 @@ public function blockSubmit($form, FormStateInterface $form_state) {
...
+
     /** @var \Drupal\commerce_order\Entity\OrderInterface[] $carts */
     $carts = $this->cartProvider->getCarts();
     $carts = array_filter($carts, function ($cart) {
       /** @var \Drupal\commerce_order\Entity\OrderInterface $cart */
-      return $cart->hasLineItems();
+      // There is a chance the cart may have converted from a draft order, but
+      // is still in session. Such as just completing check out. So we verify
+      // that the cart is still a cart.
+      // @todo find better place to put this, such as provider::isFinalized
```
bojanz: This whole part of the diff (comment, cart->cart check) is no longer relevant and can be removed, right>

bojanz (changes requested):

modules/cart/src/Plugin/Block/CartBlock.php
mglaman: Well it's half true. We can't remove it from the session, but we have no real way of identifying the order in session is no longer a cart. It's an edge case on anonymous users and checkout and that we have one session variable key for all carts, instead of like 1.x which had two.

But changing it would take a lot of work and should possibly be just a follow-up, since this block is the one bug experience.

mglaman:
@bojanz struck again by this.

```
There was 1 error:
1) Drupal\Tests\commerce_order\Functional\OrderAdminTest::testEditOrder
Behat\Mink\Exception\ExpectationException: The field "adjustments[0][definition][label]" value is "", but "10% off" expected.
```

PR #505, opened 2016-09-23
Issue #2798301 - Define admin label to checkout panes

mitrpaka:


modules/checkout/src/Plugin/Commerce/CheckoutPane/CheckoutPaneBase.php
```diff
@@ -127,6 +127,18 @@ public function getLabel() {
   /**
    * {@inheritdoc}
    */
+  public function getAdminLabel() {
+    $admin_label = $this->pluginDefinition['admin_label'];
```
mglaman: Have the plugin manager set this. That if the `admin_label` is empty, prepopulate it to the value of `label`. That way all processing is done on discovery and cached.

mglaman (changes requested):

modules/checkout/src/Plugin/Commerce/CheckoutPane/CheckoutPaneBase.php
mitrpaka: Thanks for the review and guidance way forward. PR updated.

bojanz:
Tweaked and committed in 60bae651b76ff256b470702393354eb7517177b6.

PR #507, opened 2016-09-23
Issue #2794909 by drugan: Change step property of the quantity field

drugan:
Another example how we can change step and max properties for quantity field on the Add to Cart form.


mglaman:
@drugan should we just make a decimal/integer/number ~~formatter~~ widget which supports making `#step` configurable? I'm going to search core for that issue.

drugan:
@mglaman As I know there is no UI to config number field "#step" value but it should be. It has default default value set to 1 for all types of number field. See: **core/lib/Drupal/Core/Render/Element/Number.php**, line 42. For the 'decimal' number field type the "#step" is computed based on 'scale', i.e.- the number of digits after decimal point sign. See: **core/lib/Drupal/Core/Field/Plugin/Field/FieldWidget/NumberWidget.php**, line 81.

Actually, it is a good idea to make the quantity (and not only) field configurable for "#min", "#max" and "#step" properties.

Do you think it worth to open an issue on the core with a patch relative?

mglaman:
@drugan I say make the widget in Commerce, open follow up in Core w/ patch to put in core

drugan:
@mglaman ok, I'll try it.

drugan:
@mglaman Done. But I've decided to make a new [PR](https://github.com/drupalcommerce/commerce/pull/525) as it is a bit different solution.

The quick patch and explanation are[ there](https://www.drupal.org/node/2794909#comment-11700553).

mglaman:
Carried on in #525

PR #508, opened 2016-09-23
Issue #2804227 Add $order->getBalance()

steveoliver:
See https://www.drupal.org/node/2804227


mglaman (changes requested):
We need to update the order entity kernel test to verify these methods.

@bojanz is this what you were envisioning? (to get 👍 before writing tests.)

steveoliver:
Just rebased this.

bojanz:
Continued in #786.

PR #511, opened 2016-09-24
#2805625: Add a new service to manage the product variation rendering

plopesc:
**Problem/Motivation**

Currently, the product variation fields are added to the Product render array in commerce_product_commerce_product_view() function.

When someone needs to rewrite that logic, it's necessary to implement that hook, or the _alter() version to override that behavior. That also forces to redo that logic again, hitting performance.

That's why we think a good improvement would be to add a service where that task in performed. In this way, this logic can be replaced without need to redo all the processing when users need to rewrite the default behavior.

**Proposed resolution**

Add the concept of _ProductVariationRenderer_, as tagged service where implement the logic explained above
Create a _ProductVariationRendererManager_ service that takes care of managing the different Product Variation Renderers
Create a default service _ProductVariationRendererDefault_ that implements the code the currently is in commerce_product_commerce_product_view()

**API changes**

Add a new service to manage Product Variation rendering inside products

### DIFF
--- a/modules/product/src/ProductViewBuilder.php
+++ b/modules/product/src/ProductViewBuilder.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Drupal\commerce_product;
+
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityManagerInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\EntityViewBuilder;
+use Drupal\Core\Language\LanguageManagerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines the entity view builder for products.
+ */
+class ProductViewBuilder extends EntityViewBuilder {
+
+  /**
+   * The product field variation renderer.
+   *
+   * @var \Drupal\commerce_product\ProductVariationFieldRenderer
+   */
+  protected $productVariationFieldRenderer;
+
+  /**
+   * Constructs a new ProductViewBuilder.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
+   *   The entity type definition.
+   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   *   The entity manager service.
+   * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
+   *   The language manager.
+   * @param \Drupal\commerce_product\ProductVariationFieldRenderer $product_variation_field_renderer
+   *   The product variation field renderer.
+   */
+  public function __construct(EntityTypeInterface $entity_type, EntityManagerInterface $entity_manager, LanguageManagerInterface $language_manager, ProductVariationFieldRenderer $product_variation_field_renderer) {
+    parent::__construct($entity_type, $entity_manager, $language_manager);
+    $this->productVariationFieldRenderer = $product_variation_field_renderer;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function createInstance(ContainerInterface $container, EntityTypeInterface $entity_type) {
+    return new static(
+      $entity_type,
+      $container->get('entity.manager'),
+      $container->get('language_manager'),
+      $container->get('commerce_product.variation_field_renderer')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterBuild(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
+    $product_type_storage = $this->entityManager->getStorage('commerce_product_type');
+    $variation_storage = $this->entityManager->getStorage('commerce_product_variation');
+
+    /** @var \Drupal\commerce_product\Entity\ProductTypeInterface $product_type */
+    $product_type = $product_type_storage->load($entity->bundle());
+    if ($product_type->shouldInjectVariationFields() && $entity->getDefaultVariation()) {
+      $variation = $variation_storage->loadFromContext($entity);
+      $rendered_fields = $this->productVariationFieldRenderer->renderFields($variation, $view_mode);
+      foreach ($rendered_fields as $field_name => $rendered_field) {
+        // Group attribute fields to allow them to be excluded together.
+        if (strpos($field_name, 'attribute_') !== FALSE) {
+          $build['variation_attributes']['variation_' . $field_name] = $rendered_field;
+        }
+        else {
+          $build['variation_' . $field_name] = $rendered_field;
+        }
+      }
+    }
+  }
+
+}
--- a/modules/product/src/ProductViewBuilder.php
+++ b/modules/product/src/ProductViewBuilder.php
@@ -20,10 +20,10 @@ class ProductViewBuilder extends EntityViewBuilder {
    *
    * @var \Drupal\commerce_product\ProductVariationFieldRenderer
    */
-  protected $productVariationFieldRenderer;
+  protected $variationFieldRenderer;
 
   /**
-   * Constructs a new ProductViewBuilder.
+   * Constructs a new ProductViewBuilder object.
    *
    * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
    *   The entity type definition.
@@ -31,12 +31,12 @@ class ProductViewBuilder extends EntityViewBuilder {
    *   The entity manager service.
    * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
    *   The language manager.
-   * @param \Drupal\commerce_product\ProductVariationFieldRenderer $product_variation_field_renderer
+   * @param \Drupal\commerce_product\ProductVariationFieldRenderer $variation_field_renderer
    *   The product variation field renderer.
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityManagerInterface $entity_manager, LanguageManagerInterface $language_manager, ProductVariationFieldRenderer $product_variation_field_renderer) {
+  public function __construct(EntityTypeInterface $entity_type, EntityManagerInterface $entity_manager, LanguageManagerInterface $language_manager, ProductVariationFieldRenderer $variation_field_renderer) {
     parent::__construct($entity_type, $entity_manager, $language_manager);
-    $this->productVariationFieldRenderer = $product_variation_field_renderer;
+    $this->variationFieldRenderer = $variation_field_renderer;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


plopesc:
@mglaman I worked on this this morning, because we found that this could be useful for SO catalog pages.
I know this would need some tests and improvements, but take it as a proof of concept.

Thanks :)

bojanz:
Not convinced at the moment. What's your actual use case?

Having a single service that could be swapped probably makes more sense than a resolver chain.

mglaman:
I don't think we need to have it be resolvable. I think we should move the logic out of `commerce_product_commerce_product_view`, put it in custom ViewBuilder. Maybe have the ViewBuilder invoke a service? 

Either way - you just need to change the default ViewBuilder or swap out the rendering service at runtime.

Right now the hook isn't swappable.

plopesc:
So, what you think it's a better approach?? 
The custom view builder, adding there the hook logic, or add a swappable service called from the custom view builder?? 
Thanks!

mglaman:
@plopesc View builder, entity definition alter, swap w/ own builder.

Seems a little silly to have a view builder call service and swap that out. Doing same thing, but adding odd complexity.

Unless you can find other wins by doing a canonical view builder and the service being swappaple. Otherwise I vote just ViewBuilder that we swap as needed.

plopesc:
@mglaman worked on your comments.
I know this will need some improvements, but let's see if this is closer to your thoughts.

Thanks :)

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+      $container->get('commerce_product.variation_field_renderer')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildMultiple(array $build_list) {
+    // Build the view modes and display objects.
+    $view_modes = array();
```
mglaman: Convert to `[]`

mglaman: But I just realized this is the parent method copy+paste, so that's why.

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+    foreach ($view_modes as $view_mode => $view_mode_entities) {
+      $displays = EntityViewDisplay::collectRenderDisplays($view_mode_entities, $view_mode);
+      $this->buildComponents($build_list, $view_mode_entities, $displays, $view_mode);
+      foreach (array_keys($view_mode_entities) as $key) {
+        // Allow for alterations while building, before rendering.
+        $entity = $build_list[$key][$entity_type_key];
+        $display = $displays[$entity->bundle()];
+
+        // Add product variation fields to the render array
+        $this->addProductVariationFields($build_list[$key], $entity, $view_mode, $key);
```
mglaman: So we have method invoked here, adds it to normal build process before invoking `entity_view` hooks.

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+   * Adds the product variation fields to the product render array.
+   *
+   * @param array $build
+   *   The entity render array.
+   * @param $product
+   *   The product being rendered.
+   * @param $view_mode
+   *   The view mode used to render the product.
+   */
+  protected function addProductVariationFields(array &$build, $product, $view_mode) {
```
mglaman: I like that it is in a method here, so to override we extend class and simply change the method logic and contents.

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+use Drupal\Core\Entity\EntityViewBuilder;
+use Drupal\Core\Entity\FieldableEntityInterface;
+use Drupal\Core\Language\LanguageManagerInterface;
+use Drupal\Core\Render\Element;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class ProductViewBuilder extends EntityViewBuilder {
+
+  /**
+   * The product field variation renderer
```
mglaman: Needs to end with `.`

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+   * Constructs a new BlockViewBuilder.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
+   *   The entity type definition.
+   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   *   The entity manager service.
+   * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
+   *   The language manager.
+   * @param \Drupal\commerce_product\ProductVariationFieldRenderer $product_variation_field_renderer
+   *   The product variation field renderer
```
mglaman: Needs to end with `.`

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+        $this->addProductVariationFields($build_list[$key], $entity, $view_mode, $key);
+
+        $this->moduleHandler()->invokeAll($view_hook, [&$build_list[$key], $entity, $display, $view_mode]);
+        $this->moduleHandler()->invokeAll('entity_view', [&$build_list[$key], $entity, $display, $view_mode]);
+
+        $this->alterBuild($build_list[$key], $entity, $display, $view_mode);
+
+        // Assign the weights configured in the display.
+        // @todo: Once https://www.drupal.org/node/1875974 provides the missing
+        //   API, only do it for 'extra fields', since other components have
```
mglaman: extra space

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+        //   been taken care of in EntityViewDisplay::buildMultiple().
```
mglaman: extra space

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+
+    return $build_list;
+  }
+
+  /**
+   * Adds the product variation fields to the product render array.
+   *
+   * @param array $build
+   *   The entity render array.
+   * @param $product
```
mglaman: Missing type

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+   *   The product being rendered.
+   * @param $view_mode
```
mglaman: Missing type

modules/product/src/ProductViewBuilder.php
mglaman: But lets add entity interface type hint to `$product`

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,139 @@
...
+          $build['variation_attributes']['variation_' . $field_name] = $rendered_field;
+        }
+        else {
+          $build['variation_' . $field_name] = $rendered_field;
+        }
+      }
+    }
+  }
+
+}
```
mglaman: Needs newline

mglaman (changes requested):
@plopesc looks great, to me! Tests pass :D just nitpicks on coding style (which are the only CI failures.)

modules/product/src/ProductViewBuilder.php
plopesc: This is Copy & Paste from EntityViewBuilder, but removing them :)

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * The product field variation renderer.
+   *
+   * @var \Drupal\commerce_product\ProductVariationFieldRenderer
+   */
+  protected $productVariationFieldRenderer;
+
+  /**
+   * Constructs a new BlockViewBuilder.
```
bojanz: Constructs a new ProductViewBuilder.

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,140 @@
...
+   *   The product being rendered.
+   * @param string $view_mode
+   *   The view mode used to render the product.
+   */
+  protected function addProductVariationFields(array &$build, ProductInterface $product, $view_mode) {
+    /** @var \Drupal\commerce_product\Entity\ProductTypeInterface $product_type */
+    $product_type = $this->entityManager->getStorage('commerce_product_type')
+      ->load($product->bundle());
+    if ($product_type->shouldInjectVariationFields() && $product->getDefaultVariation()) {
+
```
bojanz: Unneeded newline.

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,140 @@
...
+   * @param array $build
+   *   The entity render array.
+   * @param \Drupal\commerce_product\Entity\ProductInterface $product
+   *   The product being rendered.
+   * @param string $view_mode
+   *   The view mode used to render the product.
+   */
+  protected function addProductVariationFields(array &$build, ProductInterface $product, $view_mode) {
+    /** @var \Drupal\commerce_product\Entity\ProductTypeInterface $product_type */
+    $product_type = $this->entityManager->getStorage('commerce_product_type')
```
bojanz: Let's assign the storages to their own variables before using them, doing ->load() in the next row is ugly.

modules/product/src/ProductViewBuilder.php
```diff
@@ -0,0 +1,140 @@
...
+      $container->get('entity.manager'),
+      $container->get('language_manager'),
+      $container->get('commerce_product.variation_field_renderer')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildMultiple(array $build_list) {
```
bojanz: I don't think we need to override buildMultiple(). The base class has a alterBuild() method documented as "Specific per-entity building.", which is precisely our use case. We can move the addProductVariationFields() code there and call it a day.

bojanz (changes requested):

modules/product/src/ProductViewBuilder.php
plopesc: Problem is that alterBuild() method is called after hook_ENTITY_TYPE_view() and hook_entity_view() are invoked.

I think it's more useful to do the addProductVariationFields() before that, because this info can be useful to be available in those hooks. On the other hands, hook_entity_view_alter() can be implemented in those cases where it's necessary to modify the variation data.

What do you think?

bojanz: I think it's a general core problem, we should open a core issue about moving those hooks before the alter method. In any case, doesn't affect us at this moment.

plopesc:
@mglaman @bojanz alterBuld() approach built from scratch to remove previous approach from the git log.

Thanks

mglaman:
@plopesc I did a quick force push. The code in `commerce_product.module` wasn't removed in the recent update. fixed that quick!

PR #512, opened 2016-09-24
Issue #2803241: Create a "calculated price" formatter

mglaman:


mglaman:
@bojanz I just added unit test coverage for price resolving, and kernel test for default and this formatter.

PR #513, opened 2016-09-28
2807567 cart finalize checkout error

mglaman:


modules/cart/commerce_cart.module
```diff
@@ -112,7 +112,7 @@ function commerce_cart_commerce_order_presave(OrderInterface $order) {
     $from_states = $place_transition->getFromStates();
     $to_state = $place_transition->getToState()->getId();
     if (in_array($order->original->state->value, array_keys($from_states)) && $order->state->value == $to_state) {
-      \Drupal::service('commerce_cart.cart_provider')->finalizeCart($order);
+      \Drupal::service('commerce_cart.cart_provider')->finalizeCart($order, !isset($order->original));
```
mglaman: Late night coding. We can just always set `FALSE`....we are in presave.

modules/cart/src/CartProvider.php
```diff
@@ -107,9 +107,11 @@ public function createCart($order_type, StoreInterface $store, AccountInterface
   /**
    * {@inheritdoc}
    */
-  public function finalizeCart(OrderInterface $cart) {
+  public function finalizeCart(OrderInterface $cart, $save_cart = TRUE) {
```
mglaman, line 110: Why did I add this back. It's presave.

PR #517, opened 2016-09-28
Issue #2808345 - Capitalize 'order item types'

steveoliver:


bojanz:
Looks good. Open a matching drupal.org issue so you get credited there as well.

steveoliver:
Opened https://www.drupal.org/node/2808345

PR #518, opened 2016-09-28
Issue #2801061 - Rename 'product_variation' order item type to 'default'.

steveoliver:


bojanz:
Reviewed with Matt Glaman, we agreed to proceed with this PR.
The tests are failing though (AddToCartFormTest), needs to be fixed.

PR #520, opened 2016-09-29
Issue #2808813 by steveoliver: New user account should own previous guest checkout order(s)

steveoliver:
First failing test for this feature.


bojanz:
Responded on the issue (#2808813), we'll be adding this code to docs instead.

PR #521, opened 2016-10-02
Issue - #2809967: Modules and their associated plugins should provide plugin.yml 

scotthooker:
Add plugin_type.yml to commerce_checkout similar to how address does.

### DIFF
--- a/modules/checkout/commerce_checkout.plugin_type.yml
+++ b/modules/checkout/commerce_checkout.plugin_type.yml
@@ -3,10 +3,10 @@ commerce_checkout.checkout_pane:
   provider: commerce_checkout
   plugin_manager_service_id: plugin.manager.commerce_checkout_pane
   plugin_configuration_schema_id: commerce_checkout.commerce_checkout_pane.[id]
-  plugin_definition_decorator_class: Drupal\commerce_checkout\PluginDefinition\CheckoutPanePluginDefinitionDecorator
+  plugin_definition_decorator_class: Drupal\plugin\PluginDefinition\ArrayPluginDefinitionDecorator
 commerce_checkout.checkout_flow:
   label: Commerce checkout flow
   provider: commerce_checkout
   plugin_manager_service_id: plugin.manager.commerce_checkout_flow
   plugin_configuration_schema_id: commerce_checkout.commerce_checkout_flow.[id]
-  plugin_definition_decorator_class: Drupal\commerce_checkout\PluginDefinition\CheckoutFlowPluginDefinitionDecorator
+  plugin_definition_decorator_class: Drupal\plugin\PluginDefinition\ArrayPluginDefinitionDecorator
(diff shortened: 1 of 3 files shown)


modules/checkout/commerce_checkout.plugin_type.yml
```diff
@@ -0,0 +1,12 @@
+commerce_checkout.checkout_pane:
+  label: Commerce checkout pane
+  provider: commerce_checkout
+  plugin_manager_service_id: plugin.manager.commerce_checkout_pane
+  plugin_configuration_schema_id: commerce_checkout.commerce_checkout_pane.[id]
+  plugin_definition_decorator_class: Drupal\commerce_checkout\PluginDefinition\CheckoutPanePluginDefinitionDecorator
```
mglaman: We don't have this class.

modules/checkout/commerce_checkout.plugin_type.yml
```diff
@@ -0,0 +1,12 @@
...
+commerce_checkout.checkout_flow:
+  label: Commerce checkout flow
+  provider: commerce_checkout
+  plugin_manager_service_id: plugin.manager.commerce_checkout_flow
+  plugin_configuration_schema_id: commerce_checkout.commerce_checkout_flow.[id]
+  plugin_definition_decorator_class: Drupal\commerce_checkout\PluginDefinition\CheckoutFlowPluginDefinitionDecorator
```
mglaman: This also does not exist.

mglaman (changes requested):

scotthooker:
As there is nothing special about our plugins we can just provide the default one from plugin. 
What other commerce sub modules contain plugins? 
1. Checkout
   - checkout flow
   - checkout pane
     ...

modules/checkout/commerce_checkout.plugin_type.yml
scotthooker: As there is nothing special about our plugins we can just provide the default one from plugin. Updated.

modules/checkout/commerce_checkout.plugin_type.yml
scotthooker: As there is nothing special about our plugins we can just provide the default one from plugin. Updated

mglaman:
No idea. As far as I know this is just for the Plugin module, and not 100% sure what this really gains us. Just some developer experience around plugins? Or what does it provide as a direct gain.

scotthooker:
We supply one in address module - as far as the developer experience goes I use it like I do the old views_field list in d7. 

I included it in my own custom module prior to porting it to here once seeing we include it in address.

scotthooker:
@bojanz any specific reason that you added the definition to address module?

bojanz:
It's integration for the Plugin contrib, Bart wanted it, I figured it doesn't hurt.

Our other plugin types live in commerce_payment and commerce_promotion.

scotthooker:
I'll add the yaml for them too. I think its quite handy.

scotthooker:
Added definitions for payment and promotion

bojanz:
Def doesn't hurt. Added trailing newlines and some spaces, committed in 7c327cb0288e100e6d01de45f01c549604563c72.

PR #525, opened 2016-10-07
Issue #2794909 by drugan, mglaman: Change step property of the quantity field

drugan:
Well, this solution is totally different and I've decided to make new PR.

The quick patch and explanation are[ there](https://www.drupal.org/node/2794909#comment-11700553).


modules/order/src/Entity/OrderItem.php
```diff
@@ -108,6 +110,26 @@ public function getQuantity() {
...
+    $settings = $this->getQuantityWidgetSettings();
+    $step = isset($settings['#step']) && is_numeric($settings['#step']) ? $settings['#step'] + 0 : 1;
+    $quantity = $this->getQuantity();
+    return (string) is_int($step) ? $quantity : (is_float($step) && $quantity > 0 ? '1' : $quantity);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getQuantityWidgetSettings() {
```
mglaman, line 127: Why should the entity care the widget settings? It's just storing a value.

modules/order/src/Entity/OrderItem.php
```diff
@@ -108,6 +110,26 @@ public function getQuantity() {
   /**
    * {@inheritdoc}
    */
+  public function getItemsQuantity() {
```
mglaman, line 108: Why this over get quantity?

modules/order/src/Entity/OrderItem.php
```diff
@@ -245,9 +267,16 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setDescription(t('The number of purchased units.'))
       ->setReadOnly(TRUE)
       ->setSetting('unsigned', TRUE)
+      ->setSetting('precision', 12)
+      ->setSetting('scale', 3)
+      ->setSetting('max', '367384.999')
```
mglaman: Don't these values "just work" when using a decimal base field?

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
```diff
@@ -0,0 +1,248 @@
...
+ * )
+ */
+class ModeNumberWidget extends NumberWidget {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
+    return array(
+      'placeholder' => NULL,
```
mglaman: These values need to be defined in the schema.yml

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
```diff
@@ -0,0 +1,248 @@
...
+      'step' => NULL,
+      'prefix' => NULL,
+      'suffix' => NULL,
+    ) + parent::defaultSettings();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
```
mglaman: We're not able to use any of the parent class settings?

mglaman (dismissed):
The widget looks OK. I don't agree with the other changes. The TypedData system will return proper values, there is no need for us to couple the entity to field widget settings.

modules/order/src/Entity/OrderItem.php
drugan: I agree that for now TypedData system returns proper quantity field values just because they are saved "As Is" in the database, i.e.- in decimal type with precision 10 and scale 2. That means even if you enter, let's say "5" integer value on the "Add to cart" quantity field and save it and look into the commerce_order_item.quantity in the database, you'll see this: "5.00". Not an issue so far because the value is rounded or cast to integer value when rendering Shopping Cart Block and Shopping Cart Summary view.

See: modules/cart/src/Plugin/views/field/EditQuantity.php +98
See: modules/cart/src/Plugin/Block/CartBlock.php + 132

OK, but what if the value were entered as a decimal, like this: "5.5"? Nothing bad really if you calculate an order item total at the price $1.99 for 1 (ONE!) item (5.5 \* 1.99 = $10.945), but the quantity displayed in the Shopping Cart Block and Shopping Cart Summary view will be "5", not "5.5".

The question is (not mine, the customer's question): Why I see the right total ($10.945) but wrong quantity (5) in my Shopping Cart? Is it a professional site? Or may be it is a fraud one? Why I can't enter decimal value on a Shopping Cart Summary "EditQuantity" field despite on the "Add to Cart" it was allowed?

The main question is: if we allow decimal values on the quantity what the "quantity" is? Is it a (combined) number of items on a particular order item type or it is the amount of something?

The answer is that: they are both. If the value entered is of a decimal type then the number of items on that order item is always === 1, otherwise the number of items is the actual integer value of a quantity field.

Example 1: you obtain only one (not divisible) item if you've bought fiber cable at the quantity === length === 2.5 (any dimension units).

Example 2: you obtain 2 (separate) items if you've bought T-shirts at the quantity === items === 2 (items).

OK, but how we decide what the type of number were entered, given that the value in any case saved in the database as decimal?

The cool HTML5 number type input element gives us a solution, specifically the #step attribute. Actually it restricts the value allowed to enter not only per "step" level but per number type level too, with a maximum scale checked.

Example: given step = 1.05, min = 1 and max = 9999 you restricted to the range 1-9998.05 with the first possible value after minimum is 2.05. Look, not 2 or 2.5 or 2.01 or 2.005, etc., but strictly 2.05 with the next value 3.1 and next 4.15 etc.. Our Drupal js checks the value on the client and even does not allow to save a cart displaying pretty pop-ups for a user with an error description and closest valid values that could be entered.

Client side? But, what if ...

It is not an issue because this patch does not rely on a user input taking into account settings on an order item type's quantity made by administrator. Yes, the malicious input can cause weird results on a price totals and number of items in the cart but not broke the whole system except the Shopping Cart of that clever user. For now these "experiments" still could be done because price or tax calculation code taking whatever the quantity number is multiplyed by a raw price amount. I think that it might be useful if this code also call $order_item->getQuantityWidgetSettings() first and checked validity of a quantity in respect to #min, #max and #step settings in a try{} block and throw an Exception if something is wrong. That will give us an integrity for the quantity <-> price_total bundle and for the whole commerce workflow.

See patched: modules/order/src/Entity/OrderItem.php +115

modules/order/src/Entity/OrderItem.php
drugan: It's a basic thing, read above.

modules/order/src/Entity/OrderItem.php
drugan: We can put #min and #step settings here but we will be restricted by those settings on ModeNumberWidget form. For example I've added #max =  '367384.999' here on a base field. This setting could be decreased on a form display mode settings form but not increased. That is not the matter for the #max because it is most possible value for precision 12, scale 3 field. Conversely, it's a good thing preventing some unnecessary errors on the form, but if put here #min = 1, for example, we could never set #min < 1 on a form display mode settings form, i.e.- values like 0.001 will be forbidden. The same issue with #step.

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
drugan: Well, why not. The only condition they all need to be set to NULL as default values.

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
drugan: Actually, these settings (#min, #max, #prefix, #suffix, #step) nor the parent NumberWidget neither grandparent WidgetBase class settings form. They are the settings from NumericItemBase  class which is the parent of DecimalItem, IntegerItem and FloatItem classes and cannot be fetched there. But even if it could be fetched then there is no any point on this, because NumericItemBase form handles these settings on the very base level. Another words, this form have to be totally rewritten. I dare to say that ModeNumberWidget class settings a bit better than in NumericItemBase and I'll to try open an issue on the /core. Actually this class  from the start was written with the core target in mind.

As for the DecimalItem class (ours) settings they are '' mostly empty strings except precision = 12, and scale 3 settings which are set in our code  and we use them to decide the most possible number of digits after decimal for the #step setting of the widget form settings #step itself.

See patched: src/Plugin/Field/FieldWidget/ModeNumberWidget.php +48, +230 
See: core/lib/Drupal/Core/Field/Plugin/Field/FieldType/NumericItemBase.php

modules/order/src/Entity/OrderItem.php
drugan: Not sure if it is the right way to express this in plain english but I think what we need is to adopt the following concept on the order item quantity field:

**QUANTITY = NUMBER_OF_ITEMS ^ NUMBER_OF_UNITS**

Where **ITEM** is any separate item that must be measured in integer value == 1.
Where **UNIT** is any non-zero unit of the **ITEM**  that must be measured in decimal up to 3 digits after decimal sign and where **ITEM** is a multiple of  **UNITS**.

Given the pattern above you can decide the type of a quantity value:

$quantity = $order_item->getItemsQuantity();

if ($quantity === '1') {
// the $quantity is a decimal type **NUMBER_OF_UNITS** like this: N.000, N.N00, N.NN0, N.NNN
$units_quantity = $order_item->getQuantity();
// Do something with the $units_quantity
}
else {
// the $quantity is an "integer" type **NUMBER_OF_ITEMS** like this: N.000
// Do something with the $quantity
}

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
mglaman: No, literally, they have to be. Review your failed tests.

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
mglaman: Yeah after I commented I forgot the whole reason we need a widget is the core doesn't provide these settings.

modules/order/src/Entity/OrderItem.php
mglaman: It sounds like your use case is going past where we would have something in core Commerce.

Quantity is "I want this many of this thing." Where "this thing" is the representation of a purchaseable entity.

It sounds like the bugs are actually in the cart widget.

None of this justified any of the changes to the order item entity, as far as I can understand. It's still in the field widget and other display areas. Also the fact we need proper rounding in our price object.

Case and point:

```
$count += (int) $order_item->getQuantity();
```

The bug is "remove the `(int)` cast."

And then in quantity

```
'#default_value' => round($quantity),
```

Remove rounding.

So by adding the custom field widget and removing that casting and that round.. we should resolve the entire issue.

Anything beyond that appears to be use case specific to your project and the terminology used. An order item is an **item** and **unit** as you described. You're putting two terms based on if it has a decimal quantity or not.

modules/order/src/Entity/OrderItem.php
mglaman: An entity should not be aware of, or care for, the widget settings of a field. They're two completely different API levels.

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
drugan: Well, looked into the test once again and all that I've found is this:

src/Plugin/Field/FieldWidget/ModeNumberWidget.php

Line 7
unused use statement need to be removed.

Line 62, 67
leading white space in descriptions need to be removed.

Line 74, 115, 122

t() wrapper for dynamic string values need to be removed.

Could you point where Travis complains about undefined default values?

modules/order/src/Entity/OrderItem.php
drugan: I agree with you but in our example getQuantityWidgetSettings() is only exposed centralized functionality. Without this method exposed every class operating on order item objects need actually to duplicate the method if they need to check widget settings.

Well, to comply with API requirements could we define a separate service?

modules/order/src/Entity/OrderItem.php
drugan: The (int) casting could be removed now but I'm not agree that this was a reason for a quantity bug. It was quick and dirty solution to prevent displaying quantities like this in the  Shopping Cart Block:
`
8.25 items
`

As for the rounding in the price object this obviously have to be done keeping in mind possible decimal quantity values. I need to find the place where it happens.

drugan: I wonder does anybody know why it were decided to define quantity field as decimal?

As I see now this decision drastically extends drupalcommerce possibilities, but what was the point at the first place, keeping in mind a lot of troubles issuing from the decimal quantity. All these troubles could be killed just changing one line#243 in  modules/order/src/Entity/OrderItem.php:
`
$fields['quantity'] = BaseFieldDefinition::create('integer')
`

That's it. But if this was not done till now it means there is somebody amongst commerce guys who thinks the same as me.

src/Plugin/Field/FieldWidget/ModeNumberWidget.php
drugan: Added [core feature request](https://www.drupal.org/node/2816859).

I think that now we need to wait for decision on the core. If the patch will be accepted then this PR may loose its point.

drugan:
@mglaman 
Seems that this [core feature request](https://www.drupal.org/node/2816859#comment-11750594) has no interest on the community so I've mirrored the code from the request into this PR. What is done:
1. Renamed and rewritten ModeNumberWidget to CommerceNumberWidget.php just to have more sensible name.
2. Added CommerceNumberWidgetTest.php for testing the widget functionality.
3. The widget settings added to commerce.schema.yml.
4. Overall code refactoring.
   
   If the core feature happens ever been accepted then the CommerceNumberWidget.php and CommerceNumberWidgetTest.php  files might be removed.

mglaman:
@drugan yeah.. I have one in core for the Ajax modals to support `button` elements when it moves things around.. it'll rot until you can hunt down a system maintainer. Let's port it in here then push for core.

I'll review

drugan:
@mglaman 
Well, Travis errors are gone and I'll really appreciate if you take the time to review this PR.

As for the "hunting system maintainer" it is a bit redundant because who we need to hunt are actual users of drupal 8 and drupalcommerce in particular. If they'll want the configurable number widget feature then they themselves post comments and ask the maintainer to commit the patch. And I am sure it happens as soon as drupalcommerce will get the first release which does not require to be installed via command line interface. The current 8.x- version moved ahead substantially and is a really cool commerce solution impatiently awaited by the community. Inadvertently they'll encounter the order item quantity field issue and end their search on this PR. As we all see some useful parts of the HTML5 number field are overlooked in drupal 8. For example, the basic #step attribute which on the fly creates subtypes of this field is defined in code but has no UI to be configured by a user. Also, in current version the default_value property is not seen on the field despite having the possibility to be configured on the UI. Above all, if the number field, like our quantity, is defined as a part of a bundle it cannot be cofigured by a user in any way except placeholder attribute. User sees "Display prefix and suffix" checkbox on the field formatter but does not know where they are. Prefix and suffix are very useful attributes of the commerce quantity field as they improve customer experience and again, keeping in mind polymorphism of this field it need to be explained on the page what it defines: items, kilograms, litres, whatever...

mglaman:
@drugan I'll review. I've been a bit tied up, and Bojan's been in the taxes/shipping realm. I also want to assure you (and any other contributors) we haven't forgotten these issues. It's just been time to focus in on core release blockers. 

I'll circle up to review this and the bulk creation one

shabananavas:
I have tested this and it works perfectly. This new field greatly improves user experience and the code looks good to me. Hope it will get merged asap.

finnef:
I'm implementing C2 for a customer and they want order item quantities stepped per 1, not the current default 0.01. So +1 for this functionality

mglaman:
Closing in favor of a more succinct #755

PR #526, opened 2016-10-07
Issue #2813583 by mglaman: Cart manager needs to refresh orders expli…

mglaman:
…citly when order items change

### DIFF
--- a/modules/cart/commerce_cart.services.yml
+++ b/modules/cart/commerce_cart.services.yml
@@ -9,7 +9,7 @@ services:
 
   commerce_cart.cart_manager:
     class: Drupal\commerce_cart\CartManager
-    arguments: ['@entity_type.manager', '@commerce_cart.order_item_matcher', '@event_dispatcher']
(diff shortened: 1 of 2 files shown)


modules/cart/commerce_cart.services.yml
```diff
@@ -9,7 +9,7 @@ services:
 
   commerce_cart.cart_manager:
     class: Drupal\commerce_cart\CartManager
-    arguments: ['@entity_type.manager', '@commerce_cart.order_item_matcher', '@event_dispatcher']
+    arguments: ['@entity_type.manager', '@commerce_cart.order_item_matcher', '@event_dispatcher', '@commerce_order.order_refresh']
```
mglaman, line 12: Discussed with @bojanz, we decided that orders just need a method to flag them as "dirty" so that on save they refresh themselves.

bojanz:
We decided on a different approach (and renamed the issue), so closing this PR.
We can open a new one when (implementation) time comes.

PR #527, opened 2016-10-11
Issue #2816059 - Cart issues after Guest checkout

steveoliver:
See https://www.drupal.org/node/2816059

### DIFF
--- a/modules/cart/src/CartSession.php
+++ b/modules/cart/src/CartSession.php
@@ -29,14 +29,14 @@ class CartSession implements CartSessionInterface {
   /**
    * {@inheritdoc}
    */
-  public function getCartIds() {
+  public function getActiveCartIds() {
     return $this->session->get('commerce_cart_orders', []);
   }
 
   /**
    * {@inheritdoc}
    */
-  public function addCartId($cart_id) {
+  public function addActiveCartId($cart_id) {
     $ids = $this->session->get('commerce_cart_orders', []);
     $ids[] = $cart_id;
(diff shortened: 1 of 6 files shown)


modules/cart/src/CartSession.php
```diff
@@ -63,6 +71,10 @@ public function deleteCartId($cart_id) {
       // Remove the empty list to allow the system to clean up empty sessions.
       $this->session->remove('commerce_cart_orders');
     }
+    // Keep a list of completed order for access checks to completed orders.
+    $completed_ids = $this->session->get('commerce_cart_completed_orders', []);
+    $completed_ids = array_merge($completed_ids, [$cart_id]);
+    $this->session->set('commerce_cart_completed_orders', $completed_ids);
```
mglaman, line 83: So should this just go in `::finalizeCart` and make sure `::finalizeCart` doesn't invoke delete? Seems weird we have `deleteCartId` but its actually doing a conversions.

steveoliver: Good call.

steveoliver: I ended up moving the handling of completed_orders to a method in CartSession (called addCompletedCartId()).

mglaman:
@bojanz can we get your consensus on this? follows 1.x pattern and fixes the false cart issue on anonymous users.

bojanz:
Committed in d7317f7a696bc32ad3d97bac7a1baa2074236162. Updating the parent issue with an explanation of API differences.

PR #532, opened 2016-10-13
Issue #2818111 - Ensure order items show up in Cart block

steveoliver:
Failing test shows that order items are not being displayed in cart block.

### DIFF
--- a/modules/cart/tests/src/Functional/CartBlockTest.php
+++ b/modules/cart/tests/src/Functional/CartBlockTest.php
@@ -42,9 +42,9 @@ class CartBlockTest extends CartBrowserTestBase {
   }
 
   /**
-   * Tests the count text (display, plurality).
+   * Tests the order items display.
    */
-  public function testOrderItemsText() {
+  public function testOrderItemsDisplay() {
     $this->drupalGet('<front>');
     $this->assertSession()->pageTextContains('0 items');
 
--- a/modules/cart/tests/src/Functional/CartBlockTest.php
+++ b/modules/cart/tests/src/Functional/CartBlockTest.php
@@ -41,4 +41,17 @@ class CartBlockTest extends CartBrowserTestBase {
     $this->assertSession()->pageTextNotContains('3 items');
   }
 
+  /**
+   * Tests the count text (display, plurality).
+   */
+  public function testOrderItemsText() {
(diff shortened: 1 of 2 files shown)


modules/cart/tests/src/Functional/CartBlockTest.php
```diff
@@ -41,4 +41,17 @@ public function testCountText() {
     $this->assertSession()->pageTextNotContains('3 items');
   }
 
+  /**
+   * Tests the order items display.
+   */
+  public function testOrderItemsDisplay() {
+    $this->drupalGet('<front>');
```
bojanz: How is this test method different from the one above? They both add an entity to cart and assert the count. Which makes me think the previous one should have caught the bug?

steveoliver: Test above only tests for the display of the 'X item[s]' text -- It's different than the order items list (view), which wasn't tested above.

modules/cart/tests/src/Functional/CartBlockTest.php
```diff
@@ -41,4 +41,17 @@ public function testCountText() {
...
+    $this->assertSession()->pageTextContains('0 items');
+
+    $this->cartManager->addEntity($this->cart, $this->variation);
+    $this->drupalGet('<front>');
+    $this->assertSession()->pageTextContains('1 item');
+    $this->assertSession()->pageTextContains($this->variation->getOrderItemTitle());
```
steveoliver: We could remove the assertions for X item[s] -- they are indeed exact dupes of the first test.  This line 54 checking OrderItemTitle is the previous uncaught fail.

steveoliver:
Really weird the first commit test passes on Travis. :/

modules/cart/tests/src/Functional/CartBlockTest.php
mglaman: Could we have the test just check count + title/cart block view?

steveoliver: @mglaman, I combined the tests in the latest commit.

steveoliver (commented):
As suspected, this (original state of the view) makes the new cart order item test fail.

PR #534, opened 2016-10-14
Issue #2807537 by mglaman: Implement rounding

mglaman:


modules/price/src/Price.php
```diff
@@ -126,6 +126,7 @@ public function subtract(Price $price) {
    */
   public function multiply($number) {
     $new_number = Calculator::multiply($this->number, $number);
+    $new_number = Calculator::round($new_number, 2);
```
mglaman: @bojanz I wasn't sure when we'd invoke rounding. Would it be in the price object's multiplier or elsewhere? To be honest I'd expect the receive a price value in proper format (0.00) and not need to round it on my own.

mglaman:
@bojanz I updated so that the only time rounding is executed is on `\Drupal\commerce_order\Entity\Order::recalculateTotalPrice`

modules/price/src/Calculator.php
```diff
@@ -180,6 +180,32 @@ public static function trim($number) {
...
+   *   The number to trim.
+   * @param int $precision
+   *   The number of decimal digits to round to.
+   * @param int $mode
+   *   The rounding mode.
+   *
+   * @return string
+   *   The rounded number.
+   */
+  public static function round($number, $precision = 0, $mode = PHP_ROUND_HALF_UP) {
```
czigor, line 195: $mode is never used.

drugan: I am also agree that the addition of price rounding is a good thing but it need to be implemented in more places of drupalcommerce just to see the effect.

The first useful place that I've found is in **modules/order/src/Entity/OrderItem.php** +245

```
protected function recalculateTotalPrice() {
  if ($unit_price = $this->getUnitPrice()) {
    $this->total_price = $unit_price->multiply($this->getQuantity())->round();
  }
}
```

Works great with default precision and rounding mode but it would be better to make these values configurable in **modules/price/src/Plugin/Field/FieldWidget/PriceDefaultWidget.php**

mglaman: Yeah, rounding needs to be used. I didn't finish porting in all options as `bcmath` operations.

@drugan we're only rounding in the order total, because we want the most precision. We don't want to round until the final price is calculated.

modules/price/src/Price.php
```diff
@@ -247,6 +247,22 @@ public function lessThanOrEqual(Price $price) {
...
+   *
+   * @param string $precision
+   *   The precision.
+   * @param int $mode
+   *   The rounding mode.
+   *
+   * @return static
+   *   The resulting price.
+   */
+  public function round($precision = '2', $mode = PHP_ROUND_HALF_UP) {
```
mglaman, line 260: Remove default value. Should always be based via a currency.

mglaman: Also, we have the currency code but no data. Should we keep `$precision` and force invokers to get the currency's decimal count, or find some way to make the price object aware of it.

drugan: @mglaman 
Yes, the default precision and rounding mode may cause mayhem on some currencies. Think of currencies which don't have decimals (no coins) or which want to have precision more than 2. For example, the [price for gold per ounce](http://www.apmex.com/spotprices/gold-price) is always displayed with  precision 3 even if the decimals are zeroes. In drupal 7 the parameters for each currency is hardcoded and can be fetched by a currency code but that is not a good solution, as I think, because:
1. It forces us to depend on some external resource and to maintain the currencies data up to date.
2. It forces a user to use the precision defined by this resource.

As I said earlier the best way is to make order item price widget configurable with the same default values what you've added to the current round() version. Then invokers can fetch the widget settings and build the price to be used.

bojanz:
Continued in #562.

PR #536, opened 2016-10-18
#2820034 Add item to a fresh cart fails if Editor module is enabled

agoradesign:
let the test fail...

### DIFF
--- a/modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
+++ b/modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
@@ -1,19 +1,14 @@
 <?php
 
-use Drupal\commerce_order\Entity\OrderInterface;
 use \Drupal\Core\Entity\EntityInterface;
 
 /**
- * Implements hook_entity_update().
+ * Implements hook_ENTITY_TYPE_update().
  *
  * To replicate editor.module and catch save inceptions.
  */
-function commerce_cart_test_entity_update(EntityInterface $entity) {
-  // Only act on order entities.
-  if (!($entity instanceof OrderInterface)) {
-    return;
-  }
+function commerce_cart_test_commerce_order_update(EntityInterface $entity) {
   if (!$entity->isNew() && empty($entity->original)) {
-    throw new \Exception('Found stale data! $entity->original must be present on updated entity!');
+    throw new \Exception('Found stale data! $entity->original not found on updated entity!');
   }
 }
--- a/modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
+++ b/modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
@@ -1,0 +1,19 @@
+<?php
+
+use Drupal\commerce_order\Entity\OrderInterface;
+use \Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_entity_update().
+ *
+ * To replicate editor.module and catch save inceptions.
+ */
+function commerce_cart_test_entity_update(EntityInterface $entity) {
+  // Only act on order entities.
+  if (!($entity instanceof OrderInterface)) {
+    return;
+  }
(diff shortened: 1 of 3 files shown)


mglaman:
@agoradesign know what's weird? The tests pass for PHP7. So something about the "check if this should be refreshed" is wrong in different ways than we may think.

I really wish core had a `::isSaving()` method.

agoradesign:
That's really weird! Yes, isSaving() would be great.. but as mentioned in the original issue on d.o, I'm suspecting that the OrderRefresh is working on the loaded $entity->original too and trying to save that. This would be the bigger problem

modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
```diff
@@ -0,0 +1,19 @@
...
+use \Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_entity_update().
+ *
+ * To replicate editor.module and catch save inceptions.
+ */
+function commerce_cart_test_entity_update(EntityInterface $entity) {
+  // Only act on order entities.
+  if (!($entity instanceof OrderInterface)) {
```
bojanz: Why not implement hook_commerce_order_update() then?

modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
```diff
@@ -0,0 +1,19 @@
...
+    return;
+  }
+  if (!$entity->isNew() && empty($entity->original)) {
+    throw new \Exception('Found stale data! $entity->original must be present on updated entity!');
```
bojanz: This doesn't have to mean that the data is stale. Better to say '$entity->original not found', and perhaps make that exception EntityMalformedException?

modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
agoradesign: right... ok I'll update again...

modules/cart/tests/modules/commerce_cart_test/commerce_cart_test.module
agoradesign: PR is updated with new solution

modules/order/src/OrderStorage.php
```diff
@@ -44,6 +51,7 @@ class OrderStorage extends CommerceContentEntityStorage {
   public function __construct(EntityTypeInterface $entity_type, Connection $database, EntityManagerInterface $entity_manager, CacheBackendInterface $cache, LanguageManagerInterface $language_manager, EventDispatcherInterface $event_dispatcher, OrderRefreshInterface $order_refresh) {
     parent::__construct($entity_type, $database, $entity_manager, $cache, $language_manager, $event_dispatcher);
     $this->orderRefresh = $order_refresh;
+    $this->unchangedLoadIds = [];
```
bojanz, line 54: I just realized we can use a single flag for this, we don't need to track individual ids, they're processed one by one anyway. skipRefresh perhaps?

bojanz:
Btw, what's your refresh_frequency set to? The default is 30s, which I figure would be enough to hide this problem on most installs. I wonder if that check in needsRefresh() is somehow bypassed?

modules/order/src/OrderStorage.php
agoradesign: ok, I'll change that later this day

agoradesign:
That's a good point... Apparently I haven't taken care of this settings after updating to beta... So it's currently set to nothing :D

I'll try that with the default setting. Of course it should work with a zero value as well, but it makes the issue far less critical then

PS: I've set a refresh frequency, and the issue doesn't happen any more. Even a second should be enough during a single request

bojanz:
Committed in 0dfb84b0e335054470d5a4dda7bc90303e33f89a.

PR #537, opened 2016-10-19
Issue #2757236 - Provide workflow transition buttons on order view page

steveoliver:
PR for https://www.drupal.org/node/2757263


tests/modules/commerce_workflow_test/commerce_workflow_test.info.yml
```diff
@@ -0,0 +1,7 @@
+name: 'Commerce workflow test'
+type: module
+description: 'Module for testing commerce order workflow.'
```
bojanz: We already have an order type and a workflow, no need for this test module?

bojanz (changes requested):

tests/modules/commerce_workflow_test/commerce_workflow_test.info.yml
steveoliver: true. removed.

steveoliver:
https://github.com/drupalcommerce/commerce/pull/537 passes, unneeded test module removed.

PR #538, opened 2016-10-20
Issue #2820865: PaymentGatewayBase should implement PluginWithFormsInterface

duozersk:
PR for https://www.drupal.org/node/2820865

### DIFF
--- a/modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
+++ b/modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
@@ -11,12 +11,16 @@ use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Plugin\PluginBase;
+use Drupal\Core\Plugin\PluginWithFormsInterface;
+use Drupal\Core\Plugin\PluginWithFormsTrait;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Provides the base class for payment gateways.
  */
-abstract class PaymentGatewayBase extends PluginBase implements PaymentGatewayInterface, ContainerFactoryPluginInterface {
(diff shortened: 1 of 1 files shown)


modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
```diff
@@ -11,12 +11,16 @@
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Plugin\PluginBase;
+use Drupal\Core\Plugin\PluginWithFormsInterface;
```
bojanz: PaymentGatewayInterface already extends PluginWithFormsInterface, you don't need to add it here.

modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
```diff
@@ -11,12 +11,16 @@
...
+use Drupal\Core\Plugin\PluginWithFormsTrait;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Provides the base class for payment gateways.
  */
-abstract class PaymentGatewayBase extends PluginBase implements PaymentGatewayInterface, ContainerFactoryPluginInterface {
+abstract class PaymentGatewayBase extends PluginBase implements PaymentGatewayInterface, ContainerFactoryPluginInterface, PluginWithFormsInterface {
```
bojanz: You also need to remove the methods that are now provided by the trait.

bojanz (changes requested):

modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
duozersk: Yep, correct, fixed

modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
duozersk: Good-good, removed.

bojanz:
Thanks!

PR #539, opened 2016-10-21
Issue #2820931 - Perform payment transaction during checkout

steveoliver:
This works, but needs some cleanup (see `@todos`).  Also needs test(s) - I tried in https://gist.github.com/steveoliver/3135b75b573d0fb902ebc7cf23426505#file-paymenteventtest-php-L111, but had issues getting a payment method to be created this way.  I'd like someone else to take it from here.


modules/payment/src/EventSubscriber/PaymentListener.php
```diff
@@ -0,0 +1,69 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class PaymentListener implements EventSubscriberInterface {
+
+  /**
+   * @inheritdoc
+   */
+  public static function getSubscribedEvents() {
+    return [
+      'commerce_order.place.post_transition' => 'onOrderPostPlace',
```
mglaman: See concerns in https://www.drupal.org/node/2820931#comment-11747071

mglaman (changes requested):

modules/payment/commerce_payment.module
```diff
@@ -97,3 +101,66 @@ function template_preprocess_commerce_payment_method(array &$variables) {
     $variables['payment_method'][$key] = $variables['elements'][$key];
   }
 }
+
+/**
+ * Implements hook_form_alter().
+ */
+function commerce_payment_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if ($form_state->get('commerce_checkout_form')) {
+    $form['#validate'][] = 'commerce_payment_process_checkout_payment';
```
mglaman, line 110: I almost wonder if checkout should invoke some kind of event for each checkout step validation? "Validate STEP" and allow subscribers to set form state errors.

@bojanz thoughts?

mglaman:
Closing in favor of approach in #577

PR #540, opened 2016-10-24
Issue #2813583 by mglaman, bojanz: Expand the order refresh logic

bojanz:
- Added getRefreshState() and setRefreshState() to orders.
- Added a guarantee that order refresh frequency is always at least 1s, used it to remove the static from OrderRefresh
- Improved the order refresh frequency form element (OrderTypeForm)
- Changed the default refresh frequency to 300s (5min), default refresh mode to "customer" (matching 1.x)
- Added logic that always refreshes if the day has changed.
- Split needsRefresh into needsRefresh and shouldRefresh. Allows us to show a message to the admin saying "looks like this order needs to be refreshed, would you like to do that?"
- Rewrote the order refresh tests.
- Fixed bug where order items were saved before the processors had a chance to modify them.
- Fixed wrong typehint for $current_user.
- Copied the Drupal 8.3.x Time object into Commerce, added it to the refresh process.

### DIFF
--- a/modules/cart/src/CartManager.php
+++ b/modules/cart/src/CartManager.php
@@ -68,6 +68,7 @@ class CartManager implements CartManagerInterface {
       $order_item->delete();
     }
     $cart->setItems([]);
+    $cart->setAdjustments([]);
 
     $this->eventDispatcher->dispatch(CartEvents::CART_EMPTY, new CartEmptyEvent($cart, $order_items));
     if ($save_cart) {
--- a/modules/order/src/Entity/Order.php
+++ b/modules/order/src/Entity/Order.php
@@ -314,6 +314,24 @@ class Order extends ContentEntityBase implements OrderInterface {
   /**
    * {@inheritdoc}
    */
+  public function getRefreshState() {
+    $data = $this->getData();
+    return !empty($data['refresh_state']) ? $data['refresh_state'] : NULL;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setRefreshState($refresh_state) {
+    $data = $this->getData();
+    $data['refresh_state'] = $refresh_state;
+    $this->setData($data);
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function getData() {
     $data = [];
     if (!$this->get('data')->isEmpty()) {
@@ -391,6 +409,10 @@ class Order extends ContentEntityBase implements OrderInterface {
       }
     }
 
+    // Refresh draft orders on every save.
+    if ($this->getState()->value == 'draft' && empty($this->getRefreshState())) {
+      $this->setRefreshState(self::REFRESH_ON_SAVE);
+    }
     $this->recalculateTotalPrice();
   }
 
@@ -458,6 +480,7 @@ class Order extends ContentEntityBase implements OrderInterface {
     // If no order number has been set explicitly, set it to the order ID.
     if (!$this->getOrderNumber()) {
       $this->setOrderNumber($this->id());
+      $this->setRefreshState(self::REFRESH_SKIP);
       $this->save();
     }
 
--- a/modules/order/src/OrderRefresh.php
+++ b/modules/order/src/OrderRefresh.php
@@ -2,12 +2,12 @@
 
 namespace Drupal\commerce_order;
 
+use Drupal\commerce\TimeInterface;
 use Drupal\commerce_order\Entity\OrderInterface;
 use Drupal\commerce_order\Entity\OrderType;
-use Drupal\commerce_price\Resolver\ChainPriceResolver;
 use Drupal\commerce_price\Resolver\ChainPriceResolverInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
-use Drupal\Core\Session\AccountProxyInterface;
+use Drupal\Core\Session\AccountInterface;
 
 /**
  * Default implementation for order refresh.
@@ -36,13 +36,18 @@ class OrderRefresh implements OrderRefreshInterface {
   protected $currentUser;
 
   /**
+   * The time.
+   *
+   * @var \Drupal\commerce\TimeInterface
+   */
+  protected $time;
+
+  /**
    * The order processors.
    *
    * @var \Drupal\commerce_order\OrderProcessorInterface[]
    */
-  protected $processors;
-
-  protected static $refreshingOrders = [];
+  protected $processors = [];
 
   /**
    * Constructs a new OrderRefresh object.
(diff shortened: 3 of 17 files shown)


bojanz:
I've split off and committed https://www.drupal.org/node/2823523 and https://www.drupal.org/node/2823524 from this PR.

modules/cart/src/CartManager.php
```diff
@@ -68,6 +68,7 @@ public function emptyCart(OrderInterface $cart, $save_cart = TRUE) {
       $order_item->delete();
     }
     $cart->setItems([]);
+    $cart->setAdjustments([]);
```
mglaman, line 71: Feels weird to have to call this. I know we need to. I wonder if orders need a "clear me" method. I could see this being a missing error in contributed code. But also not sure about adding yet another method.

modules/order/src/Entity/Order.php
```diff
@@ -314,6 +314,24 @@ public function getState() {
   /**
    * {@inheritdoc}
    */
+  public function getRefreshState() {
+    $data = $this->getData();
+    return !empty($data['refresh_state']) ? $data['refresh_state'] : NULL;
```
mglaman, line 319: Another side topic, what if `getData` had a `$key = NULL` to take care of a lot of this logic?

modules/order/src/OrderRefresh.php
```diff
@@ -98,41 +97,65 @@ public function needsRefresh(OrderInterface $order) {
...
+    foreach ($order->getItems() as $order_item) {
+      $order_item->setAdjustments([]);
 
-      foreach ($this->processors as $processor) {
-        $processor->process($order);
+      $purchased_entity = $order_item->getPurchasedEntity();
+      if ($purchased_entity) {
+        $order_item->setTitle($purchased_entity->getOrderItemTitle());
+        $unit_price = $this->chainPriceResolver->resolve($purchased_entity, $order_item->getQuantity());
+        $order_item->setUnitPrice($unit_price);
```
mglaman, line 144: Should we update the changed time for unit items here, too?

modules/order/src/Entity/Order.php
```diff
@@ -458,6 +480,7 @@ public function postSave(EntityStorageInterface $storage, $update = TRUE) {
     // If no order number has been set explicitly, set it to the order ID.
     if (!$this->getOrderNumber()) {
       $this->setOrderNumber($this->id());
+      $this->setRefreshState(self::REFRESH_SKIP);
```
mglaman, line 483: We set it as `REQUEST_SKIP` because order storage will then set it as `NULL` correct?

modules/order/src/OrderStorage.php
```diff
@@ -83,10 +85,33 @@ public function loadUnchanged($id) {
...
    * {@inheritdoc}
    */
+  protected function doPreSave(EntityInterface $entity) {
+    $id = parent::doPreSave($entity);
+    /** @var \Drupal\commerce_order\Entity\OrderInterface $entity */
+    if ($entity->getRefreshState() == OrderInterface::REFRESH_ON_SAVE) {
+      $this->orderRefresh->refresh($entity);
+    }
+    // Only the REFRESH_ON_LOAD state needs to be persisted on the entity.
+    if ($entity->getRefreshState() != OrderInterface::REFRESH_ON_LOAD) {
```
mglaman, line 95: To be honest it took me a bit to grok this.

I would if 

```
switch (refreshState) :

  case REFRESH_ON_LOAD:
     // We do nothing so this persists.
  case REFRESH_ON_SAVE:
    refresh($entity);
   // no break so it also uses skip/default

  case REFRESH_SKIP:
  default:
    setRefreshState(NULL);
  break;

endswitch
```

Little bit easier to understand.

modules/order/src/OrderRefresh.php
```diff
@@ -98,41 +97,65 @@ public function needsRefresh(OrderInterface $order) {
...
+
+    // Allow the processors to modify the order and its items.
+    foreach ($this->processors as $processor) {
+      $processor->process($order);
+    }
 
-      $order->save();
-      unset(self::$refreshingOrders[$order->id()]);
+    // @todo Evaluate which order items have changed.
+    foreach ($order->getItems() as $order_item) {
```
mglaman, line 154: We want to save this before running order processors, no?

mglaman: Or, I guess not. We have the unit and quantity as needed. This way they aren't save until we've prepared to exit the refresh process.

modules/order/src/OrderStorage.php
```diff
@@ -83,10 +85,33 @@ public function loadUnchanged($id) {
...
+   */
   protected function postLoad(array &$entities) {
-    foreach ($entities as $entity) {
-      if (!$this->skipRefresh && $this->orderRefresh->needsRefresh($entity)) {
-        $this->orderRefresh->refresh($entity);
+    if (!$this->skipRefresh) {
+      /** @var \Drupal\commerce_order\Entity\OrderInterface[] $entities */
+      foreach ($entities as $entity) {
+        $explicitly_requested = $entity->getRefreshState() == OrderInterface::REFRESH_ON_LOAD;
+        if ($explicitly_requested || $this->orderRefresh->shouldRefresh($entity)) {
```
mglaman, line 110: I like this a lot. We'll only refresh on load if it's been a day or more. I'd expect most transactions to complete in under 5 minutes anyways.

mglaman (commented):
Here's my feedback and thoughts @bojanz. Looks good and makes sense. A hell of a lot better than what we have now.

modules/cart/src/CartManager.php
bojanz: Yes, this is just a quick fix, setItems() / addItem / removeItem should clear adjustments each time.

modules/order/src/Entity/Order.php
bojanz: Good idea!

modules/order/src/Entity/Order.php
bojanz: We set it as REFRESH_SKIP because the order is being saved right after being saved (during which it was also refreshed). I'll add a comment.

modules/order/src/OrderRefresh.php
bojanz: If we did that then order processors would be useless for order items, which would be both confusing and limiting.

modules/order/src/OrderStorage.php
bojanz: I don't see how you got that impression, we still check order frequency like before?

modules/order/src/OrderRefresh.php
bojanz: Definitely.

modules/order/src/OrderRefresh.php
mglaman: Makes sense

bojanz:
Merged in 1f2ab8ebc9b2257574328348639c6c33acd4eb3c.

PR #542, opened 2016-10-25
Issue #2822063 - Onsite payment gateway plugins determine payment capture default

steveoliver:
See https://www.drupal.org/node/2822063

### DIFF
--- a/modules/payment_example/src/Plugin/Commerce/PaymentGateway/Onsite.php
+++ b/modules/payment_example/src/Plugin/Commerce/PaymentGateway/Onsite.php
@@ -84,6 +84,16 @@ class Onsite extends OnsitePaymentGatewayBase implements OnsiteInterface {
   /**
    * {@inheritdoc}
    */
+  public function shouldCapturePaymentByDefault() {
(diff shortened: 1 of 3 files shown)


modules/payment_example/src/Plugin/Commerce/PaymentGateway/Onsite.php
```diff
@@ -84,6 +84,16 @@ public function submitConfigurationForm(array &$form, FormStateInterface $form_s
   /**
    * {@inheritdoc}
    */
+  public function shouldCapturePaymentByDefault() {
+    // Determine default capture flag from current mode (test / live).
+    // Another way for payment gateways to determine this value could be for
+    // them to evaluate the value of another field specific to this setting.
+    return ($this->configuration['mode'] === 'live');
```
mglaman: I think by default we can just have it return `TRUE` live or sandbox doesn't affect capture immediately. Payment gateways supporting modes can use their own logic here.

mglaman:
Closing, as it's now part of the pane proposed in #577

PR #543, opened 2016-11-02
Catch exceptions thrown by checkout panes

czigor:
https://www.drupal.org/node/2824193


czigor:
This is not a good solution at least for caching reasons. That is: if we have no payment gateways, the checkout page will be cached in the (dynamic) page cache. Adding a payment gateway does not invalidate this cache.

bojanz:
Yeah, this is the wrong idea. Exceptions are meant to end the world. Our only mistake is using an exception when there are no payment gateways, instead of setting a form error.

PR #544, opened 2016-11-02
Remove PaymentInformation Exception

czigor:


czigor:
https://www.drupal.org/node/2824193

PR #545, opened 2016-11-02
Issue #2755529 by ndf, drugan, olafkarsten: Product variant bulk creation

drugan:
1. Created **Drupal\commerce_product\ProductVariationBulkCreator** service.

2. Added new **Drupal\commerce_product\Plugin\Field\FieldWidget\ProductVariationSkuWidget** to make auto SKU configurable on form display widget settings.

Also, some refactoring work are done.

### DIFF
--- a/modules/product/commerce_product.module
+++ b/modules/product/commerce_product.module
@@ -259,12 +259,153 @@ function commerce_product_field_widget_form_alter(&$element, FormStateInterface 
   $entity_type = $field_definition->getTargetEntityTypeId();
   $widget_name = $context['widget']->getPluginId();
   $required = $field_definition->isRequired();
-  if ($field_name == 'path' && $entity_type == 'commerce_product' && $widget_name == 'path') {
-    $element['alias']['#description'] = t('The alternative URL for this product. Use a relative path. For example, "/my-product".');
-  }
-  elseif ($field_name == 'title' && $entity_type == 'commerce_product_variation' && !$required) {
-    // The title field is optional only when its value is automatically
-    // generated, in which case the widget needs to be hidden.
-    $element['#access'] = FALSE;
+  switch (TRUE) {
+  // The filtering only current owner stores is now moved to: https://github.com/drugan/commerce_multistore
+//     case ($field_name == 'stores'):
+//       $user = \Drupal::currentUser();
+//       // If current user has no admin permission then display only those stores
+//       // which are owned by the user.
+//       if (!$user->hasPermission('administer commerce_store')) {
+//         $stores = \Drupal::service('entity.manager')->getStorage('commerce_store')->loadMultiple();
+//         $id = $user->id();
+//         $i = 0;
+//         foreach ($stores as $store) {
+//           if ($store->getOwnerId() != $id) {
+//             $element['alter_data_' . $i] = [
+//               '#parents' => ['target_id', 'value'],
+//               $store->id() => [],
+//             ];
+//             $i++;
+//           }
+//         }
+//         if ($i) {
+//           $creator = \Drupal::service('commerce_product.variation_bulk_creator');
+//           $element['#after_build'][] = [get_class($creator), 'afterBuildPreRenderArrayAlter'];
+//         }
+//       }
+//     break;
+    case ($field_name == 'variations' && $entity_type == 'commerce_product'):
+      /** @var \Drupal\commerce_product\ProductVariationBulkCreator $creator */
+      $creator = \Drupal::service('commerce_product.variation_bulk_creator');
+      /** @var \Drupal\commerce_product\Entity\Product $product */
+      $product = $form_state->getFormObject()->getEntity();
+      /** @var \Drupal\commerce_product\Entity\ProductVariation $variation */
+      $variation = $creator->getProductVariation($product);
+      $sku_settings = $creator::getSkuwidget($variation)->getSettings();
+      $form_state->setValue('sku_settings', $sku_settings);
+      extract($sku_settings);
+      $widget_settings = $context['widget']->getSettings();
+      $one = isset($widget_settings['label_singular']) ? $widget_settings['label_singular'] : FALSE;
+      $many = $one ? $widget_settings['label_plural'] : FALSE;
+      // Disable autocompletion and autocreation if the widget is not inline
+      // entity form or if all commerce SKU settings are empty.
+      if (!$one || (empty($uniqid_enabled) && empty($prefix) && empty($suffix))) {
+        break;
+      }
+      $ief = isset($element['form']['inline_entity_form']);
+      $add = isset($element['actions']['ief_add']);
+      $add_existing = isset($element['actions']['ief_add_existing']);
+      $autocomplete = isset($element['form']['entity_id']);
+      $id = isset($element['#ief_id']) ? $element['#ief_id'] : FALSE;
+      if ($new = !isset($element['entities']['0'])) {
+        $all = $creator->getNotUsedAttributesCombination([$variation]);
+        $count = $all ? $all['count'] : 0;
+        if ($count > 1 && ($add || $add_existing || $autocomplete || $ief)) {
+          $i = 0;
+          // The disabled button is shown to reveal the bulk creator service
+          // availability and the number of variations to create.
+          $parents = ['form', 'inline_entity_form', 'actions', 'ief_add_all'];
+          $element['alter_data_' . $i] = [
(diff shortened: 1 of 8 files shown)


drugan:
On the second commit I've tried to resolve Travis CI errors.

drugan:
Well, Travis code errors are gone. Now I'm waiting for some thoughts on the PR before continuing with adjusting drupalcommerce tests to the new SKU widget introduced.

drugan:
Resolved WSOD when the SKU widget set "Hidden" on a form display settings page.

Note that in this case you still have valid SKUs assigned to variations on the background automatically.

drugan:
Variations bulk creation on **ief** form with human intervention is separated from programmatically created variations. New method **createAllProductVariations()** is added.

Now you can do something like this in your code:

```
$creator = \Drupal::service('commerce_product.variation_bulk_creator');
/** @var \Drupal\commerce_product\Entity\Product $product */
$variations = $creator->createAllProductVariations($product);
foreach($variations as $variation) {
  // Created variations optionally may be processed before saving them
  // on a product. For example, the specific image could be assigned
  // if the variation type has this type of a field.
  //code ... ... ...
  $variation->updateOriginalValues();
}
$product->setVariations($variations);
$product->isNew() ? $product->setTitle(t('My Programmatically Generated Product')) : NULL;
$product->updateOriginalValues();
$product->save();
```

The generated product could be found here: **admin/commerce/products**.

All created variations by default have price of 1 of whatever the first currency is imported to your store. You may assign your own variation fields values by passing an optional argument to the method:

```
$creator = \Drupal::service('commerce_product.variation_bulk_creator');
$my_variation_values = [
  'price' => new  \Drupal\commerce_price\Price('99.99', 'EUR'),
  'status' => 1,
  // etc..
];
$variations = $creator->createAllProductVariations($product, $my_variation_values);
$product->setVariations($variations);
$product->isNew() ? $product->setTitle(t('My Programmatically Generated Product')) : NULL;
$product->updateOriginalValues();
$product->save();
```

Note that the method also works for products which already have some variations created and saved. All the missed variations are added based on the last variation on the product.

So, if the patch will be accepted then the way to creation **SampleProductCreator** service will be clear, I think.

drugan:
One more attempt to make Travis happy is done.

modules/product/commerce_product.module
```diff
@@ -339,12 +339,66 @@ function commerce_product_field_widget_form_alter(&$element, FormStateInterface
...
   $widget_name = $context['widget']->getPluginId();
   $required = $field_definition->isRequired();
-  if ($field_name == 'path' && $entity_type == 'commerce_product' && $widget_name == 'path') {
-    $element['alias']['#description'] = t('The alternative URL for this product. Use a relative path. For example, "/my-product".');
-  }
-  elseif ($field_name == 'title' && $entity_type == 'commerce_product_variation' && !$required) {
-    // The title field is optional only when its value is automatically
-    // generated, in which case the widget needs to be hidden.
-    $element['#access'] = FALSE;
+  switch (TRUE) {
```
mglaman, line 262: This seems weird. I get it works. But feels really weird.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @FieldWidget(
+ *   id = "commerce_auto_sku",
+ *   label = @Translation("Commerce auto SKU"),
+ *   field_types = {
+ *     "string"
+ *   }
+ * )
+ */
+class ProductVariationSkuWidget extends StringTextfieldWidget {
```
mglaman: Would it be best if we put this into its own issue to deprecate the contrib module? Allows us to write better test coverage, too

modules/product/src/ProductVariationBulkCreator.php
```diff
@@ -0,0 +1,286 @@
...
+  public function createAllProductVariations(Product $product, array $variation_custom_values = []) {
+    $variations = $product->getVariations();
+    $timestamp = time();
+    if (empty($variations)) {
+      $variation_default_values = [
+        'type' => $product->getFieldDefinition('variations')->getSettings()['handler_settings']['target_bundles'][0],
+        'created' => $timestamp,
+        'changed' => $timestamp,
+      ];
+      $variation = ProductVariation::create(array_merge($variation_default_values, $variation_custom_values));
```
mglaman: Inject entityTypeManager to get the entity storage for loading

modules/product/src/ProductVariationBulkCreator.php
```diff
@@ -0,0 +1,286 @@
...
+      $sku = static::getAutoSku($variation);
+      $variation->setSku(empty($sku) ? \uniqid() : $sku);
+
+      $field = $this->getAttributeFieldOptionIds($variation);
+      foreach ($field['ids'] as $field_name => $ids) {
+        $variation->get($field_name)->setValue(['target_id' => $ids[0] == '_none' ? NULL : $ids[0]]);
+      }
+
+      if (!$variation->getPrice() instanceof Price) {
+        $currency_storage = \Drupal::service('entity_type.manager')->getStorage('commerce_currency');
```
mglaman: See previous comment, use DI

mglaman:
@drugan I think we should make an issue and work on the autosku widget there, just so that its highlighted and marks deprecation of the contrib module. And we can then write some good test coverage for it.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
drugan: @mglaman 
So, if product module is going to deprecate (?) I'll move **ProductVariationSkuWidget.php** into **src/Plugin/Field/FieldWidget** folder and open an issue "Make SKU widget configurable for autoSKU value".

Let me know if I'm wrong there.

modules/product/commerce_product.module
drugan: It's just a coding style replacing

```
if(condition) {
}
elseif(condition) {
}
```

to

```
case (condition):
break;
case (condition):
break;
```

IMO it makes code more readable when there is a lot blocks better separating them visually. If you see "break" you'd know that this is the end of a block. Not the same with braces which are small and hard to detect sometimes in a complex code snippet.

modules/product/src/ProductVariationBulkCreator.php
drugan: Good idea. I'll do it.

drugan:
1. **ProductVariationSkuWidget** moved to **src/Plugin/Field/FieldWidget**. [Issue is open](https://www.drupal.org/node/2829487). New [PR](https://github.com/drupalcommerce/commerce/pull/563) is done.
2. All the widgets settings made optional. Now you can disable uniqid() and use just a prefix/suffix or nothing at all.
3. Added **ProductVariationBulkCreator->createSampleProductVariation()** to be more flexible when  adding variations programmatically. Now you can do this:

    /** @var \Drupal\commerce_product\ProductVariationBulkCreator $creator */
    $creator = \Drupal::service('commerce_product.variation_bulk_creator');
    /** @var \Drupal\commerce_product\Entity\ProductVariation $variation */
    $variation = $creator->createSampleProductVariation($product);
    $all = $creator->getAllAttributesCombinations([$variation]);

Only one variation is created and based on that you can fetch all attributes combinations, their number, field labels and IDs. No need to save a product or variaton for this. Also, might be useful for sample products auto creation. No need to create, say a hundred or a thousand variations if you have such complex attributes. 

4. Added **ProductVariationBulkCreator::afterBuildPreRenderArrayAlter()** as otherwise I had no idea how to alter IEF widget on the new pristine product to add a disabled button. Hope it will be useful somewhere else as a callback to alter form render arrays. Also, the attributes are filled in from the very first IEF form with the help of this method.

As ever, some refactoring work are done.

drugan:
Changed logic for disabled button displayed on a new product. The button reveals the bulk creator service availability and the number of variations to create. 
The button is not rendered when:

1. The product variation type has no attributes AND/OR the number of variations to create < 2.
2. The number of already created variations > 0.
3. The "Allow users to add new variations." AND the "Allow users to add existing variations." boxes are unchecked on a product type Variations form display widget.

drugan:
On the 51bd0f8 I've made changes to be compatible with #575 and be able to generate combinations for a variation with combined attribute field(s).

bojanz:
Let's continue on the issue, now that variations are a tab and can have a "Bulk add" action (just like coupons).

PR #550, opened 2016-11-09
Issue #2824320 Checkout access

czigor:


czigor:
https://www.drupal.org/node/2824320

modules/checkout/src/Controller/CheckoutController.php
```diff
@@ -119,4 +142,98 @@ public function checkAccess(RouteMatchInterface $route_match, AccountInterface $
...
+    $checkout_flow = $this->checkoutOrderManager->getCheckoutFlow($order);
+    $requested_step = $checkout_flow->getPlugin()->getStepId();
+    $visible_steps = $checkout_flow->getPlugin()->getVisibleSteps();
+    $visible_step_ids = array_keys($visible_steps);
+    $first_step = reset($visible_step_ids);
+    $order_step = $this->getOrderCheckoutStep($order);
+
+    // If the order is not in checkout, return FALSE for any page but the
+    // completion page.
+    if ($order_step === 'complete' && $requested_step !== 'complete') {
```
bojanz: We should improve this to take into account the order workflow.
- If the order has been placed, allow the "complete" step (and only the "complete" step).
- If the order is still a draft, allow the current step and the previous ones.
Otherwise 403. That covers both cancelled and refunded.

Thoughts:
- I guess it's safe to assume that every checkout flow will have a "complete" step (but how do we enforce that?)
- We need to implement the "payment" use case somehow, where once you reach the payment step you can't go back to the previous steps, and the order in checkout is no longer considered a cart.
This is double tricky cause ideally checkout would not depend on cart (even though it currently does)

mglaman: > I guess it's safe to assume that every checkout flow will have a "complete" step (but how do we enforce that?)

It explodes when they unset it.

mglaman:
Wrapping this up.

Moving checkout access to kernel test and expanding. We can mock the request and check access via `$request = Request::create($path, 'GET');`

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
```diff
@@ -104,42 +110,71 @@ public function testOrderAccess() {
...
       'type' => 'default',
-      'state' => 'in_checkout',
+      'state' => 'draft',
       'order_number' => '6',
       'mail' => 'test@example.com',
       'uid' => $user->id(),
       'ip_address' => '127.0.0.1',
       'billing_profile' => $profile,
       'order_items' => [$order_item],
+      'cart' => TRUE,
```
steveoliver: Just need to add a ```store_id``` here also, since https://github.com/drupalcommerce/commerce/commit/f3e2093c7a2e6c4426e70145b0568c1f4ebc5d5d

steveoliver: A rebase should do the trick.

bojanz:
Finished in fc4412b8eecbfa39dc4d871a8fcd339f2f3f44f3.

PR #551, opened 2016-11-09
Issue #2824320 Checkout access - only tests, should FAIL. 

czigor:


czigor:
https://www.drupal.org/node/2824320

bojanz:
Great, let's continue in #550.

PR #553, opened 2016-11-16
Issue #2807559: Add owner filter to order/cart overview

cornifex:


modules/order/config/install/views.view.commerce_orders.yml
```diff
@@ -912,6 +977,48 @@ display:
...
+          relationship: none
+          group_type: group
+          admin_label: ''
+          operator: contains
+          value: ''
+          group: 1
+          exposed: true
+          expose:
+            operator_id: combine_op
+            label: Owner
```
bojanz: This should be "Customer" instead of "Owner"

modules/order/config/install/views.view.commerce_orders.yml
```diff
@@ -788,6 +788,71 @@ display:
           destination: true
           entity_type: commerce_order
           plugin_id: entity_operations
+        name:
```
bojanz, line 791: Why did you add this? We already have a Customer column that shows the user name.

bojanz (changes requested):

modules/order/config/install/views.view.commerce_orders.yml
cornifex: Had to add it to use in the combined filter, otherwise it would only search by UID

mglaman: @bojanz we'll only be able to search by contact mail without it. The combined filters filter requires fields in the "Fields" section.

mglaman:
This needs codestyle fix. Newline.

```
FILE: ...alcommerce/commerce/modules/order/commerce_order.post_update.php
 86 | ERROR | [x] Expected 1 newline at end of file; 0 found
    |       |     (Drupal.Files.EndFileNewline.NoneFound)
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
Time: 5.88 secs; Memory: 36Mb
```

PR #555, opened 2016-11-16
Issue #2828112: Cart listing page does not have same filters as norma…

cornifex:
…l order listing page


mglaman:
This has same user filter as in #553 added so that users can search by username and order email

PR #557, opened 2016-11-17
Issue #2828520: Create price twig filter

cornifex:


cornifex:
Added a filter that can be used on a number. This will obviously only work when the the given arg is a currency code that belongs to a configured currency.

modules/price/commerce_price.services.yml
```diff
@@ -20,3 +20,9 @@ services:
     class: Drupal\commerce_price\Resolver\DefaultPriceResolver
     tags:
       - { name: commerce_price.price_resolver, priority: -100 }
+
+  commerce_price.twig_extension:
```
mglaman, line 24: Should we do `commerce_price.twig_extension.price_format` ?

modules/price/src/TwigExtension/CommercePriceFormat.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Drupal\commerce_price\TwigExtension;
+
+use Drupal\Core\Template\TwigExtension;
+
+class CommercePriceFormat extends TwigExtension {
```
mglaman: We should do a class doc block here.

```
/**
 * Provides a `price_format` Twig filer
 * 
 * @example: 12.34|price_format('USD') would return $12.34
 */
```

modules/price/src/TwigExtension/CommercePriceFormat.php
```diff
@@ -0,0 +1,41 @@
...
+   */
+  public function getFilters() {
+    return [new \Twig_SimpleFilter('commerce_price_format', [$this, 'formatPrice'])];
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public function getName() {
+    return 'commerce_price.twig_extenstion';
```
mglaman: See naming Q in service.yml

mglaman (changes requested):
@bojanz what are your thoughts on the naming?
@cornifex want to tweak up the filter class doc

modules/price/commerce_price.services.yml
bojanz: I think it's fine to have a single twig extension per module.

modules/price/src/TwigExtension/CommercePriceFormat.php
bojanz: We also need a better name for the class. We don't prefix our class names with Commerce unless it's somehow imposed by core (like on annotation classes).  How about simply calling it TwigExtension? \Drupal\commerce_price\TwigExtension sounds descriptive enough

modules/price/src/TwigExtension/CommercePriceFormat.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Drupal\Core\Template\TwigExtension;
+
+class CommercePriceFormat extends TwigExtension {
+
+  /**
+   * @inheritdoc
+   */
+  public function getFilters() {
+    return [new \Twig_SimpleFilter('commerce_price_format', [$this, 'formatPrice'])];
```
bojanz: Let's break the new part into a separate line for readability, like core does..

modules/price/src/TwigExtension/CommercePriceFormat.php
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * @inheritdoc
+   */
+  public function getName() {
+    return 'commerce_price.twig_extenstion';
+  }
+
+  /**
+   *
```
bojanz: Extra unneeded newline in docblock. The description needs to come first.

modules/price/src/TwigExtension/CommercePriceFormat.php
```diff
@@ -0,0 +1,41 @@
...
+   * Creates a formatted price from a passed number and currency code.
+   *
+   * @param $price_number
+   * @param $price_currency_code
+   *
+   * @return mixed
+   */
+  public static function formatPrice($price_number, $price_currency_code) {
+    $currency_storage = \Drupal::entityTypeManager()->getStorage('commerce_currency');
```
bojanz: You're in a service. Inject these two dependencies.

modules/price/src/TwigExtension/CommercePriceFormat.php
cornifex: Seems like it might be kinda silly to call it TwigExtension, since it extends a class with that name. What about just dropping "Commerce" and calling it PriceFormat, or something similar?

modules/price/src/TwigExtension/CommercePriceFormat.php
cornifex: Will dependency injection work if the method is static?

mglaman (approved):

modules/price/commerce_price.services.yml
mglaman: Yeah, I forgot the service returns multiple things.

mglaman (changes requested):
We need a test.
- Create a .module for `commerce_test` https://github.com/drupalcommerce/commerce/tree/8.x-2.x/tests/modules/commerce_test
- Add two theme hooks with matching twig. 
  - One theme hook renders a Price object properly
  - One does w/ bad array, throws exception
- Write a kernel test which renders the theme hooks, verifies exception or content.

tests/src/Kernel/PriceTwigExtensionTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Tests an improperly formatted price array.
+   */
+  public function testBrokenPrice() {
+    $theme = ['#theme' => 'broken_commerce_price'];
+    try {
+      $this->render($theme);
+    }
+    catch (\InvalidArgumentException $e) {
```
bojanz: Don't do try / catch in tests. Instead, use $this->setExpectedException(\InvalidArgumentException::class); before the render call.

tests/src/Kernel/PriceTwigExtensionTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Load in USD currency.
+   */
+  public function setUp() {
+    parent::setUp();
+    $this->container->get('commerce_price.currency_importer')->import('USD');
+  }
+
+  public static $modules = [
```
bojanz: $modules needs to be defined above any methods, and needs to have a docblock like in other tests.

tests/modules/commerce_test/commerce_test.module
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+use Drupal\commerce_price\Price;
+
+/**
+ * Implements hook_theme().
+ */
+function commerce_test_theme($existing, $type, $theme, $path) {
+  return [
+    'working_commerce_price' => [
```
bojanz: You can use the 'template' key to reuse the same template for multiple theme implementations.

modules/price/src/TwigExtension/PriceTwigExtension.php
```diff
@@ -0,0 +1,62 @@
...
+   * @param mixed $price
+   *   Either a Price object, or an array with number and currency_code keys.
+   *
+   * @return mixed
+   *   A formatted price, suitable for rendering in a twig template.
+   *
+   * @throws \InvalidArgumentException
+   */
+  public static function formatPrice($price) {
+
```
bojanz: Unneeded newline at the beginning of method.

modules/price/src/TwigExtension/PriceTwigExtension.php
```diff
@@ -0,0 +1,62 @@
...
+use Drupal\commerce_price\Price;
+use Drupal\Core\Template\TwigExtension;
+
+/**
+ * Provides a twig filter to be applied to a price object or array.
+ *
+ * Example: {{ order.getTotalPrice|commerce_price_format }}
+ */
+class PriceTwigExtension extends TwigExtension {
```
bojanz: should be \TwigExtension, since it's in the root namespace (no use statement)

modules/price/src/TwigExtension/PriceTwigExtension.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Drupal\commerce_price\TwigExtension;
+
+use Drupal\commerce_price\Price;
+use Drupal\Core\Template\TwigExtension;
+
+/**
+ * Provides a twig filter to be applied to a price object or array.
```
bojanz: Leave room for future filters by only documenting something like "Provides Price-specific Twig extensions."
The code example can be moved to the formatPrice docblock.

modules/price/src/TwigExtension/PriceTwigExtension.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Drupal\commerce_price\TwigExtension;
+
+use Drupal\commerce_price\Price;
+use Drupal\Core\Template\TwigExtension;
```
bojanz: There's no need to extend the core TwigExtension, extend the Twig one instead (\TwigExtension).

mglaman:
@bojanz looks good to me, and we have three tests: price object, array, malformed array.

modules/price/commerce_price.services.yml
```diff
@@ -20,3 +20,9 @@ services:
...
+      arguments: ['@renderer']
```
bojanz: No need to inject the render once we stop inheriting from the core twig extension class.

bojanz (changes requested):

mglaman (approved):

PR #558, opened 2016-11-17
adding explicit NULL return statements to chain resolvers

agoradesign:
found this by accident. It's a little code style improvement. All chain resolvers just exit their resolve() functions, if no resolution was achieved. This PR is adding explicit NULL returns.


bojanz:
I don't see this as an improvement. Any non-explicit return is a NULL.

PR #559, opened 2016-11-17
Issue #2827721 - AddToCartForm should support all purchasable entities

steveoliver:
See https://www.drupal.org/node/2827721

### DIFF
--- a/modules/cart/src/Form/AddToCartForm.php
+++ b/modules/cart/src/Form/AddToCartForm.php
@@ -105,7 +105,9 @@ class AddToCartForm extends ContentEntityForm {
    * {@inheritdoc}
    */
   public function getFormId() {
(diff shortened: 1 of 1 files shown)


mglaman:
Looks good, here. We have test coverage in MultipleCartFormsTest.

modules/cart/src/Form/AddToCartForm.php
```diff
@@ -105,15 +105,17 @@ public function getBaseFormId() {
    * {@inheritdoc}
    */
   public function getFormId() {
-    $product_id = $this->entity->getPurchasedEntity()->getProductId();
+    static $form_id_suffix = 0;
```
bojanz: Wouldn't this make more sense as a static class property? Class methods generally don't use the static keyword.

PR #563, opened 2016-11-21
Issue #2829487 by drugan: Make configurable auto SKU widget

drugan:
As it were suggested by mglaman in [#545](https://github.com/drupalcommerce/commerce/pull/545) the auto SKU widget is moved to core and [issue #2829487 ](https://www.drupal.org/node/2829487) is open.

### DIFF
--- a/config/schema/commerce.schema.yml
+++ b/config/schema/commerce.schema.yml
@@ -33,6 +33,29 @@ field.field_settings.commerce_plugin_item:*:
       sequence:
         - type: string
 
+field.widget.settings.commerce_auto_sku:
+  type: mapping
+  label: 'Commerce auto SKU format settings'
+  mapping:
+    iniqid_enabled:
+      type: boolean
+      label: 'Enable unique auto SKU values generation'
+    more_entropy:
+      type: boolean
+      label: 'More unique'
+    prefix:
+      type: string
+      label: 'SKU prefix'
+    suffix:
+      type: string
+      label: 'SKU suffix'
+    size:
+      type: integer
+      label: 'Size of textfield'
+    placeholder:
+      type: label
+      label: 'Placeholder'
+
 views.argument_validator.commerce_current_user:
   type: mapping
   label: 'Current user'
--- a/config/schema/commerce.schema.yml
+++ b/config/schema/commerce.schema.yml
@@ -37,7 +37,7 @@ field.widget.settings.commerce_auto_sku:
   type: mapping
   label: 'Commerce auto SKU format settings'
   mapping:
-    iniqid_enabled:
+    uniqid_enabled:
       type: boolean
       label: 'Enable unique auto SKU values generation'
     more_entropy:
--- a/config/schema/commerce.schema.yml
+++ b/config/schema/commerce.schema.yml
@@ -51,7 +51,7 @@ field.widget.settings.commerce_auto_sku:
       label: 'SKU suffix'
     size:
       type: integer
-      label: 'Size of textfield'
+      label: 'Size of SKU field'
     placeholder:
       type: label
       label: 'Placeholder'
--- a/src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
+++ b/src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
@@ -90,7 +90,7 @@ class ProductVariationSkuWidget extends StringTextfieldWidget {
     $summary = [];
     $none = $this->t('None');
     $settings = $this->getSettings();
-    $sku =  uniqid($settings['prefix'], $settings['more_entropy']) . $settings['suffix'];
+    $sku = uniqid($settings['prefix'], $settings['more_entropy']) . $settings['suffix'];
     $settings['auto SKU sample'] = $settings['uniqid_enabled'] ? $sku : $none;
     unset($settings['uniqid_enabled'], $settings['more_entropy']);
     foreach ($settings as $name => $value) {
--- a/src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
+++ b/src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Drupal\commerce\Plugin\Field\FieldWidget;
+
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Field\Plugin\Field\FieldWidget\StringTextfieldWidget;
+
+/**
+ * Plugin implementation of the 'commerce_auto_sku' widget.
+ *
+ * @FieldWidget(
+ *   id = "commerce_auto_sku",
+ *   label = @Translation("Commerce auto SKU"),
+ *   field_types = {
+ *     "string"
+ *   }
+ * )
+ */
+class ProductVariationSkuWidget extends StringTextfieldWidget {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
+    return array(
+      'iniqid_enabled' => TRUE,
+      'more_entropy' => FALSE,
+      'prefix' => 'sku-',
+      'suffix' => '',
+      'size' => 60,
+      'placeholder' => '',
+    ) + parent::defaultSettings();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $none = t('None');
+    $settings = $this->getSettings();
+    $element['iniqid_enabled'] = array(
+      '#type' => 'checkbox',
+      '#title' => t('Enable unique auto SKU values generation'),
+      '#default_value' => $settings['iniqid_enabled'],
+    );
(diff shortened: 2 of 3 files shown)


config/schema/commerce.schema.yml
```diff
@@ -33,6 +33,29 @@ field.field_settings.commerce_plugin_item:*:
       sequence:
         - type: string
 
+field.widget.settings.commerce_auto_sku:
+  type: mapping
+  label: 'Commerce auto SKU format settings'
+  mapping:
+    iniqid_enabled:
```
steveoliver: It looks like this should be 'uniqid', instead of 'iniqid'

config/schema/commerce.schema.yml
```diff
@@ -33,6 +33,29 @@ field.field_settings.commerce_plugin_item:*:
...
+      label: 'More unique'
+    prefix:
+      type: string
+      label: 'SKU prefix'
+    suffix:
+      type: string
+      label: 'SKU suffix'
+    size:
+      type: integer
+      label: 'Size of textfield'
```
steveoliver: Is this the size of the textfield or the size of the sku?

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
```diff
@@ -0,0 +1,119 @@
...
+      '#placeholder' => $none,
+    );
+    $element['size'] = array(
+      '#type' => 'number',
+      '#title' => t('Size of textfield'),
+      '#default_value' => $settings['size'],
+      '#required' => TRUE,
+      '#min' => 1,
+    );
+    $element['placeholder'] = array(
```
steveoliver: I don't see how a placeholder will be useful for this widget. The user should not be entering text should they?

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsSummary() {
+    $summary = array();
+    $none = t('None');
+    $settings = $this->getSettings();
+    $settings['Uniqueness sample'] = $settings['iniqid_enabled'] ? ($settings['more_entropy'] ? uniqid('', TRUE) : uniqid()) : $none;
```
steveoliver: I think the prefix and suffix should be used here.

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
```diff
@@ -0,0 +1,119 @@
...
+    return $summary;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+    $element['value'] = $element + array(
+      '#type' => 'textfield',
+      '#default_value' => isset($items[$delta]->value) ? $items[$delta]->value : NULL,
```
steveoliver: Isn't this where the id should be generated?

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
```diff
@@ -0,0 +1,119 @@
...
+ *   }
+ * )
+ */
+class ProductVariationSkuWidget extends StringTextfieldWidget {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
+    return array(
```
steveoliver: Use shorthand array syntax throughout.

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
```diff
@@ -0,0 +1,119 @@
...
+      'size' => 60,
+      'placeholder' => '',
+    ) + parent::defaultSettings();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $none = t('None');
```
steveoliver: You can use $this->t() instead of t().

config/schema/commerce.schema.yml
drugan: Right, thank you. I've changed it.

config/schema/commerce.schema.yml
drugan: Thanks again!

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
drugan: What I think the array shortcut syntax need to be used carefully in the code. Obviously, it is more appropriate in the places like this:

`public function myFunc(array $my_array = []) {}`

But it looks like more readable in a multiline array declaration as in the snippet above. Just because the word "array" is clearly highlighted whereas brackets not.

The same says [Drupal coding standard](https://www.drupal.org/docs/develop/standards/coding-standards#array).  [Discussion](https://www.drupal.org/node/2135291) will never stop on this issue.

steveoliver: Yep. However, in Commerce you'll find 0 instances of the old style array syntax.

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
drugan: As I understand $this->t('None') is a shortcut to this:

`t('None', [], ['context' => get_class($this)]);`

What does it mean "None" in the first place takes context translations and if they are not available then it takes global translations. Not sure with english word "None" because I translate it to my own language and for me it looks like as a ***global*** notation.

**UPD**: ok, I agree - it needs $this->t().

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
drugan: Yes, they might. If the uniqid, prefix and suffix is disabled then user will see a placeholder. Also, user sees the placeholder when auto generated SKU is removed from the field.

There are a lot of things which seems to us as "useless" but really we don't know what a user needs might be and may be some day someone will post an issue: "Please, add placeholder to the SKU widget!".

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
drugan: That's a good point. I'll do it.

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
drugan: Thank you, that is a good point too. I need thoroughly to think about it.

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
steveoliver: Rather, t() is a global function that instantiates a new TranslatableMarkup object. But, being a global function, is no longer preferred as "All t() usage in non-static methods in classes is suspect and should be removed." ([ref](https://www.drupal.org/node/2072251#comment-11236863)), especially when the Widget base class already uses the StringTranslationTrait.  This is better for unit testability also.

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
steveoliver: Ah, I see what you're saying.

src/Plugin/Field/FieldWidget/ProductVariationSkuWidget.php
drugan: OK, as to be in sync with commerce coding standard I need to replace this old syntax form all my patches.

drugan:
Made some changes suggested by steveoliver:

1. t() changed for $this->t()
2. array() for []
3. Added prefix and suffix to the SKU sample in settingsSummary()

@steveoliver 

As for you suggestion to implement uniqid() in formElement() for #default_value I think this not what we'd want.
Reasons:

1. It applies unnecessary restrictions on the field value. No other code could change this except theme handlers but changing data is not their concern.
2. It makes difficult to access auto SKU when you want it in programmatically generated entities, for example in tests.

I have idea for the future to make **SampleProductCreator** service so it will be better to implement **ProductVariationSkuWidget** settings in some additional service as it done Product variant bulk creation PR.
See there modules/product/src/Entity/ProductVariation.php +395:

```
    $fields['sku'] = BaseFieldDefinition::create('string')
      // .....
       ->setDefaultValueCallback('Drupal\commerce_product\ProductVariationBulkCreator::getAutoSku')
       //....
       ])
       ->setDisplayOptions('form', [
       'type' => 'commerce_auto_sku',
        'weight' => -4,
    ])
```

If some other module wants to override this callback they can do something like this:

`    $variation->get('sku')->getFieldDefinition()->setDefaultValueCallback('Drupal\my_module\MySku::getAutoSku');`

If your suggestion about #default_value would be applied this will impossible.

bojanz:
We said we'd let this functionality mature in the commerce_autosku contrib for now. Let's revisit after 2.0

PR #565, opened 2016-11-23
2757283 template admin order view

cornifex:
This is just a rough template using all twig with no preprocessing. Definitely needs more work but I've been given other priorities for the time being.

### DIFF
--- a/modules/order/commerce_order.module
+++ b/modules/order/commerce_order.module
@@ -26,6 +26,10 @@ function commerce_order_theme($existing, $type, $theme, $path) {
     'commerce_order_edit_form' => [
       'render element' => 'form',
     ],
+    'commerce_order__default' => [
+      'base hook' => 'commerce_order',
+      'render element' => 'elements',
+    ],
   ];
 }
 
@@ -79,6 +83,11 @@ function template_preprocess_commerce_order(&$variables) {
   $variables['order'] = [];
   foreach (Element::children($variables['elements']) as $key) {
     $variables['order'][$key] = $variables['elements'][$key];
+  }
+
+  if ($order->getBillingProfile()) {
+    $profile_view_bulder = \Drupal::entityTypeManager()->getViewBuilder('profile');
+    $variables['order']['billing_information'] = $profile_view_bulder->view($order->getBillingProfile());
   }
 }
 
--- a/modules/order/templates/commerce-order--default.html.twig
+++ b/modules/order/templates/commerce-order--default.html.twig
@@ -54,23 +54,23 @@
           {% endfor %}
         </div>
       </details>
-      {% if order['billing_information'] %}
+      {% if order.billing_information %}
         <details open>
           <summary role="button">
             {{ 'Billing information'|t }}
           </summary>
           <div class="details-wrapper">
-            {{ order['billing_information'] }}
+            {{ order.billing_information }}
           </div>
         </details>
       {% endif %}
-      {% if order['shipping_information'] %}
+      {% if order.shipping_information %}
         <details open>
           <summary role="button">
             {{ 'Shipping information'|t }}
           </summary>
           <div class="details-wrapper">
-            {{ order['shipping_information'] }}
+            {{ order.shipping_information }}
           </div>
         </details>
       {% endif %}
--- a/modules/order/templates/commerce-order--default.html.twig
+++ b/modules/order/templates/commerce-order--default.html.twig
@@ -1,0 +1,85 @@
+{#
+/**
+ * @file
+ * Order template used on the admin order page.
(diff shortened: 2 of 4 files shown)


modules/order/commerce_order.module
```diff
@@ -26,6 +26,10 @@ function commerce_order_theme($existing, $type, $theme, $path) {
     'commerce_order_edit_form' => [
       'render element' => 'form',
     ],
+    'commerce_order__default' => [
```
mglaman, line 29: Discussed w/ @cornifex we'd make an admin view mode and use that instead

mglaman: Looks like we'd have to alter the default canonical route, Core provides

```
'_entity_view' => "{$entity_type_id}.full",
```

mglaman: Discussed w/ Bojan. Doesn't really give us a gain. And tests showed all kinds of madness by doing this.

modules/order/templates/commerce-order--default.html.twig
```diff
@@ -0,0 +1,77 @@
...
+          {% for key in ['uid', 'mail', 'ip_address'] %}
+            {% if order[key] %}
+              <div class="form-item">
+                {{ order[key] }}
+              </div>
+            {% endif %}
+          {% endfor %}
+        </div>
+      </details>
+      {% for key,value in {'billing_information': 'Billing information'|t, 'shipping_information': 'Shipping information'|t} %}
```
mglaman: @bojanz I followed the same concepts from the email, since this is pretty hardcoded-ish.

mglaman:
@bojanz this is ready to take a peak at.

modules/order/templates/commerce-order--default.html.twig
```diff
@@ -0,0 +1,77 @@
...
+          {% if order[key] %}
+            <div class="form-item">
+              {{ order[key] }}
+            </div>
+          {% endif %}
+        {% endfor %}
+      </div>
+      <details open>
+        <summary role="button">
+          Customer Information
```
bojanz: Missing |t for Customer information.

modules/order/templates/commerce-order--default.html.twig
```diff
@@ -0,0 +1,77 @@
...
+ *   {{ order|without('order_number') }}
+ *   @endcode
+ * - order_entity: The order entity.
+ *
+ * @ingroup themeable
+ */
+#}
+
+{{ attach_library('commerce_order/form') }}
+{% set order_state = order_entity.getState.value|capitalize %}
```
bojanz: You're capitalizing the machine name of the state, instead of taking the label. That's not right.

modules/order/commerce_order.module
```diff
@@ -80,6 +84,13 @@ function template_preprocess_commerce_order(&$variables) {
   foreach (Element::children($variables['elements']) as $key) {
     $variables['order'][$key] = $variables['elements'][$key];
   }
+
+  if ($order->getBillingProfile()) {
+    $variables['order']['billing_information'] = $order->getBillingProfile()->get('address')->view([
```
bojanz: Why would we not render the entire billing profile? Doing it this way means no phone field / VAT number is shown even if the fields are attached.

modules/order/templates/commerce-order--default.html.twig
bojanz: It feels like it would be better for TX (themer experience) if these for loops were unrolled.
We have two elements here, repeating their HTML twice is probably no big deal?

modules/order/templates/commerce-order--default.html.twig
```diff
@@ -0,0 +1,77 @@
+{#
+/**
+ * @file
+ * Default order template.
```
bojanz: Let's add something like "Used on the admin order page." so that the themers know which template they need.

bojanz (changes requested):

modules/order/templates/commerce-order--default.html.twig
mglaman: Ah so we need `getState.getLabel`?

modules/order/commerce_order.module
mglaman: That works, get can just render it full

modules/order/templates/commerce-order--default.html.twig
mglaman: Works for me

mglaman:
Captain @bojanz, your needs have been addressed

bojanz (approved):

bojanz:
Continued in #574.

PR #566, opened 2016-11-24
Add support for manually created attribute fields

artemvd:
Use the attribute names from the field_map. Fixes fatal error for custom product types, because $attribute_type object in line 240 is always NULL, because the name in load method is always wrong

### DIFF
--- a/modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
+++ b/modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
@@ -232,10 +232,11 @@ class ProductVariationAttributesWidget extends ProductVariationWidgetBase implem
     $field_definitions = $this->attributeFieldManager->getFieldDefinitions($selected_variation->bundle());
     $field_map = $this->attributeFieldManager->getFieldMap($selected_variation->bundle());
     $field_names = array_column($field_map, 'field_name');
+    $attribute_names = array_column($field_map, 'attribute_id');
     $index = 0;
     foreach ($field_names as $field_name) {
       /** @var \Drupal\commerce_product\Entity\ProductAttributeInterface $attribute_type */
-      $attribute_type = $this->attributeStorage->load(substr($field_name, 10));
+      $attribute_type = $this->attributeStorage->load($attribute_names[$index]);
       $field = $field_definitions[$field_name];
       $attributes[$field_name] = [
         'field_name' => $field_name,


artemvd:
Also when you try to add the product attribute to product variation type as an entity reference field this patch also provide the solution to avoid fatal error

steveoliver:
Thanks, @artemvd - Is there a failing test we can add in a separate issue that demonstrates the problem this patch solves? I can help write the test if you need help - just add detailed steps to reproduce both cases.

artemvd:
@steveoliver thanks for the reply. I'm not so good in writing the tests, but I definitely want to learn that, so if you would help I wil appreciate this. 

Steps to reproduce:
1. Create new product variation type (not default)
2. Create new product type
3. Create product attribute (if there is no defined yet)
4. Add product attribute to product variation type.
5. Populate product fields, create the product.
6. Go to product view page
7. Observe fatal error

mglaman:
Two cents: does the field map thing have a cache, and we need to clear it
when adding a new one?

artemvd:
The other case is when you try to add Entity reference (product attribute) field to product variation. I know this is not the usual behaviour, but then it should be also not possible to do that then.

artemvd:
@mglaman I believe it is cached, but it also worked well for default product variation type

mglaman:
@artemvd oh, so this comes down to the fact that the field might not be named `attribute_`

artemvd:
As it is not a usual field, you can't see it in the field UI in any list, I think it is ok to name it with attribute_ , the only thing that substr($var, 10) - is too flawfull. Why not use the atribute_id that is already in the field_map. I believe this substr($var, 10) was made only to extract the value that is already available without any coding to be involved. This patch will just use the bulletproof reference to attribute id instead of shaky substr solution

drugan:
I've made all the steps suggested and have no any bugs. The product with assigned attributes is created and could be seen and added to the cart.

To be more precise:

1. Created new product variation type (no any attributes checked in).
2.  Created new product type with this variation type.
3.  Created **new** product attribute.
4.  Added product attribute to the product variation type (no matter is it new or existing one or both).
5.  Populated product fields, created the product.
6. Redirected to a product view page, added variation to cart.
7. Opened the cart and see added variation as expected.

@artemvd  Why? Is it some weird configuration of installed commerce?

artemvd:
Maybe because I tried to add the attribute as an entity reference field first in field UI. It was not obvious that attributes list in product variation edit page. And the caches were not cleaned.

But also anyway I don't understand why to use substr($var, 10) instead of attribute_id that is already available there in field_map ?

artemvd:
Also I suggest to show the user somehow that attributes are selected in product variation type edit form and not in the fields list and field settings like it was in commerce 1.x

drugan:
@artemvd could you post here verbose list of steps (with the attribute as an entity reference field) so I'd try to reproduce it in my environment.

artemvd:
@drugan here is the steps that caused an error for me
1. Create new product variation type (no any attributes checked in).
2. Create new product type with this variation type.
3. Create new product attribute.
4. Add product attribute to the product variation type as an entity reference field through Field List UI tab 4a. Observe that the attribute field name now is "field_attribute_name" (this comes form field UI) so one can set any name possible.
5. Populate product fields, create the product.
6. Navigate to a product view page.
7. Observe fatal error $attribute_type is undefined.

drugan:
I confirm the bug with the latest list of steps to reproduce.

`Fatal error: Call to a member function getElementType() on a non-object in /home/uladmin/public_html/astep/web/modules/contrib/commerce/modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php on line 244`

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -232,10 +232,11 @@ protected function getAttributeInfo(ProductVariationInterface $selected_variatio
     $field_definitions = $this->attributeFieldManager->getFieldDefinitions($selected_variation->bundle());
     $field_map = $this->attributeFieldManager->getFieldMap($selected_variation->bundle());
     $field_names = array_column($field_map, 'field_name');
+    $attribute_names = array_column($field_map, 'attribute_id');
     $index = 0;
     foreach ($field_names as $field_name) {
       /** @var \Drupal\commerce_product\Entity\ProductAttributeInterface $attribute_type */
-      $attribute_type = $this->attributeStorage->load(substr($field_name, 10));
+      $attribute_type = $this->attributeStorage->load($attribute_names[$index]);
```
drugan, line 239: I agree substr() here is a weird thing but could it be better to have this instead:

`$attribute_type = $this->attributeStorage->load($field_map[$index]['attribute_id']);`

That way we have no additional lines and variables introduced.

mglaman: @artemvd want to give this a try and see if all works?

I agree we can skip the magical `substr` if this works.

drugan:
Further exploring this edge use case I've discovered some mess up when you add product attribute as a reference field to a variation type (no one regular attributes checkboxes is checked) and choose to reference some existing attribute field. Let's say you've decided to choose a reference to existing "Colors" attribute field and in the process of field adding you'd suggested to choose a bundle(s) for this reference and you've choosen "Sizes" instead of "Colors". That way you can add only default variation to a product although the autocomplete widget on the product/add form as expected suggests references to "Sizes" options.

Obviously the logic for variation type constructing need to be improved but I don't think that there is some urgency on it. Edge use cases only for advanced users therefore the issues appearing have to be resolved by those users. A regular user have pretty simple and clear algorithm for a variation type constructing.

bojanz:
Avoiding substr() in the changed line is a good idea, so I support committing a variant of this fix.

However, the real problem here is that Commerce doesn't actually support creating an attribute field manually. Various parts of code assume the "attribute_" prefix which is only present if Commerce itself created the field.
Personally I am inclined to not support this use case at all (and make the attribute field manager ignore any fields that don't start with "attribute_").

If we want to fix the problem fully, then we'd also need to fix these two methods, by making them use the attribute field map:
- ProductViewBuilder::alterBuild
- ProductAttributesOverview::getAttributeItemList

We would also need to make sure $attribute_field_manager->clearCaches(); is called whenever a qualified entity reference is created. We'd also need a test.

@mglaman Thoughts?

mglaman:
Okay, so @bojanz is right on his items for final fixes.

I think this actually validates the following issue: [#2817905 When creating a new attribute, allow adding to existing variation types](https://www.drupal.org/node/2817905).

@artemvd would you say having the "attach to X types" on the create form would have mitigated this whole ordeal?

mglaman:
In regards to my last comment, see #570.

artemvd:
@mglaman this new feature will definitely make the life easier 👍

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
artemvd: @drugan change makes sense. I agree, that there is no need for the new variable, but then we maybe need to make sure that the $field_map[$index]['attribute_id'] is set, just in case (I believe that this will never happen), just to be 100% sure

mglaman:
Let's take this to #740 and combine with some pieces of #575 to handle custom attribute fields.

I have credited @artemvd in that D.o issue.

PR #567, opened 2016-11-25
In case resolver returns other currency

artemvd:
In the case if resolver returns the price in different currency we need this currency object to be format in correctly. By default $items variable has only the default currency code inside.

### DIFF
--- a/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
+++ b/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
@@ -85,7 +85,7 @@ class PriceCalculatedFormatter extends PriceDefaultFormatter implements Containe
       $container->get('entity_type.manager'),
       $container->get('commerce_price.number_formatter_factory'),
       $container->get('commerce_price.chain_price_resolver'),
-      $container->get('entity.manager')->getStorage('commerce_currency'),
+      $container->get('entity.manager')->getStorage('commerce_currency')
     );
   }
 
--- a/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
+++ b/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
@@ -84,11 +84,7 @@ class PriceCalculatedFormatter extends PriceDefaultFormatter implements Containe
    * {@inheritdoc}
    */
   public function viewElements(FieldItemListInterface $items, $langcode) {
-    $currency_codes = [];
-    foreach ($items as $delta => $item) {
-      $currency_codes[] = $item->currency_code;
-    }
-    $currencies = $this->currencyStorage->loadMultiple($currency_codes);
+    $currencies = $this->currencyStorage->loadMultiple();
 
     $elements = [];
     /** @var \Drupal\commerce_price\Plugin\Field\FieldType\PriceItem $item */
--- a/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
+++ b/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
@@ -7,7 +7,6 @@ use Drupal\commerce_price\NumberFormatterFactoryInterface;
 use Drupal\commerce_price\Resolver\ChainPriceResolverInterface;
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
-use Drupal\Core\Entity\EntityStorageInterface;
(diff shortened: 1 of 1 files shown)


modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
```diff
@@ -84,11 +84,7 @@ public static function create(ContainerInterface $container, array $configuratio
    * {@inheritdoc}
    */
   public function viewElements(FieldItemListInterface $items, $langcode) {
-    $currency_codes = [];
-    foreach ($items as $delta => $item) {
-      $currency_codes[] = $item->currency_code;
-    }
-    $currencies = $this->currencyStorage->loadMultiple($currency_codes);
+    $currencies = $this->currencyStorage->loadMultiple();
```
mglaman: This would cause all currencies to be loaded. Which generally should only be a few. But perhaps it'd be best to just move this to a single load call for the specific currency inside the `foreach` loop?

artemvd: I suggest to remove foreach at all. The problem is that in $item objects there is the currency code that is set for the commerce_price field. But if in the resolver I want to show the price in another currency? I tested this for the shop with multicurrency. I don't have price conversion and I created another price field in different currency. My resolver takes the price from the newly created field, but the price has other currency code. 

So when you collect currency codes in the loop as it is now, you collect the currency code that is set for the default field. And if the field is not multivalued (and it is not) $items will consists only of 1 item with 1 currency_code. So there is no way that formatter would show the price from the resolver in any other currency than in the currency that is set in the default price field.

mglaman: Yeah, I agree most cases it'll be 1 field with 1 currency. But we need to support any use cases :)

I see what you're saying, we shouldn't load the currency from the field value, but rather from

```
      $resolved_price = $this->chainPriceResolver->resolve($purchasable_entity, '1');
      $number = $resolved_price->getNumber();
+     $currency = $this->currencyStorage->load($number->getCurrencyCode());
```

bojanz: I agree with Matt. We also need a test for multicurrency, to demonstrate that the use case is working with the fix.

artemvd:
I rerolled the patch with @mglaman suggestions

modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
```diff
@@ -32,7 +33,13 @@ class PriceCalculatedFormatter extends PriceDefaultFormatter implements Containe
    * @var \Drupal\commerce_price\Resolver\ChainPriceResolverInterface
    */
   protected $chainPriceResolver;
-
+  
+  /**
+   *
```
bojanz, line 41: Missing single line description.

modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
```diff
@@ -76,28 +84,23 @@ public static function create(ContainerInterface $container, array $configuratio
       $configuration['third_party_settings'],
       $container->get('entity_type.manager'),
       $container->get('commerce_price.number_formatter_factory'),
-      $container->get('commerce_price.chain_price_resolver')
+      $container->get('commerce_price.chain_price_resolver'),
+      $container->get('entity.manager')->getStorage('commerce_currency')
```
bojanz: entity manager is deprecated, you need the entity type manager.
We also prefer to inject the entity type manager and then get the storage in __construct()

bojanz (changes requested):

modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
mglaman: Which is already injected, by the way @artemvd

artemvd:
@bojanz @mglaman thanks for the reviews, I update the request.

bojanz:
@artemvd Thanks! What's your drupal.org username? Opened a matching issue https://www.drupal.org/node/2831839 wanted to credit you there before merging this

artemvd:
My drupal profile name is a.dmitriiev .

artemvd:
@bojanz do you know why the changed file has the conflict now?

bojanz:
Committed in 3646c5fa53b66ac25792e50331248977ee676b21. Made a few tweaks. Thanks!

PR #568, opened 2016-11-26
Issue 282622: Ensure that the offer type target entity matches the incoming entity.

mitrpaka:
Test added for offer target entity type.


mglaman:
@mitrpaka thanks! committed in @5c7ffacc906a10b63d74461d119bfdd5d7b4345f

PR #570, opened 2016-11-27
Issue #2817905: When creating a new attribute, allow adding to existi…

mglaman:
…ng variation types


bojanz:
Added a test and committed as d247695abbd4be759ed72c62bdebd1a1523d34c1.

PR #573, opened 2016-11-29
Refactor commerce_cart_commerce_order_presave() into a transition event subscriber

cornifex:


modules/cart/src/EventSubscriber/CartEventSubscriber.php
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Finalizes cart.
+   *
+   * Reacts to order "place" transition event.
+   *
+   * @param \Drupal\state_machine\Event\WorkflowTransitionEvent $event
+   *   The workflow transition event.
+   */
+  public function finalizeCart(WorkflowTransitionEvent $event) {
+    $order = $event->getEntity();
```
bojanz: We no longer check $order->cart->value == TRUE, which means this runs needlessly on all placed orders, even non-cart ones?

modules/cart/src/EventSubscriber/CartEventSubscriber.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Drupal\commerce_cart\EventSubscriber;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Drupal\commerce_cart\CartProviderInterface;
+use Drupal\state_machine\Event\WorkflowTransitionEvent;
+
+/**
+ * Event reactions for commerce_cart.
```
bojanz: "Reacts to order events."

bojanz (changes requested):

bojanz:
Committed in f04be4d0df109e0eaf9367471a627e544ad684e0.

PR #575, opened 2016-11-30
Issue #2831739 by drugan: Allow combined attributes

drugan:
Thanks to great work of @mglaman and @bojanz on #570 now we can allow to combine any with any attributes on a variation type. Read [more](https://www.drupal.org/node/2831739) on the issue page.


bojanz:
See https://www.drupal.org/node/2831739#comment-12111797

PR #578, opened 2016-11-30
Issue #2831895 - AddToCartForm should support multiple instances

steveoliver:
Follow-up to #559 (https://www.drupal.org/node/2827721) - fixes form id to use purchasable entity id instead of static counter which didn't work.
See this issue on d.o: https://www.drupal.org/node/2831895

### DIFF
--- a/modules/cart/src/Form/AddToCartForm.php
+++ b/modules/cart/src/Form/AddToCartForm.php
@@ -88,7 +88,6 @@ class AddToCartForm extends ContentEntityForm {
     $this->storeContext = $store_context;
     $this->chainPriceResolver = $chain_price_resolver;
 
-    self::$formInstanceId++;
   }
 
   /**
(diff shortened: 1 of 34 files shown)


steveoliver:
currently submits add to cart forms to purchasable entities' pages. Not ideal as it directs you to another page and cannot continue adding products to cart from the view. :(

drugan:
@steveoliver as for your [comment](https://www.drupal.org/node/2831895#comment-11805221) I'd recommend apply [PR#575](https://patch-diff.githubusercontent.com/raw/drupalcommerce/commerce/pull/575.diff) and see if that works.

drugan:
Would it have sense to remove also $formInstanceId property declaration?

drugan:
@steveoliver  Please, check [this patch](https://www.drupal.org/node/2827721#comment-11812288) if that will work.

mglaman:
This ticket was closed as a duplicate for #582

PR #574, opened 2016-11-29
Issue #2757283 by mglaman, cornifex: Re-template the admin and user order views

bojanz:

PR #580, opened 2016-12-06
Issue #2832595 - Payments tab on orders no longer shows up

steveoliver:
Follow-up to #574.
See https://www.drupal.org/node/2832595.


bojanz:
Committed in eaa3be0e0815bd858a4a37392fb5da74b1476733.

ahirdev:
@bojanz 
No it's still not showing. If we open the link directly we can get access but the Payments tab is still not showing.
Drupal Version: 9.X.X

ganeshdc:
@bojanz @ahirdev 
I am having the same issue. After I update commerce module to 2.33, tab is missing, but can access that page using URL.

PR #581, opened 2016-12-10
Add an API for offsite payments

mglaman:
Experimental WIP for off-site.

2v5mm8a


mglaman:
Next step is to add the cancel/return webhooks.

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
```diff
@@ -156,7 +174,7 @@ public function getSteps() {
     return [
       'payment' => [
         'label' => $this->t('Payment'),
-        'next_label' => $this->t('Pay and complete purchase'),
```
bojanz, line 159: Didn't we say that "Pay and complete purchase" makes sense for both offsite and onsite? I find it more descriptive than just "Complete purchase"

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
```diff
@@ -374,9 +392,11 @@ protected function actions(array $form, FormStateInterface $form_state) {
     // Hide the actions element if it has no buttons.
     $actions['#access'] = isset($actions['previous']) || isset($actions['next']);
     // Once these two steps are reached, the user can't go back.
-    if (in_array($this->stepId, ['offsite_payment', 'complete'])) {
+    if ($this->stepId == 'complete') {
```
bojanz: Why are we removing this logic? It's a part of the old 1.x logic, once payment is reached, the order is no longer a cart, and the customer cannot modify it anymore.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentInformation.php
```diff
@@ -122,8 +126,57 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+      /** @var \Drupal\profile\Entity\ProfileInterface $billing_profile */
+      $billing_profile = $this->order->getBillingProfile();
+      if (!$billing_profile) {
+        $billing_profile = Profile::create([
+          'uid' => $this->order->getCustomerId(),
+          'type' => 'customer',
+        ]);
+      }
+
+      $form_display = EntityFormDisplay::collectRenderDisplay($billing_profile, 'default');
```
bojanz, line 150: We keep repeating this bit, here, on the billing information pane, soon for shipping. We need to abstract it out somehow/somewhere.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentProcess.php
```diff
@@ -133,38 +133,43 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
       }
       catch (PaymentGatewayException $e) {
         drupal_set_message($e->getMessage(), 'error');
         $this->redirectToPreviousStep();
       }
     }
+    elseif ($payment_gateway_plugin instanceof OffsitePaymentGatewayInterface) {
+      if ($payment_gateway_plugin instanceof OffsiteRedirectPaymentGatewayInterface) {
+        $redirect_url = $payment_gateway_plugin->getRedirectUrl();
+        $pane_form['offsite_payment']['#prefix'] = '<div class="payment-redirect-form" data-redirect-url="' . $redirect_url . '">';
```
bojanz: In most cases a direct header-based redirect would be sufficient, no? Gateways that need an HTML redirect are a good demonstration of why we need an "offsite" plugin form, this code should definitely be overridable.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentProcess.php
```diff
@@ -133,38 +133,43 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+        $pane_form['offsite_payment']['#suffix'] = '</div>';
+        $pane_form['help']['#markup'] = '<div class="checkout-help">' . t('Please wait while you are redirected to the payment server. If nothing happens within 10 seconds, please click on the button below.') . '</div>';
+        $pane_form['#attached']['library'][] = 'commerce_payment/offsite_redirect';
+      }
+
+      $pane_form['offsite_payment'] = [
```
bojanz, line 163: Ah,so here it is, but we're just not using it yet.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentProcess.php
```diff
@@ -133,38 +133,43 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
     $payment_gateway = $this->order->payment_gateway->entity;
     $payment_gateway_plugin = $payment_gateway->getPlugin();
 
+    $payment_storage = $this->entityTypeManager->getStorage('commerce_payment');
+    $payment = $payment_storage->create([
```
bojanz, line 137: For offsite we are creating but not saving the payment? In general, our idea was to not create the payment until the user comes back, if possible. That way we don't get an entity per redirect.

mglaman:
@bojanz this is working now. I am using https://github.com/mglaman/commerce_paypal setup with PayPal standard. Ignore the code there ;) It's hardcoded a bit still.

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
mglaman: Sure, I'll change it back

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
mglaman: People should be able to go back to edit information or change payment gateway. You can in any checkout I've seen, especially single page step ones (Apple, etc.)

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentInformation.php
mglaman: True. But the one line we're repeating is this, which I don't think we can really abstract that much, because the lines following change.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentProcess.php
mglaman: We don't save it. We use it so that the `offsite-payment` knows the order and any other contexts when building the form. I chose that to be the `EntityWithPaymentGatewayInterface` passed.

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
bojanz: Confirmed with Matt that we do want this (the ability to go back)

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
```diff
@@ -148,6 +150,22 @@ public function getNextStepId() {
   }
 
   /**
+   * Redirects to previous step page.
```
bojanz: The docblock needs to be {inheritdoc}, the method is on the interface.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentProcess.php
```diff
@@ -133,38 +133,49 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+      // root form.
+      $pane_form['offsite_payment'] = [
+        '#type' => 'commerce_payment_gateway_form',
+        '#operation' => 'offsite-payment',
+        '#default_value' => $payment,
+      ];
+
+      if ($payment_gateway_plugin instanceof OffsiteRedirectPaymentGatewayInterface) {
+        $redirect_url = $payment_gateway_plugin->getRedirectUrl();
+        $pane_form['help']['#markup'] = '<div class="checkout-help">' . t('Please wait while you are redirected to the payment server. If nothing happens within 10 seconds, please click on the button below.') . '</div>';
```
bojanz: This still doesn't live in the offsite plugin form, even though it should.

modules/payment/commerce_payment.routing.yml
```diff
@@ -62,3 +62,25 @@ entity.commerce_payment_method.collection:
     parameters:
       user:
         type: entity:user
+
+# @todo only generate if checkout is enabled.
```
bojanz: If I remember correctly, a route can have a requirement on an enabled module, which would solve this.

bojanz (changes requested):

modules/payment/commerce_payment.routing.yml
mglaman: Yep, `\Drupal\Core\EventSubscriber\ModuleRouteSubscriber, with `_module_dependencies`. Good call.

modules/payment/src/Controller/CheckoutPaymentController.php
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Drupal\commerce_payment\Controller;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+use Drupal\commerce_payment\Plugin\Commerce\PaymentGateway\OffsitePaymentGatewayInterface;
+use Drupal\Core\Access\AccessException;
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class CheckoutPaymentController implements ContainerInjectionInterface {
```
mglaman: @bojanz should this be PaymentCheckoutController?

modules/payment/src/Controller/CheckoutPaymentController.php
```diff
@@ -0,0 +1,66 @@
...
+    return new static();
+  }
+
+  /**
+   * Controller callback for offsite payments cancelled.
+   *
+   * @param \Drupal\commerce_order\Entity\OrderInterface $commerce_order
+   *   The order.
+   */
+  public function checkoutPaymentCancel(OrderInterface $commerce_order) {
```
mglaman: `paymentCheckoutCancel` ?

modules/payment/src/Controller/CheckoutPaymentController.php
```diff
@@ -0,0 +1,66 @@
...
+    $checkout_flow_plugin->redirectToStep($checkout_flow_plugin->getPreviousStepId());
+  }
+
+  /**
+   * Controller callback for offsite payments returned.
+   *
+   * @param \Drupal\commerce_order\Entity\OrderInterface $commerce_order
+   *   The order.
+   */
+  public function checkoutPaymentReturn(OrderInterface $commerce_order) {
```
mglaman: `paymentCheckoutReturn` ?

modules/payment/src/Plugin/Commerce/PaymentGateway/OffsitePaymentGatewayInterface.php
```diff
@@ -0,0 +1,50 @@
...
+   */
+  public function getPaymentRedirectReturnUrl(OrderInterface $order);
+
+  /**
+   * Invoked when the off-site payment return.
+   *
+   * @param \Drupal\commerce_order\Entity\OrderInterface $order
+   *   The order.
+   */
+  public function onRedirectReturn(OrderInterface $order);
```
mglaman: `onPaymentRedirectReturn` ?

modules/payment/src/Plugin/Commerce/PaymentGateway/OffsitePaymentGatewayInterface.php
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Invoked when the off-site payment way cancelled or failed.
+   *
+   * @param \Drupal\commerce_order\Entity\OrderInterface $order
+   *   The order.
+   */
+  public function onRedirectCancel(OrderInterface $order);
```
mglaman: `onPaymentRedirectCancel` ?

modules/payment/src/Plugin/Commerce/PaymentGateway/OffsiteRedirectPaymentGatewayInterface.php
```diff
@@ -0,0 +1,22 @@
...
+   * Gets the off-site redirect URL.
+   *
+   * If this is TRUE, a form wrapper and JavaScript snippet will be added that
+   * submits the off-site payment form, causing a redirect to the payment
+   * gateway's payment page.
+   *
+   * @return bool
+   *   Whether to automatically redirect or not.
+   */
+  public function getRedirectUrl();
```
mglaman: `getPaymentRedirectUrl` ?

modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
```diff
@@ -107,6 +107,9 @@ protected function getDefaultForms() {
     if ($this instanceof SupportsRefundsInterface) {
       $default_forms['refund-payment'] = 'Drupal\commerce_payment\PluginForm\PaymentRefundForm';
     }
+    if ($this instanceof OffsitePaymentGatewayInterface) {
+      $default_forms['offsite-payment'] = 'Drupal\commerce_payment\PluginForm\OffsitePaymentForm';
```
mglaman: I just realized this is an abstract class. @bojanz we should just remove this, yeah? I think if its missing it'll throw an exception or something.

modules/payment/src/PluginForm/OffsitePaymentForm.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Drupal\commerce_payment\PluginForm;
+
+use Drupal\commerce_payment\Plugin\Commerce\PaymentGateway\OffsiteRedirectPaymentGatewayInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Render\Element;
+
+abstract class OffsitePaymentForm extends PaymentGatewayFormBase {
```
mglaman: @bojanz should we document that all off-site payments _must_ extend this and they're expected to override `buildConfigurationForm` and enter any data to be POST'd to the gateway? Or do handhsake and throw `NeedsRedirectException` in their form build?

mglaman (commented):
@bojanz can you see these comments, mostly naming on some logic.

mglaman:
Discussed with @bojanz as well: Going to add an Offsite iframe and Offsite 302 redirect example [with tests.]

modules/payment/src/Controller/CheckoutPaymentController.php
bojanz: That might better, yeah.

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
```diff
@@ -374,7 +392,7 @@ protected function actions(array $form, FormStateInterface $form_state) {
     // Hide the actions element if it has no buttons.
     $actions['#access'] = isset($actions['previous']) || isset($actions['next']);
     // Once these two steps are reached, the user can't go back.
-    if (in_array($this->stepId, ['offsite_payment', 'complete'])) {
+    if ($this->stepId == 'complete') {
       $actions['#access'] = FALSE;
```
bojanz: Looks like we can now remove the entire if. The "complete" step doesn't have buttons defined, so that's enough to hide the actions anyway.

modules/payment/src/Controller/CheckoutPaymentController.php
```diff
@@ -0,0 +1,66 @@
...
+use Drupal\Core\Access\AccessException;
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class CheckoutPaymentController implements ContainerInjectionInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
```
bojanz: No point in having a create() if we aren't injecting anything.

modules/payment/src/Controller/CheckoutPaymentController.php
bojanz: what about cancelPaymentPage()? We tend to suffix these methods with Page, and it sounds more descriptive.

modules/payment/src/Controller/CheckoutPaymentController.php
bojanz: Same as above, returnPaymentPage?

modules/payment/src/Controller/CheckoutPaymentController.php
```diff
@@ -0,0 +1,66 @@
...
+   * Controller callback for offsite payments cancelled.
+   *
+   * @param \Drupal\commerce_order\Entity\OrderInterface $commerce_order
+   *   The order.
+   */
+  public function checkoutPaymentCancel(OrderInterface $commerce_order) {
+    /** @var \Drupal\commerce_payment\Entity\PaymentGatewayInterface $payment_gateway */
+    $payment_gateway = $commerce_order->payment_gateway->entity;
+    $payment_gateway_plugin = $payment_gateway->getPlugin();
+
```
bojanz: Unneeded newline before if(

modules/payment/src/Controller/CheckoutPaymentController.php
```diff
@@ -0,0 +1,66 @@
...
+    if (!$payment_gateway_plugin instanceof OffsitePaymentGatewayInterface) {
+      throw new AccessException('The payment gateway for the order does not implement ' . OffsitePaymentGatewayInterface::class);
+    }
+
+    $payment_gateway_plugin->onRedirectCancel($commerce_order);
+
```
bojanz: Unneeded newline before @ var

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentProcess.php
```diff
@@ -133,38 +133,43 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+        $this->checkoutFlow->redirectToStep($this->checkoutFlow->getNextStepId());
 
       }
       catch (PaymentGatewayException $e) {
         drupal_set_message($e->getMessage(), 'error');
         $this->redirectToPreviousStep();
       }
     }
+    elseif ($payment_gateway_plugin instanceof OffsitePaymentGatewayInterface) {
+      // Attach to the complete form, so that the form elements are part of the
```
bojanz: This comment doesn't seem to be up to date anymore?

modules/payment/src/Plugin/Commerce/PaymentGateway/OffsitePaymentGatewayInterface.php
bojanz: Maybe we just remove Payment from all of then? getRedirectReturnUrl() getRedirectCancelUrl(), onRedirectReturn() onRedirectCancel()? The return/cancel duality is confusing, cause technically both urls are return urls, one is for success one is for failure. Maybe we do SuccessUrl and CancelUrl() instead?

modules/payment/src/PluginForm/OffsitePaymentForm.php
bojanz: This form is usable out of the box, no? Cause it just sends you to the redirect url?
We still aren't covering all three use cases (direct redirect, form redirect, iframe), for which we probably want 3 different classes.

bojanz (changes requested):

modules/payment/src/Plugin/Commerce/PaymentGateway/OffsitePaymentGatewayInterface.php
mglaman: I made those changes except Return/Success. Just because the payment gateway returned doesn't mean it was successful, does it? It just means that payment was submitted and returned and the transaction can be processed.

I was wondering same thing about naming in 1.x, but it makes sense that there could be other processing See PayPal code for 1.x

```
  // If we didn't get an approval response code...
  switch ($ipn['payment_status']) {
    case 'Failed':
      $transaction->status = COMMERCE_PAYMENT_STATUS_FAILURE;
      $transaction->message = t("The payment has failed. This happens only if the payment was made from your customer’s bank account.");
      break;

    case 'Voided':
      $transaction->status = COMMERCE_PAYMENT_STATUS_FAILURE;
      $transaction->message = t('The authorization was voided.');
      break;

    case 'Pending':
      $transaction->status = COMMERCE_PAYMENT_STATUS_PENDING;
      $transaction->message = commerce_paypal_ipn_pending_reason($ipn['pending_reason']);
      break;

    case 'Completed':
      $transaction->status = COMMERCE_PAYMENT_STATUS_SUCCESS;
      $transaction->message = t('The payment has completed.');
      break;

    case 'Refunded':
      $transaction->status = COMMERCE_PAYMENT_STATUS_SUCCESS;
      $transaction->message = t('Refund for transaction @txn_id', array('@txn_id' => $ipn['parent_txn_id']));
      break;
  }
```

modules/payment/src/PluginForm/OffsitePaymentForm.php
mglaman: I do not think we need a form for each case.

1. Direct redirect: you must do some kind of handshake with the gateway. You send order details, they send a one time checkout link. This has to be performed in the form and `NeedsRedirectException` thrown. No need for its own form class.
2. POST redirect: you'll need to add form items, this covers it.
3. Iframe: covered. If the gateway plugin does not implement `OffsiteRedirectPaymentGatewayInterface` it will include their form, which should have hte iframe code. It does not attempt to redirect.

Each implementation is so wildly different I don't know if we could provide enough boilerplate code to support it. We have _just enough_ now to make working a lot easier.

modules/payment/src/Plugin/Commerce/PaymentGateway/PaymentGatewayBase.php
mglaman: Made non-abstract

PR #586, opened 2016-12-23
Issue #2659032 by mglaman, steveoliver: Add the cache contexts for th…

mglaman:
…e cart block

@bojanz with our test coverage this can be easily tested. Such as commenting out code within `\Drupal\commerce_cart\Plugin\Block\CartBlock::getCacheTags`.

### DIFF
--- a/modules/cart/src/Cache/Context/CartCacheContext.php
+++ b/modules/cart/src/Cache/Context/CartCacheContext.php
@@ -59,7 +59,11 @@ class CartCacheContext implements CacheContextInterface {
    * {@inheritdoc}
    */
   public function getCacheableMetadata() {
-    return new CacheableMetadata();
+    $metadata = new CacheableMetadata();
+    foreach ($this->cartProvider->getCarts($this->account) as $cart) {
+      $metadata->addCacheableDependency($cart);
+    }
+    return $metadata;
   }
 
 }
--- a/modules/cart/src/Plugin/Block/CartBlock.php
+++ b/modules/cart/src/Plugin/Block/CartBlock.php
@@ -4,6 +4,7 @@ namespace Drupal\commerce_cart\Plugin\Block;
 
 use Drupal\commerce_cart\CartProviderInterface;
 use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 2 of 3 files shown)


modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -206,11 +205,25 @@ protected function getCartViews(array $carts) {
...
-  public function getCacheMaxAge() {
-    return 0;
+  public function getCacheContexts() {
+    return Cache::mergeContexts(parent::getCacheContexts(), ['cart']);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
```
mglaman, line 216: I thought this would just work with the cache tags from the cache context, but it does not. If you remove this bit the test fails. So even though the cache context's tags bust, these do not.

At the same time if the cache tags are not in the context it does not bust.

mglaman: Also, for what it's worth, I sent a tweet to WimLeers to see if he has any input to make sure this is right

https://twitter.com/nmdmatt/status/812119091548917760

modules/cart/tests/src/Unit/CartCacheContextTest.php
```diff
@@ -20,9 +21,15 @@ public function testCartCacheContext() {
     $account = $this->getMock(AccountInterface::class);
     $cartProvider = $this->getMock(CartProviderInterface::class);
     $cartProvider->expects($this->once())->method('getCartIds')->willReturn(['23', '34']);
+    $cartProvider->expects($this->once())->method('getCarts')->willReturn([
+      new TestCacheableDependency([], ['commerce_cart:23'], 0),
+      new TestCacheableDependency([], ['commerce_cart:24'], 0),
+
```
bojanz: Trailing newline.

modules/cart/src/Cache/Context/CartCacheContext.php
```diff
@@ -59,7 +59,12 @@ public function getContext() {
    * {@inheritdoc}
    */
   public function getCacheableMetadata() {
-    return new CacheableMetadata();
+    $metadata = new CacheableMetadata();
+    // Add each cart as a cacheable dependency.
```
bojanz: Redundant comment of the day.

mglaman: You're snarky today

modules/cart/src/Plugin/Block/CartBlock.php
wimleers: And I totally failed to respond to that. Got list in the twitter whirlwinds.

wimleers: And I totally failed to respond to that. Got list in the twitter whirlwinds.

wimleers: A cache context's tags are only used when a cache context is "optimized away". E.g. when you have both `'user'` and `'user.permissions'` cache contexts, the `'user.permissions'` cache context is optimized away. Only then is `\Drupal\Core\Cache\Context\AccountPermissionsCacheContext::getCacheableMetadata()` used.

The `cart` cache context cannot ever be optimized away. So it's rather pointless for that cache context to even return cache tags.

See https://www.drupal.org/docs/8/api/cache-api/cache-contexts#optimizing.

PR #588, opened 2016-12-23
Issue #2838978 by bojanz: Allow offsite gateways to be used without checkout

bojanz:


bojanz:
Committed in 10032eee06628ad8e33ddc88c8fa83f3a90729f7.

PR #592, opened 2016-12-30
Issue #2816237: Create an AdjustmentType plugin for Adjustments

steveoliver:
See https://www.drupal.org/node/2816237


mglaman:
On phone, but

* Why field definitions in base?
* No need for schema.yml changes, no config entity 
* Should it be a YAML plugin? Having the class doesn't do too much for each adjustment type, does it?

steveoliver:
ah-ha - I wasn't sure exactly how to implement the plugins, but yaml does make sense. And those schema changes made it in accidentally, from when i originally thought we'd have UI and config based types.

mglaman:
Cool! Thanks for tackling this

mglaman:
Looks great. Can we maybe get a test module which provides an adjustment type and add test coverage to widget?

PR #594, opened 2017-01-04
Issue #2840864: Implement the order summary totals service.

steveoliver:
See https://www.drupal.org/node/2840864.

### DIFF
--- a/modules/order/src/Entity/Order.php
+++ b/modules/order/src/Entity/Order.php
@@ -298,6 +298,29 @@ class Order extends ContentEntityBase implements OrderInterface {
   /**
    * {@inheritdoc}
    */
+  public function getSubtotalPrice() {
+    $total_price = $this->getTotalPrice();
+    if ($total_price) {
+      $currency_code = $total_price->getCurrencyCode();
+    }
+    else {
+      $currency_code = $this->initializeCurrencyCode();
+      if (!$currency_code) {
+        // The order object is not complete enough to have a subtotal price yet.
+        return NULL;
+      }
+    }
+
+    $total_price = new Price('0', $currency_code);
+    foreach ($this->getItems() as $order_item) {
+      $total_price = $total_price->add($order_item->getTotalPrice());
+    }
+    return $total_price;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function getTotalPrice() {
     if (!$this->get('total_price')->isEmpty()) {
       return $this->get('total_price')->first()->toPrice();
@@ -422,6 +445,34 @@ class Order extends ContentEntityBase implements OrderInterface {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function collectAdjustments() {
+    $adjustments = [];
+
+    foreach ($this->getItems() as $order_item) {
+      foreach ($order_item->getAdjustments() as $adjustment) {
+        $adjustments[] = [
+          'type' => $adjustment->getType(),
+          'label' => $adjustment->getLabel(),
+          'source_id' => $adjustment->getSourceId(),
+          'amount' => $adjustment->getAmount()->multiply($order_item->getQuantity()),
+        ];
+      }
+    }
+    foreach ($this->getAdjustments() as $adjustment) {
+      $adjustments[] = [
+        'type' => $adjustment->getType(),
+        'label' => $adjustment->getLabel(),
+        'source_id' => $adjustment->getSourceId(),
+        'amount' => $adjustment->getAmount(),
+      ];
+    }
+
+    return $adjustments;
+  }
+
+  /**
    * Recalculates the order item total price.
    */
   protected function recalculateTotalPrice() {
@@ -440,13 +491,9 @@ class Order extends ContentEntityBase implements OrderInterface {
     $total_price = new Price('0', $currency_code);
     foreach ($this->getItems() as $order_item) {
       $total_price = $total_price->add($order_item->getTotalPrice());
-      foreach ($order_item->getAdjustments() as $adjustment) {
-        $adjustment_total = $adjustment->getAmount()->multiply($order_item->getQuantity());
-        $total_price = $total_price->add($adjustment_total);
-      }
     }
-    foreach ($this->getAdjustments() as $adjustment) {
-      $total_price = $total_price->add($adjustment->getAmount());
+    foreach ($this->collectAdjustments() as $adjustment) {
+      $total_price = $total_price->add($adjustment['amount']);
     }
     $this->total_price = $total_price;
   }
@@ -619,6 +666,10 @@ class Order extends ContentEntityBase implements OrderInterface {
       ->setLabel(t('Total price'))
       ->setDescription(t('The total price of the order.'))
       ->setReadOnly(TRUE)
+      ->setDisplayOptions('view', [
+        'label' => 'hidden',
+        'type' => 'commerce_order_total_summary'
+      ])
       ->setDisplayConfigurable('form', FALSE)
       ->setDisplayConfigurable('view', TRUE);
 
--- a/modules/order/src/OrderTotalSummary.php
+++ b/modules/order/src/OrderTotalSummary.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Drupal\commerce_order;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+
+class OrderTotalSummary implements OrderTotalSummaryInterface {
+
+  /**
+   * The adjustment type manager.
+   *
+   * @var \Drupal\commerce_order\AdjustmentTypeManager
+   */
+  protected $adjustmentTypeManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(AdjustmentTypeManager $adjustment_type_manager) {
+    $this->adjustmentTypeManager = $adjustment_type_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildTotals(OrderInterface $order) {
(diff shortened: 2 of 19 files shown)


modules/order/src/Entity/Order.php
```diff
@@ -422,6 +445,43 @@ public function preSave(EntityStorageInterface $storage) {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function collectAdjustments() {
```
mglaman, line 450: I know this is what @bojanz wanted, but it feels weird to have this here and then the next change where we don't just do `::getAdjustments()` directly. This feels like it belongs in a service to get the adjustments and their type metadata

modules/order/src/Entity/Order.php
```diff
@@ -440,13 +500,9 @@ protected function recalculateTotalPrice() {
...
     foreach ($this->getItems() as $order_item) {
       $total_price = $total_price->add($order_item->getTotalPrice());
-      foreach ($order_item->getAdjustments() as $adjustment) {
-        $adjustment_total = $adjustment->getAmount()->multiply($order_item->getQuantity());
-        $total_price = $total_price->add($adjustment_total);
-      }
     }
-    foreach ($this->getAdjustments() as $adjustment) {
-      $total_price = $total_price->add($adjustment->getAmount());
+    foreach ($this->collectAdjustments() as $adjustment) {
```
mglaman: See above, feels weird.

modules/order/src/Entity/Order.php
mglaman: Like it should just be part of OrderTotalSummary.

modules/order/src/OrderTotalSummary.php
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildTotals(OrderInterface $order) {
+    $total = $order->getTotalPrice();
+    $subtotal = $order->getSubtotalPrice();
+    $collected_adjustments = $order->collectAdjustments();
+    $types = $this->adjustmentTypeManager->getDefinitions();
```
mglaman, line 30: collectAdjustments gets definitions, then we do again here. Furhter why I think collectAdjustments should be removed / moved here.

modules/order/src/OrderTotalSummary.php
```diff
@@ -0,0 +1,87 @@
...
+
+      // Group by adjustment type.
+      if (!isset($adjustments[$type])) {
+        $adjustments[$type] = [
+          'label' => $label,
+          'weight' => $weight,
+          'items' => [],
+        ];
+      }
+      if (is_null($adjustment['source_id'])) {
```
mglaman: why not just `empty()`

modules/order/templates/commerce-order-total-summary.html.twig
```diff
@@ -0,0 +1,31 @@
...
+ *   - total: \Drupal\commerce_price\Price|null The order total amount.
+ *
+ * @ingroup themeable
+ */
+#}
+<div{{ attributes }}>
+  {{ 'Subtotal'|t }}: {{ totals.subtotal|commerce_price_format }}<br/>
+  {% for group in totals.adjustments %}
+    {% for item in group.items %}
+      {{ item.label }}: {{ item.amount|commerce_price_format }}<br/>
```
mglaman: We should do some basic `div` splitting it up, or something inline. Spans? so it can be easily styled w/o overriding

modules/order/tests/src/Kernel/Entity/OrderTest.php
```diff
@@ -105,6 +107,9 @@ protected function setUp() {
    * @covers ::setCompletedTime
    */
   public function testOrder() {
+    /** @var \Drupal\commerce_order\OrderTotalSummaryInterface $order_total_summary */
+    $order_total_summary = \Drupal::service('commerce_order.order_total_summary');
```
mglaman: Order summary should go in its own kernel?

mglaman (changes requested):
feels weird having the collect adjustment methods internals within the order class, willing to debate

modules/order/src/Entity/Order.php
bojanz: You need a way to get all order adjustments, both the ones on the order-level and the ones on the order-item level. getAdjustments() doesn't give you that.
This is a by-product of storing order-item-level adjustments on the order items themselves. Having it on the OrderTotalSummary would make sense if the Order class itself didn't need it.

modules/order/src/OrderTotalSummary.php
bojanz: We can remove the definitions from collectAdjustment(), question was where to filter out adjustments with invalid types. If we want to make client classes responsible for it, that's fine.

modules/order/templates/commerce-order-total-summary.html.twig
bojanz: Yeah, we need something that resembles the 1.x styling (two columns).

modules/order/src/Entity/Order.php
mglaman: Ah that's right. I forgot we need a way for the order to get all adjustments from its items. I guess I just got thrown off by the content of the method, thinking we could offload it.

modules/order/src/OrderTotalSummary.php
mglaman: ah, gotcha. yeah I suppose we do not want adjustments existing with invalid types :/

bojanz: I guess it makes sense to move the type check to Adjustment::__construct and remove it from collectAdjustments(). That way we explode right away instead of silently ignoring broken data, probably a better idea.

steveoliver:
I addressed all the comments @mglaman and @bojanz made above.  The new kernel test has notes about its fails--even though the totals reflect the adjustments, adjustments with the custom 'test_adjustment_type' do not show up in the 'adjustments' of OrderTotalSummary::buildTotals; I've tried adding and using other types, only 'custom' ever shows up in the adjustments list... :/

modules/order/src/OrderTotalSummary.php
steveoliver: I did move the [validation of type into the Adjustment class](https://github.com/drupalcommerce/commerce/pull/594/files#diff-8361968a5f1463fc9c3e3043ba5b7c43) itself, with [a test for the expected exception](https://github.com/drupalcommerce/commerce/pull/594/files#diff-11a14da6d844f9a1efd7e467b1b41a93).

steveoliver: We still need the adjustment type manager here though to get the weights for ordering.

modules/order/tests/src/Kernel/OrderTotalSummaryTest.php
```diff
@@ -0,0 +1,176 @@
...
+    $order->save();
+
+    /** @var \Drupal\commerce_order\Entity\OrderItemInterface $order_item */
+    $order_item = OrderItem::create([
+      'type' => 'default',
+      'quantity' => 2,
+      'unit_price' => new Price('12.00', 'USD'),
+    ]);
+    $order_item->save();
+
```
steveoliver, line 116: subtotal: 24.00
total: 24.00

modules/order/tests/src/Kernel/OrderTotalSummaryTest.php
```diff
@@ -0,0 +1,176 @@
...
+    $order_item->addAdjustment(new Adjustment([
+      'type' => 'custom',
+      'label' => '1 dollar off',
+      'amount' => new Price('-1.00', 'USD'),
+    ]));
+    $order_item->save();
+
```
steveoliver, line 123: subtotal: 24.00
total: 22.00

modules/order/tests/src/Kernel/OrderTotalSummaryTest.php
```diff
@@ -0,0 +1,176 @@
...
+    // This adjustment should be first.
+    $order_item->addAdjustment(new Adjustment([
+      'type' => 'test_adjustment_type',
+      'label' => '50 cent item fee',
+      'amount' => new Price('0.50', 'USD'),
+    ]));
+
```
steveoliver, line 130: subtotal: 24.00
total: 23.00

modules/order/tests/src/Kernel/OrderTotalSummaryTest.php
```diff
@@ -0,0 +1,176 @@
...
+    $order_item = $this->reloadEntity($order_item);
+    $order->addItem($order_item);
+
+    // Add order adjustments.
+    $order->addAdjustment(new Adjustment([
+      'type' => 'custom',
+      'label' => '5 dollars off',
+      'amount' => new Price('-5.00', 'USD'),
+    ]));
+
```
steveoliver, line 141: subtotal: 24.00
total: 18.00

modules/order/tests/src/Kernel/OrderTotalSummaryTest.php
```diff
@@ -0,0 +1,176 @@
...
+    $order->addAdjustment(new Adjustment([
+      'type' => 'custom',
+      'label' => 'Handling fee',
+      'amount' => new Price('10.00', 'USD'),
+    ]));
+
```
steveoliver, line 147: subtotal: 24.00
total: 28.00

steveoliver (commented):
Just pointing out why I think the amount assertions are correct.

bojanz:
This was committed in ea1ce2a2d57a959ea0817b2bf81ad0941a6ca79f

PR #595, opened 2017-01-05
Pass the transaction mode from checkout to the offsite payment form

maciejzgadzaj:


mglaman:
@bojanz we'll need this for the off-site API

PR #597, opened 2017-01-07
Issue #2810723: Allow order types to have no carts

scotthooker:
Started looking at the more generic way to disable the cart (and subsequently the add to cart message for straight to checkout)


bojanz:
Let's continue on the issue, which has the more recent work.

PR #598, opened 2017-01-10
Issue #2842256: Provide a getter method for adjusted price of order items

steveoliver:
See https://www.drupal.org/node/2842256.


modules/order/src/Entity/OrderItemInterface.php
```diff
@@ -99,14 +99,22 @@ public function getUnitPrice();
...
    */
   public function getTotalPrice();
 
   /**
+   * Gets the order item total price, with all order item adjustments applied.
+   *
+   * @return @return \Drupal\commerce_price\Price|null
+   *   The order item adjusted total price, or NULL.
+   */
+  public function getTotalAdjustedPrice();
```
bojanz: getAdjustedTotalPrice() sounds better?

No need to change the docblock for total price, that's still the total that people will use.
Instead, we need to explain the adjusted total price, saying that it's for informative use only.
Let's bikeshed it on IRC.

steveoliver:
I agree. Just pushed changes requested, also fixes coding standards fix.

modules/order/tests/src/Kernel/Entity/OrderItemTest.php
```diff
@@ -87,7 +89,6 @@ public function testOrderItem() {
     $order_item->setUnitPrice($unit_price);
     $this->assertEquals($unit_price, $order_item->getUnitPrice());
 
-    $order_item->setQuantity('1');
```
steveoliver: I am removing this setQuantity() call here in order to test Unit / Total prices below (otherwise they would be the same value, and the quantity check is already happening in the test assertions above).

finnef:
Review: The code might need to be aware of non-percentage adjustments on order items. You could have a plain €20 off adjustment. We need to know if the adjustment is a percentage or not, I think.

Edit: Thinking this through it is fine here, but needs to be taken into account when applying adjustments.

bojanz:
The note about knowing if an adjustment is a percentage was resolved in RC2 by https://www.drupal.org/node/2894792

bojanz:
Did a round of cleanup (to account for included adjustments, rounding logic) and committed in 5fcec50d2c23e5cbfb1676e62e81442ea7ed0a6b. Thanks!

PR #600, opened 2017-01-11
Updating Workflow, adding new capture_pending state for bank transfers.

cmcintosh:
This PR adds in a new workflow state, capture_pending.  The idea is that some payment methods like Bank transfers may take a while to complete, so we should have a state that payments can sit in while we wait on a response from the processor. This was based on a brief conversation on IRC with @bojanz


bojanz:
We'll need to expand this PR before committing it, cause we'll need a way for a capture_pending payment to get to capture_complete (manual payment form, or a method that contacts a remote API).

Leaving it for after RC1 once we have more clarity.

bojanz:
Continued in https://www.drupal.org/project/commerce/issues/2988714

PR #602, opened 2017-01-12
Adding condition for Promotions, based on customer role.

cmcintosh:
This PR adds a condition plugin to allow setting a discount based on the order's customer role.


bojanz:
Replaced by https://www.drupal.org/node/2885365
Gave you credit in that issue.

PR #606, opened 2017-01-16
Avoid WSOD when adding a new gateway when no plugins are available

mglaman:
…teway when no gateway plugins are available

### DIFF
--- a/modules/payment/src/Form/PaymentGatewayForm.php
+++ b/modules/payment/src/Form/PaymentGatewayForm.php
@@ -39,6 +39,21 @@ class PaymentGatewayForm extends CommercePluginEntityFormBase {
   /**
    * {@inheritdoc}
    */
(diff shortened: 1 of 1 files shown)


modules/payment/src/Form/PaymentGatewayForm.php
```diff
@@ -39,6 +39,21 @@ public static function create(ContainerInterface $container) {
   /**
    * {@inheritdoc}
    */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    // Skip building the form if there are no available gateway plugins.
+    if (empty($this->pluginManager->getDefinitions())) {
+      $form['warning'] = [
+        '#markup' => $this->t('No payment gateways are defined'),
```
mglaman: > No payment gateway plugins found. Please install a module which provides one.

PR #608, opened 2017-01-16
Issue #2833780 by mglaman: Cancelled orders that are carts keep havin…

mglaman:
…g products added to them

### DIFF
--- a/modules/cart/src/EventSubscriber/OrderEventSubscriber.php
+++ b/modules/cart/src/EventSubscriber/OrderEventSubscriber.php
@@ -29,7 +29,8 @@ class OrderEventSubscriber implements EventSubscriberInterface {
    * {@inheritdoc}
    */
   public static function getSubscribedEvents() {
-    $events = ['commerce_order.place.pre_transition' => 'finalizeCart'];
+    $events['commerce_order.place.pre_transition'] = 'finalizeCart';
+    $events['commerce_order.cancel.pre_transition'] = 'cancelCart';
     return $events;
   }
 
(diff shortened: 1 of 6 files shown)


modules/cart/tests/src/Kernel/CartOrderCanceledTest.php
```diff
@@ -0,0 +1,139 @@
...
+    $this->variation->save();
+
+    // Create a user to use for orders.
+    $this->user = $this->createUser();
+  }
+
+  /**
+   * Tests that a draft order is no longer a cart once placed.
+   */
+  public function testCartOrderPlaced() {
```
bojanz: Wrong method name and docblock.

modules/cart/tests/src/Kernel/CartOrderCanceledTest.php
```diff
@@ -0,0 +1,139 @@
+<?php
+
+namespace Drupal\Tests\commerce_cart\Kernel;
+
+use Drupal\Component\Render\FormattableMarkup;
+use Drupal\Tests\commerce\Kernel\CommerceKernelTestBase;
+
+/**
+ * Tests the unsetting of the cart flag when order is canceled.
```
bojanz: when *the* order is canceled.

modules/cart/src/EventSubscriber/OrderEventSubscriber.php
```diff
@@ -46,4 +47,17 @@ public function finalizeCart(WorkflowTransitionEvent $event) {
     }
   }
 
+  /**
+   * Finalizes the cart when the order is cancelled.
```
bojanz: "cancelled" should be "canceled" to avoid making the bald eagle'sad?

modules/cart/src/EventSubscriber/OrderEventSubscriber.php
```diff
@@ -46,4 +47,17 @@ public function finalizeCart(WorkflowTransitionEvent $event) {
...
+   *
+   * @param \Drupal\state_machine\Event\WorkflowTransitionEvent $event
+   *   The workflow transition event.
+   */
+  public function cancelCart(WorkflowTransitionEvent $event) {
+    $order = $event->getEntity();
+    if ($order->cart->value == TRUE) {
+      $this->cartProvider->finalizeCart($order, FALSE);
```
bojanz: This is wrong, because finalizeCart adds the cart ID to the completed carts list in the session.
I guess CartProvider needs a cancelCart as well.

bojanz:
Committed the final version in 8f1e82977037238b31b970425932c5e459fad0e1.

PR #609, opened 2017-01-18
#2842356 Rethink the default order number generation

agoradesign:
see https://www.drupal.org/node/2842356


modules/order/src/EventSubscriber/OrderNumberSubscriber.php
```diff
@@ -0,0 +1,34 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class OrderNumberSubscriber implements EventSubscriberInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    $events = [
+      'commerce_order.place.pre_transition' => ['setOrderNumber', -100],
```
mglaman, line 21: We'll want to document this, considering it's on event level priority and not resolving like our normal cases.

bojanz:
Added the docs that @mglaman wanted and committed!

PR #610, opened 2017-01-20
Add payment logging

maciejzgadzaj:
See https://www.drupal.org/node/2845321

### DIFF
--- a/modules/log/src/EventSubscriber/PaymentEventSubscriber.php
+++ b/modules/log/src/EventSubscriber/PaymentEventSubscriber.php
@@ -167,6 +167,7 @@ class PaymentEventSubscriber implements EventSubscriberInterface {
     $payment = $event->getPayment();
     $this->logStorage->generate($payment, 'payment_partially_captured', [
       'payment_remote_id' => $payment->getRemoteId(),
+      'captured_amount' => $event->getAmount(),
     ])->save();
   }
 
@@ -193,6 +194,7 @@ class PaymentEventSubscriber implements EventSubscriberInterface {
     $payment = $event->getPayment();
     $this->logStorage->generate($payment, 'payment_partially_refunded', [
       'payment_remote_id' => $payment->getRemoteId(),
+      'refunded_amount' => $event->getAmount(),
     ])->save();
   }
 
--- a/modules/log/src/EventSubscriber/PaymentEventSubscriber.php
+++ b/modules/log/src/EventSubscriber/PaymentEventSubscriber.php
@@ -2,6 +2,9 @@
 
 namespace Drupal\commerce_log\EventSubscriber;
 
+use Drupal\commerce_payment\Entity\PaymentInterface;
+use Drupal\commerce_payment\Event\PaymentEvent;
+use Drupal\commerce_payment\Event\PaymentEvents;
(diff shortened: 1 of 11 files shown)


maciejzgadzaj:
Result:

- with full captures/refunds:

order-activity-log-full

- with partial captures/refunds:

order-activity-log

heddn (changes requested):
I think the events that were added are a duplicate of the already added ones provided by state_machine. No need to add the Event and Events classes. I'll do a quick review of the rest of the patch as well.

modules/log/src/EventSubscriber/PaymentEventSubscriber.php
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /**
+   * Creates a log when a payment is authorized.
+   *
+   * @param \Drupal\state_machine\Event\WorkflowTransitionEvent $event
+   *   The transition event.
+   */
+  public function onAuthorizeTransition(WorkflowTransitionEvent $event) {
+    $this->dispatch(PaymentEvents::PAYMENT_AUTHORIZED, $event->getEntity());
```
heddn, line 59: Shouldn't this log?

modules/log/src/EventSubscriber/PaymentEventSubscriber.php
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /**
+   * Creates a log when a payment is voided.
+   *
+   * @param \Drupal\state_machine\Event\WorkflowTransitionEvent $event
+   *   The transition event.
+   */
+  public function onVoidTransition(WorkflowTransitionEvent $event) {
+    $this->dispatch(PaymentEvents::PAYMENT_VOIDED, $event->getEntity());
```
heddn, line 69: Why no logging?

modules/log/src/EventSubscriber/PaymentEventSubscriber.php
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /**
+   * Creates a log when a payment is expired.
+   *
+   * @param \Drupal\state_machine\Event\WorkflowTransitionEvent $event
+   *   The transition event.
+   */
+  public function onExpireTransition(WorkflowTransitionEvent $event) {
+    $this->dispatch(PaymentEvents::PAYMENT_EXPIRED, $event->getEntity());
```
heddn, line 79: And on through the rest of this subscriber...

modules/payment/src/Entity/Payment.php
```diff
@@ -24,6 +24,7 @@
  *   bundle_plugin_type = "commerce_payment_type",
  *   handlers = {
  *     "access" = "Drupal\commerce_payment\PaymentAccessControlHandler",
+ *     "event" = "Drupal\commerce_payment\Event\PaymentEvent",
```
heddn, line 27: I don't think this is necessary if we state_machine

heddn (changes requested):

modules/payment/src/Entity/Payment.php
mglaman: It is, because payments do not use transitions.

mglaman:
Closing for #797. Will make sure issues credits correct on D.o issue.

PR #613, opened 2017-01-25
Issue #2838380: Allow offsite payment gateways to create and use payment methods

maciejzgadzaj:
See https://www.drupal.org/node/2838380

### DIFF
--- a/modules/payment/src/Plugin/Commerce/PaymentMethodType/PayPal.php
+++ b/modules/payment/src/Plugin/Commerce/PaymentMethodType/PayPal.php
@@ -20,10 +20,15 @@ class PayPal extends PaymentMethodTypeBase {
    * {@inheritdoc}
    */
   public function buildLabel(PaymentMethodInterface $payment_method) {
-    $args = [
-      '@paypal_mail' => $payment_method->paypal_mail->value,
-    ];
-    return $this->t('PayPal account (@paypal_mail)', $args);
+    if ($payment_method->paypal_mail->value) {
+      $args = [
+        '@paypal_mail' => $payment_method->paypal_mail->value,
+      ];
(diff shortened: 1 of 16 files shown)


modules/payment/src/Plugin/Commerce/PaymentMethodType/PayPal.php
```diff
@@ -20,10 +20,15 @@ class PayPal extends PaymentMethodTypeBase {
...
-    ];
-    return $this->t('PayPal account (@paypal_mail)', $args);
+    if ($payment_method->paypal_mail->value) {
+      $args = [
+        '@paypal_mail' => $payment_method->paypal_mail->value,
+      ];
+      return $this->t('PayPal account (@paypal_mail)', $args);
+    }
+    else {
+      return $this->t('New PayPal account');
```
bojanz: There is no need to modify the buildLabel() of the payment method types, you're just duplicating the create_label from the annotation.

bojanz:
Doesn't modify PaymentGatewayStorage::loadForUser() nor PaymentGatewayBase::getDefaultForms().

maciejzgadzaj:
- payment method type label builders reverted to original version
- `PaymentGatewayBase::getDefaultForms()` updated
- `PaymentGatewayStorage::loadForUser()` not updated - not sure what/if anything needs to be changed there?
- payment method is no longer saved in case of offsite redirects (related: https://www.drupal.org/node/2827152)
- added new `SupportsCreatingStoredPaymentMethodsOffsiteInterface` for stand-alone offsite payment method creation

bojanz:
Let's continue on the issue, since it has continued discussion, and we're phasing out PRs.

PR #614, opened 2017-01-27
Issue #2847378: Add support for partial captures

maciejzgadzaj:
See https://www.drupal.org/node/2847378


bojanz:
See https://www.drupal.org/node/2847378#comment-12115052

PR #619, opened 2017-01-29
Issue #2822211: Limit allowed billing addresses by store billing countries

steveoliver:
See https://www.drupal.org/node/2822211


bojanz:
Let's add an else, so that if there's no order, we load the default store and use that.

Would be great to have test coverage as well, in PaymentMethodTest and PaymentCheckoutTest, similar to how the CheckoutPaneTest in commerce_shipping confirms that the right countries are used.

bojanz:
Finished and committed in a7d02d792779fd647c21ce02493d20b0c77d0ae7.

PR #621, opened 2017-01-30
Adds machine name into attribute list and selection

b-sharpe:
See https://www.drupal.org/node/2842961


modules/product/src/Form/ProductVariationTypeForm.php
```diff
@@ -103,7 +103,7 @@ public function form(array $form, FormStateInterface $form_state) {
     $attributes = $this->entityTypeManager->getStorage('commerce_product_attribute')->loadMultiple();
     $attribute_options = array_map(function ($attribute) {
       /** @var \Drupal\commerce_product\Entity\ProductAttributeInterface $attribute */
-      return $attribute->label();
+      return $this->t("%label => @id", array("%label" => $attribute->label(), "@id" => $attribute->id()));
```
mglaman, line 106: Use short syntax for array, sub `"` for `'`

mglaman (changes requested):

bojanz:
We went another way in https://www.drupal.org/node/2842961

PR #622, opened 2017-01-30
Issue #2847862: Views support for adjustments

steveoliver:
See https://www.drupal.org/node/2847862

- Changes adjustment field schema to save type
- Adds views filter to filter by type
- Adds field formatter that also allows filtering by type (this is necessary for creating views with a column for each type):
screen shot 2017-01-30 at 9 15 01 am


bojanz:
Is the views filter worth having to write that update hook?
Adding a column to an existing field is a non-trivial task, as evidenced by the update code in address module.

Your formatter approach won't work, because it only takes into account order-level adjustments. Instead, we should have a views field that takes adjustments from $order->collectAdjustments() and then filters & renders them.

bojanz:
https://www.drupal.org/project/commerce/issues/2847862#comment-12845976

PR #623, opened 2017-01-31
Issue #2848519: Grant all view access to logs.

steveoliver:
See https://www.drupal.org/node/2848519.

### DIFF
--- a/modules/log/commerce_log.module
+++ b/modules/log/commerce_log.module
@@ -5,6 +5,9 @@
  * Provides activity logs for Commerce entities.
  */
 
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Entity\EntityInterface;
+
 /**
  * Implements hook_preprocess_commerce_order().
  */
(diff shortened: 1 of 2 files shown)


bojanz:
What do you say we just use Drupal\commerce\EmbeddedEntityAccessControlHandler for logs?

See the docblock:
```
/**
 * Provides an access control handler for embedded entities.
 *
 * Embedded entities (product variations, order items) are always managed in
 * the context of their parent entity (product, order). For simplicity's sake,
 * we always grant access to the embedded entity, assuming that the parent
 * access control handling has already happened.
 */
class EmbeddedEntityAccessControlHandler extends CoreEntityAccessControlHandler {
```
I think that logic applies here as well, at least initially. 
If you can view the order in admin (with customer information and everything), then the logs are safe as well. We can raise this to an admin permission if people end up disagreeing.

steveoliver:
Makes sense. Now using EmbeddedEntityAccessControlHandler.

PR #631, opened 2017-02-07
Profile selection

czigor:
Tests are still needed. 
https://www.drupal.org/node/2844920

### DIFF
--- a/modules/order/src/Element/ProfileSelect.php
+++ b/modules/order/src/Element/ProfileSelect.php
@@ -3,10 +3,15 @@
 namespace Drupal\commerce_order\Element;
 
 use Drupal\commerce\Element\CommerceElementTrait;
+use Drupal\Component\Utility\Html;
+use Drupal\Component\Utility\NestedArray;
 use Drupal\Core\Entity\Entity\EntityFormDisplay;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Render\Element;
 use Drupal\Core\Render\Element\RenderElement;
 use Drupal\profile\Entity\ProfileInterface;
+use Drupal\profile\Entity\Profile;
+use Symfony\Component\HttpFoundation\Request;
 
 /**
  * Provides a form element for selecting a customer profile.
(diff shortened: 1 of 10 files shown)


modules/order/src/Element/ProfileSelect.php
```diff
@@ -84,6 +91,80 @@ public static function processForm(array $element, FormStateInterface $form_stat
...
+        ->condition('uid', $profile_uid)
+        ->condition('type', $bundle)
+        ->condition('status', TRUE)
+        ->sort('profile_id', 'DESC')
+        ->execute();
+      $profiles = Profile::loadMultiple($profile_ids);
+      $profile_options = [];
+      /** @var Profile $profile_option */
+      foreach ($profiles as $profile_option) {
+        if (empty($default_profile_id)) {
```
haringsrob: This will always be true as we set it empty on line 123

modules/order/src/Element/ProfileSelect.php
```diff
@@ -84,6 +91,80 @@ public static function processForm(array $element, FormStateInterface $form_stat
...
+          $default_profile_id = $profile_option->id();
+          $default_profile = $profile_option;
+        }
+        $profile_options[$profile_option->id()] = $profile_option->label();
+      }
+      $profile_options['new_address'] = t('+ Enter a new address');
```
haringsrob: Should the + be part of the translatable string?

modules/order/src/Element/ProfileSelect.php
```diff
@@ -84,6 +91,80 @@ public static function processForm(array $element, FormStateInterface $form_stat
...
+      $profile_ids = \Drupal::service('entity.query')
+        ->get('profile')
+        ->condition('uid', $profile_uid)
+        ->condition('type', $bundle)
+        ->condition('status', TRUE)
+        ->sort('profile_id', 'DESC')
+        ->execute();
+      $profiles = Profile::loadMultiple($profile_ids);
+      $profile_options = [];
+      /** @var Profile $profile_option */
```
haringsrob: I think this should include the full namespace.

modules/order/src/Element/ProfileSelect.php
```diff
@@ -130,9 +294,102 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+  }
+
+  /**
+   * Form AJAX callback.
+   *
+   * @param array $form
+   *   The complete form array.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   * @param Symfony\Component\HttpFoundation\Request $request
```
haringsrob: Missing \ before Symfony

modules/order/src/Element/ProfileSelect.php
```diff
@@ -130,9 +294,102 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+   *   The Request object.
+   *
+   * @return array
+   *   The form element replace the wrapper with.
+   */
+  public static function profileSelectAjax(array &$form, FormStateInterface &$form_state, Request $request) {
```
haringsrob: No need to pass be reference if I am correct.

haringsrob: For the $form_state at least

modules/order/src/Element/ProfileSelect.php
```diff
@@ -130,9 +294,102 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+    $triggering_element = $form_state->getTriggeringElement();
+    $array_parents = $triggering_element['#array_parents'];
+    array_pop($array_parents);
+    $element = NestedArray::getValue($form, $array_parents);
+    return $element;
```
haringsrob: I think we can return directly without first assigning it.

modules/order/src/Element/ProfileSelect.php
```diff
@@ -130,9 +294,102 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+    $storage = $form_state->getStorage();
+    if (in_array($storage['pane_' . $pane_id]['mode'], ['new', 'edit'])) {
+      $form_display = EntityFormDisplay::collectRenderDisplay($element['#profile'], 'default');
+      $form_display->extractFormValues($element['#profile'], $element, $form_state);
+      $element['#profile']->save();
+    }
+  }
+
+  /**
+   * Form AJAX callback.
```
haringsrob: Commerce profile AJAX form callback? Just a suggestion

modules/order/src/Element/ProfileSelect.php
czigor: Yes, this enables us to include/exclude + in the translation.

czigor:
Thanks, @haringsrob, I have fixed the issues.

modules/order/src/Element/ProfileSelect.php
```diff
@@ -100,6 +183,85 @@ public static function processForm(array $element, FormStateInterface $form_stat
...
     }
 
+    $called_class = get_called_class();
+    if (!empty($profile_uid) && $mode != 'edit' && !empty($profile_options)) {
+      $element['profile_selection'] = [
+        '#title' => t('Select an address'),
+        '#options' => $profile_options,
+        '#type' => 'select',
+        '#weight' => -5,
+        '#default_value' => $default_profile_id,
```
ReplicaObscura: $default_profile_id is not defined here if the user doesn't have any profiles created. Setting it to NULL initially in this function in case it doesn't get redefined seems to work fine for me.

czigor: Fixed, thanks!

ReplicaObscura:
I've got a combined PR containing everything from this PR plus #714, submitted as #715.

Just wanted to explain what that PR is here. I'm not intending to take any credit for your work, and I will make sure to pull in any further fixes you make here as well. However #714 modifies much of the same code as this and applying both patches together is not possible, so I thought it might be easier to have one PR that has both features as an option.

mglaman:
Profile select tests are now passing. Working on fixing failing tests. Discovered that the profile is being saved on all AJAX :(

mglaman:
Need to review IEF, handle saving like that.

mglaman:
Things remaining

- [x] Fix AdminOrderTest
- [x] Fix PaymentMethodTest
- [x] Fix CheckoutOrderTest
- [x] Fix PaymentCheckoutTest
- [ ] Prevent saving profile during AJAX operations, only final form submit (PITA.)
- [x] Verify default value respected.

mglaman:
Something about the checkout pane doesn't set values.

order_information___drupal_and_billinginformation_php_-_commerce2x_-____drupal_sites_commerce2x_


EDIT: JK! This is about default values.

bojanz:
Continued in #760. We're almost there!

PR #633, opened 2017-02-10
Show payment gateway in user Payment methods listing

maciejzgadzaj:
See https://www.drupal.org/node/2851488


bojanz:
We're exposing an admin label to users. That won't do.

We can take the display label from the gateway plugin ($gateway->getPlugin()->getDisplayLabel()) but I'm interested in why you'd want to show this in the first place. Doesn't make sense to me.

maciejzgadzaj:
Just for information purposes. Many a time I have been wondering which gateway a payment belongs to, and besides looking into its devel dump, there is no way to check it easily. This small change would help, though it's true that it's targeted more at the development/testing scenarios.

Updated display label coming in a sec.

bojanz:
Since this is developer-oriented, but on a customer-oriented screen, decided not to merge it.

PR #634, opened 2017-02-11
Issue #2831914: PaymentStorage should provide a method to load an orders payments

mitrpaka:
See https://www.drupal.org/node/2831914


bojanz:
Finished and committed in 239a97336dd2707b126e2d15a6bd6e5792a3f1c3.

PR #636, opened 2017-02-14
Issue #2852363: Provide a Views default argument for the current user's cart ID

ReplicaObscura:
Here's a PR for the patch from here: https://www.drupal.org/node/2852363#comment-11933905


ReplicaObscura:
I'm not sure why the tests would be failing. I haven't added any new tests, but I also haven't modified any code that should break any existing tests I wouldn't think.

f1rf1n:
I used the patch from the D.O. issue. Great result! Test failure might be from the whitespace error maybe? Just a blind stab, thank you for this anyway!

Patch from #5 works brilliantly for me. I can hide the empty cart now. Thanks @bmcclure!

Nitpick: whitespace error?

> <code> git apply -v commerce-Views_argument_default_current_cart-2852363-5.patch 
> Checking patch modules/cart/src/Plugin/views/argument_default/CurrentCart.php...
> commerce-Views_argument_default_current_cart-2852363-5.patch:122: new blank line at EOF.
> +
> Applied patch modules/cart/src/Plugin/views/argument_default/CurrentCart.php cleanly.
> warning: 1 line adds whitespace errors.</code>

calbasi:
Another issue that could be fixed with this PR: [Option to hide empty cart block](https://www.drupal.org/project/commerce/issues/3038346)

bojanz:
Let's continue on the issue. Thanks, everyone!

PR #637, opened 2017-02-16
Issue #2851477: Reset cart: Tests and method call

haringsrob:
I have used a specific css selector because I am not sure if the strings in the cart/other parts of the flow would ever change.

I felt it was not reliable to use Review on its own.

### DIFF
--- a/modules/cart/src/CartManager.php
+++ b/modules/cart/src/CartManager.php
@@ -71,8 +71,8 @@ class CartManager implements CartManagerInterface {
     $cart->setAdjustments([]);
 
     $this->eventDispatcher->dispatch(CartEvents::CART_EMPTY, new CartEmptyEvent($cart, $order_items));
+    $this->resetCheckoutStep($cart);
     if ($save_cart) {
-      $this->resetCheckoutFlow($cart);
       $cart->save();
     }
   }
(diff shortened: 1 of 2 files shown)


modules/cart/src/CartManager.php
```diff
@@ -151,8 +154,20 @@ public function removeOrderItem(OrderInterface $cart, OrderItemInterface $order_
...
       $cart->save();
     }
   }
 
+  /**
+   * Resets the checkout flow status.
+   *
+   * @param OrderInterface $cart
+   */
+  private function resetCheckoutFlow(OrderInterface $cart) {
```
bojanz: 1) We use protected instead of private.
2) The @param must have a FQCN instead of just OrderInterface, and it needs a description
3) You are reseting the checkoutStep, not the checkoutFlow
4) Your if() will crash if commerce_checkout is not enabled. Instead you need $cart->hasField('checkout_step')

modules/cart/src/CartManager.php
```diff
@@ -123,6 +124,7 @@ public function addOrderItem(OrderInterface $cart, OrderItemInterface $order_ite
     }
 
     if ($save_cart) {
+      $this->resetCheckoutFlow($cart);
```
bojanz: Let's perform the reset before the if(), that way the step is reset even if saving the cart is skipped.

haringsrob:
Thanks @bojanz, makes perfect sense..

bojanz:
Committed in cbcdb3ef3202d79ec46c653f03d2ae7f36545d0e.

PR #638, opened 2017-02-17
#2827144 Support multiple payment gateways on Checkout Payment information pane

vasike:


sumanthkumarc:
After the recent commit at http://cgit.drupalcode.org/commerce/commit/?h=8.x-2.x&id=9f451cddd49e56ac06aca433d0fbca0786a36af8, this patch doesn't apply and needs re roll.

bojanz:
Continued in #640.

PR #639, opened 2017-02-20
fix #2853527 - implement cart expiration

rthornton128:


modules/cart/commerce_cart.module
```diff
@@ -223,3 +247,19 @@ function commerce_cart_views_data_alter(array &$data) {
...
 }
+
+/**
+ * Implements hook_cron().
+ */
+function commerce_cart_cron() {
+  $queue = \Drupal::queue('cart_expiries');
+
+  $ids = \Drupal::entityQuery('commerce_order')
+    ->notExists('completed')
```
mglaman: Since we're putting this in cart, we need to check for things flagged a cart (not just drafts.) So let's change this to `->condition('cart', TRUE)`

modules/cart/commerce_cart.module
```diff
@@ -154,6 +154,7 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
   $order_type = $form_state->getFormObject()->getEntity();
   $cart_form_view = $order_type->getThirdPartySetting('commerce_cart', 'cart_form_view', 'commerce_cart_form');
   $cart_block_view = $order_type->getThirdPartySetting('commerce_cart', 'cart_block_view', 'commerce_cart_block');
+  $cart_expiry = $order_type->getThirdPartySetting('commerce_cart', 'cart_expiry');
```
mglaman: Any reason to not just go `cart_expiration`

modules/cart/commerce_cart.module
```diff
@@ -223,3 +247,19 @@ function commerce_cart_views_data_alter(array &$data) {
 function template_preprocess_commerce_cart_block(&$variables) {
   $variables['attributes']['class'][] = 'cart--cart-block';
 }
+
+/**
+ * Implements hook_cron().
+ */
+function commerce_cart_cron() {
+  $queue = \Drupal::queue('cart_expiries');
```
mglaman: Maybe `cart_expirations`

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,42 @@
...
+namespace Drupal\commerce_cart\Plugin\QueueWorker;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+use Drupal\Core\Queue\QueueWorkerBase;
+
+/**
+ * Removes an expired Cart.
+ *
+ * @QueueWorker(
+ *  id = "cart_expiries",
```
mglaman: see above, maybe, `cart_expirations`

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,42 @@
...
+ *  cron = {}
+ * )
+ */
+class CartExpire extends QueueWorkerBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processItem($data) {
+    $orders = \Drupal::entityTypeManager()->getStorage('commerce_order')
```
mglaman: Let's implement the ContainerFactoryInjectionInterface, or whatever it's called. This allows us to properly inject the service. If you're not too sure, we have an example somewhere within the module.

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,42 @@
...
+      ->loadMultiple($data);
+    foreach ($orders as $order) {
+      // Ensure that orders slated for clearing have not been completed since
+      // they were last queued.
+      if ($order instanceof OrderInterface && is_null($order->getCompletedTime())) {
```
mglaman: Again I think we should be checking the `cart` flag. See how we handle that in the cart block.

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,42 @@
...
+        $order_type_storage = \Drupal::entityTypeManager()
+          ->getStorage('commerce_order_type');
+        $order_type = $order_type_storage->load($order->bundle());
+        $elapsed = REQUEST_TIME - $order->getCreatedTime();
```
mglaman: We actually have a `time` service! Because REQUEST_TIME can't be trusted in long running ops (CLI). It's provided by core in 8.3.x, but we made our own. You can find it in the commerce.services.yml

mglaman (changes requested):
My biggest concern is using expiry vs expiration in full. Just for non-native English speakers.

modules/cart/commerce_cart.module
```diff
@@ -188,6 +189,26 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
...
+  $form['commerce_cart']['enable_expiration'] = [
+    '#type' => 'checkbox',
+    '#title' => t('Enable cart expiration for this order type'),
+    '#default_value' => !empty($cart_expiry),
+  ];
+
+  $form['commerce_cart']['cart_expiration'] = [
+    '#type' => 'number',
+    '#title' => t('Days until cart expires'),
+    '#description' => t('The cart will be removed in the number of days specified. Set to 0 disable cart expiry.'),
```
mglaman: s/expiry/expiration

modules/cart/commerce_cart.module
```diff
@@ -223,3 +247,19 @@ function commerce_cart_views_data_alter(array &$data) {
...
+ */
+function commerce_cart_cron() {
+  $queue = \Drupal::queue('cart_expirations');
+
+  $ids = \Drupal::entityQuery('commerce_order')
+    ->condition('cart', TRUE)
+    ->execute();
+
+  // Queue ids in groups of 50.
+  for ($i = 0; $i < count($ids); $i += 50) {
```
mglaman: At first, I didn't get this, but now I do. You're putting groups of 50 cart orders into multiple queue items.  

Perhaps we should just put `->range(0, 50)` on the query and just purge 50 at a time. Out of the box, Drupal's automated_cron should run every three hours. I'm sure most people configure it to run more via a command line operation.

@bojanz thoughts? should we just create one queue item for up to 50 orders per run?

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,96 @@
...
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   * @param \Drupal\commerce\TimeInterface $time
+   *   The time service.
+   */
+  public function __construct(array $configuration,
+                              $plugin_id,
```
mglaman: I know PhpStorm can do this, but regardless length the constructor stays on same line

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * The commerce time service.
+   *
+   * @var \Drupal\commerce\Time
+   */
+  protected $time;
+
+  /**
+   * Constructs a new CartBlock.
```
mglaman: We follow the `Constructs a new CLASSNAME object` block here.

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,96 @@
...
+      $container->get('entity_type.manager'),
+      $container->get('commerce.time')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processItem($data) {
+    $orders = $this->entityTypeManager->getStorage('commerce_order')
```
mglaman: Instead of just having the entity type manager as a class property, let's make it be the order storage, then you can just invoke `$this->orderStorage`

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,96 @@
...
+      ->loadMultiple($data);
+    foreach ($orders as $order) {
+      // Ensure that orders slated for clearing have not been completed since
+      // they were last queued.
+      if ($order instanceof OrderInterface && $order->cart->value == 1) {
+        $order_type_storage = $this->entityTypeManager
```
mglaman: Instead of just having the entity type manager as a class property, let's make it be the order type storage, then you can just invoke `$this->orderTypeStorage`

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,96 @@
...
+          ->getStorage('commerce_order_type');
+        $order_type = $order_type_storage->load($order->bundle());
+        $elapsed = $this->time->getCurrentTime() - $order->getCreatedTime();
```
mglaman: We should be checked last updated time. A cart could have been created a while ago.. but still active.

modules/cart/src/Plugin/QueueWorker/CartExpire.php
```diff
@@ -0,0 +1,96 @@
...
+        $expiry = $order_type->getThirdPartySetting('commerce_cart', 'cart_expiration') * 3600 * 24;
```
mglaman: We say `0` disables, but with this check.. wouldn't it always delete?

modules/cart/commerce_cart.module
mglaman: Actually, I have an idea which completely changes the queue worked.

1. Load all possible order types.
1. Check if the order type has cart expiration configured
1. If cart expiration is configured, query orders that have `cart => 1` and `updated < (time() - EXPIRATION_TIME`
1. Pass all of the returned order IDs into a new queue item

The result is a queue entry per order type with all the valid IDs. Perhaps we should also pass in the order type and settings, along with the IDs. So we can easily re-validate.

mglaman (changes requested):
One more: See https://github.com/drupalcommerce/commerce/blob/8.x-2.x/modules/cart/config/schema/commerce_cart.schema.yml#L5

We need to add to the `commerce_order.commerce_order_type.*.third_party.commerce_cart` schema definition to add this value, as an integer.

mglaman (approved):

modules/cart/commerce_cart.module
```diff
@@ -223,3 +247,34 @@ function commerce_cart_views_data_alter(array &$data) {
...
+  $time = \Drupal::service('commerce.time');
+  $order_storage = \Drupal::entityTypeManager()->getStorage('commerce_order');
+  $type_storage = \Drupal::entityTypeManager()->getStorage('commerce_order_type');
+  $order_types = $type_storage->loadMultiple();
+
+  foreach ($order_types as $order_type) {
+    $expiration_days = $order_type->getThirdPartySetting('commerce_cart', 'cart_expiration');
+
+    if ($expiration_days > 0) {
+      $expiration = $time->getCurrentTime() - ($expiration_days * 3600 * 24);
```
mglaman: Lets change the storage to assume we've saved the seconds offset.

mglaman (changes requested):

mglaman (approved):

modules/cart/commerce_cart.module
```diff
@@ -191,6 +192,26 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
...
+    '#type' => 'checkbox',
+    '#title' => t('Enable cart expiration for this order type'),
+    '#default_value' => !empty($cart_expiration),
+  ];
+
+  $form['commerce_cart']['cart_expiration'] = [
+    '#type' => 'number',
+    '#title' => t('Days until cart expires'),
+    '#description' => t('The cart will be removed in the number of days specified. Set to 0 disable cart expiration.'),
+    '#default_value' => $cart_expiration ?: 0,
```
mglaman, line 205: We don't convert the seconds back into days, here.

modules/cart/commerce_cart.module
```diff
@@ -191,6 +192,26 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
...
+    '#required' => TRUE,
```
shabananavas, line 206: It would be nice if we could add a '#min' => 0, here so that the user can't set a negative value.

modules/cart/commerce_cart.module
```diff
@@ -191,6 +192,26 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
...
+  $form['commerce_cart']['enable_expiration'] = [
+    '#type' => 'checkbox',
+    '#title' => t('Enable cart expiration for this order type'),
+    '#default_value' => !empty($cart_expiration),
+  ];
+
+  $form['commerce_cart']['cart_expiration'] = [
+    '#type' => 'number',
+    '#title' => t('Days until cart expires'),
+    '#description' => t('The cart will be removed in the number of days specified. Set to 0 disable cart expiration.'),
```
shabananavas, line 204: Typo here, it should be: 

'#description' => t('The cart will be removed in the number of days specified. Set to 0 **to** disable cart expiration.'),

shabananavas:
I have added a couple of comments/suggestions above. But otherwise, this patch is working as intended. I have applied the patch and can confirm the setting of an expiration is working correctly and it is deleting the cart items that have dates past the expiration time on Cron runs.

bojanz:
A newer version of the code is available on the issue, closing the PR.

PR #646, opened 2017-02-23
Issue #2855104 by mglaman, niko-: Orders should have a field containi…

mglaman:
…ng the coupons that have been redeemed


mglaman:
https://www.drupal.org/node/2855104

bojanz:
Looks good.

EDIT: Actually, we need an update hook that creates the field for people who already have Promotion enabled.

mglaman:
@bojanz okay, so we need to add to post_update to run entity definition changes

PR #648, opened 2017-02-23
Issue #2855119 by mglaman, niko-: Use a separate order processor for …

mglaman:
…coupon based promotions

https://www.drupal.org/node/2855119

### DIFF
--- a/modules/promotion/src/CouponOrderProcessor.php
+++ b/modules/promotion/src/CouponOrderProcessor.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Drupal\commerce_promotion;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+use Drupal\commerce_order\OrderProcessorInterface;
+use Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\PromotionOfferInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+
+/**
+ * Applies coupon promotions to orders during the order refresh process.
+ *
+ * @see \Drupal\commerce_promotion\PromotionOrderProcessor
+ */
+class CouponOrderProcessor implements OrderProcessorInterface {
+
+  /**
+   * The promotion storage.
+   *
+   * @var \Drupal\commerce_promotion\PromotionStorageInterface
+   */
+  protected $promotionStorage;
+
+  /**
+   * The order type storage.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $orderTypeStorage;
+
+  /**
+   * The coupon storage.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $couponStorage;
+
+  /**
+   * Constructs a new CouponOrderProcessor object.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager) {
+    $this->promotionStorage = $entity_type_manager->getStorage('commerce_promotion');
+    $this->orderTypeStorage = $entity_type_manager->getStorage('commerce_order_type');
+    $this->couponStorage = $entity_type_manager->getStorage('commerce_promotion_coupon');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function process(OrderInterface $order) {
+    if (!$order->hasField('coupons') || $order->get('coupons')->isEmpty()) {
+      return;
+    }
+
+    $order_type = $this->orderTypeStorage->load($order->bundle());
+    /** @var \Drupal\commerce_promotion\Entity\CouponInterface[] $coupons */
+    $coupons = $order->get('coupons')->referencedEntities();
+    foreach ($coupons as $index => $coupon) {
+      /** @var \Drupal\commerce_promotion\Entity\PromotionInterface $promotion */
+      $promotion = $this->promotionStorage->loadByCoupon($order_type, $order->getStore(), $coupon);
+
+      // The promotion may have become invalid (inactive/expired), causing the
+      // query in loadByCoupon() to filter it out.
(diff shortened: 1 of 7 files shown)


modules/promotion/src/PromotionOrderProcessor.php
```diff
@@ -43,6 +43,11 @@ public function process(OrderInterface $order) {
     $order_type = $this->orderTypeStorage->load($order->bundle());
     $promotions = $this->promotionStorage->loadValid($order_type, $order->getStore());
     foreach ($promotions as $promotion) {
+      // Coupons have a separate order processor.
+      if (!$promotion->get('coupons')->isEmpty()) {
```
bojanz: Shouldn't we instead ensure that $this->promotionStorage->loadValid() doesn't return promotions that have coupons? We can filter them out at the query level, skipping the loading of potentially dozens of entities.

mglaman:
@bojanz yeah I was wondering if we should add a flag of `$with_coupons = FALSE` to the method. But I suppose restricting anything with coupons makes sense, because nothing with a coupon would be valid unless the coupon was applied. I'll add a `noexists` check.

mglaman:
Also, let's not merge this until #646 is in, because it copies its code, but missing update.

mglaman:
This failure is a test only issue :/ manually running proves it works.

modules/promotion/src/CouponOrderProcessor.php
```diff
@@ -0,0 +1,85 @@
...
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $orderTypeStorage;
+
+  /**
+   * The coupon storage.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $promotionCouponStorage;
```
bojanz: Why not just call it $couponStorage? This feels too verbose

modules/promotion/src/CouponOrderProcessor.php
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Drupal\commerce_promotion;
+
+use Drupal\commerce_order\Entity\OrderInterface;
+use Drupal\commerce_order\OrderProcessorInterface;
+use Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\PromotionOfferInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+
+/**
+ * Applies promotions to orders during the order refresh process.
```
bojanz: Wrong docblock. Let's do "Applies coupon promotions to orders during the order refresh process." and a @see to the PromotionOrderProcessor. We also want PromotionOrderProcessor to have a @see to this one.

modules/promotion/src/CouponOrderProcessor.php
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * The coupon storage.
+   *
+   * @var \Drupal\Core\Entity\EntityStorageInterface
+   */
+  protected $promotionCouponStorage;
+
+  /**
+   * Constructs a new PromotionOrderProcessor object.
```
bojanz: Wrong class name.

modules/promotion/src/CouponOrderProcessor.php
```diff
@@ -0,0 +1,85 @@
...
+      return;
+    }
+
+    $order_type = $this->orderTypeStorage->load($order->bundle());
+    /** @var \Drupal\commerce_promotion\Entity\CouponInterface[] $coupons */
+    $coupons = $order->get('coupons')->referencedEntities();
+    foreach ($coupons as $index => $coupon) {
+      /** @var \Drupal\commerce_promotion\Entity\PromotionInterface $promotion */
+      $promotion = $this->promotionStorage->loadByCoupon($order_type, $order->getStore(), $coupon);
+      if (!$promotion) {
```
bojanz: Needs comment to explain why $promotion could be FALSE.

modules/promotion/src/CouponOrderProcessor.php
```diff
@@ -0,0 +1,85 @@
...
+        $order->get('coupons')->removeItem($index);
+        continue;
+      }
+
+      /** @var \Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\PromotionOfferInterface $plugin */
+      $plugin = $promotion->get('offer')->first()->getTargetInstance();
```
bojanz, line 73: Isn't there a getter to replace $promotion->get('offer')->first()->getTargetInstance()?

modules/promotion/src/PromotionStorage.php
```diff
@@ -42,7 +42,10 @@ protected function loadValidQuery(OrderTypeInterface $order_type, StoreInterface
    * {@inheritdoc}
    */
   public function loadValid(OrderTypeInterface $order_type, StoreInterface $store) {
-    $query = $this->loadValidQuery($order_type, $store);
+    $query = $this->loadQuery($order_type, $store);
+    // We do not want promotions with coupons to be loaded in this query,
```
bojanz: I don't understand this comment. Perhaps something like "Only load promotions without coupons. Promotions with coupons are loaded coupon-first in a different process." ?

modules/promotion/src/PromotionStorage.php
```diff
@@ -23,7 +23,7 @@ class PromotionStorage extends CommerceContentEntityStorage implements Promotion
    * @return \Drupal\Core\Entity\Query\QueryInterface
    *   The entity query.
    */
-  protected function loadValidQuery(OrderTypeInterface $order_type, StoreInterface $store) {
+  protected function loadQuery(OrderTypeInterface $order_type, StoreInterface $store) {
```
bojanz: While we're at it, protected methods should go after the public ones.

bojanz (changes requested):

modules/promotion/src/CouponOrderProcessor.php
mglaman: Nope. We should add a `getOfferPlugin` to interface.

PR #649, opened 2017-02-23
Issue #2656818 by vasike, smccabe, bojanz: Implement optimistic locking for orders

vasike:


modules/cart/commerce_cart.module
```diff
@@ -108,6 +109,55 @@ function commerce_cart_commerce_order_delete(OrderInterface $order) {
...
+  }
+}
+
+/**
+ * Validate handler for Views commerce cart forms.
+ */
+function commerce_cart_commerce_cart_form_validate($form, FormStateInterface $form_state) {
+  // Re-load the order.
+  /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
+  $order = Order::load($form_state->get('order_id'));
```
mglaman: Should we use `loadUnchanged` from storage to ensure it's fresh?

modules/cart/commerce_cart.module
```diff
@@ -138,6 +188,10 @@ function commerce_cart_form_commerce_order_form_alter(array &$form, FormStateInt
       // Move the cart element to the bottom of the meta sidebar container.
       $form['cart']['#group'] = 'meta';
       $form['cart']['#weight'] = 100;
+      // Warning it the order is cart and it was modified in the last hour.
+      if (!empty($order->get('cart')->value) && (REQUEST_TIME - $order->getChangedTime()) < 3600) {
```
mglaman: Time service

modules/order/src/Exception/ExceptionInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Drupal\commerce_order\Exception;
+
+/**
+ * Exception interface for all exceptions thrown by the Order module.
+ */
+interface ExceptionInterface {}
```
mglaman: OrderExceptionInterface?

modules/order/src/Form/OrderForm.php
```diff
@@ -61,6 +61,11 @@ public function form(array $form, FormStateInterface $form_state) {
       '#type' => 'hidden',
       '#default_value' => $order->getChangedTime(),
     ];
+    // Version must be sent to the client, for later overwrite error checking.
+    $form['version'] = [
+      '#type' => 'hidden',
```
mglaman, line 74: Can't we use #value?

modules/cart/commerce_cart.module
vasike: indeed. i'll do

modules/cart/commerce_cart.module
vasike: i did intentionally as i think soon we'll switch to 8.3 and use core time.
however i'll switch to Commerce Time till then. thanks

modules/order/src/Exception/ExceptionInterface.php
vasike: agree

modules/order/src/Form/OrderForm.php
vasike: it seems it won't work with value. we need to send it to client.

bojanz:
An updated patch is being posted to the issue.

PR #650, opened 2017-02-24
Issue #2855435 by vasike, bojanz: Create a custom purchased_entity views field plugin

vasike:
…


bojanz:
Let's continue on the issue.

PR #651, opened 2017-02-24
Issue #2855762 by mglaman: Create a coupon redemption element

mglaman:


modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides a form element for embedding the coupon redemption form.
+ *
+ * Usage example:
+ * @code
+ * $form['coupons'] = [
+ *   '#type' => 'commerce_promotion_coupon_redemption_form',
+ *   // An order entity that valid redeemed coupons will be applied to.
```
bojanz: 'that valid' doesn't sound valid
"The order to which the coupon will be applied to."

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+ *   '#order' => $order,
+ *   // Set to FALSE to not use AJAX submission
```
bojanz: #use_ajax was removed, so that + the commend above need sto go.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+ *   '#use_ajax' => FALSE,
+ * ];
+ * @endcode
+ *
+ * @RenderElement("commerce_promotion_coupon_redemption_form")
```
bojanz: Why not call the element "commerce_coupon_redemption_form"? It's shorter.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+ */
+class CouponRedemptionForm extends CommerceElementBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getInfo() {
+    $class = get_class($this);
+    return [
+      // The order which redeemed coupons will apply to.
```
bojanz: Let's use the same comment as in the (updated) example.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+   *
+   * @param array $element
+   *   The form element being processed.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   * @param array $complete_form
+   *   The complete form structure.
+   *
+   * @throws \InvalidArgumentException
+   *   Thrown when the #operation or #default_value properties are empty, or
```
bojanz: Untrue, you're not validating #operation nor #default_value.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+   *   when the #default_value property is not a valid entity.
+   *
+   * @return array
+   *   The processed form element.
+   */
+  public static function processForm(array $element, FormStateInterface $form_state, array &$complete_form) {
+    if (!$element['#order'] instanceof OrderInterface) {
```
bojanz: We can throw an \InvalidArgumentException here, makes more sense than exiting silently.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+    $coupon_code = $form_state->getValue($coupon_parents);
+    if (empty($coupon_code)) {
+      return;
+    }
+    $code_path = implode('][', $coupon_parents);
+    /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
+    $order = $form_state->get('order');
+    $entity_type_manager = \Drupal::entityTypeManager();
+
+    // Validate that the coupon code exists.
```
bojanz: Redundant comment.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    foreach ($order->get('coupons') as $item) {
+      if ($item->target_id == $coupon->id()) {
+        $form_state->setErrorByName($code_path, t('Coupon is already redeemed'));
+        return;
+      }
+    }
+
+    // If it does, ensure its promotion is still valid.
```
bojanz: "If it does, ensure its promotion is still valid." is not above the code that actually does it, so I was confused.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+      $form_state->setErrorByName($code_path, t('Coupon is invalid'));
+      return;
+    }
+
+    $promotion_applies = FALSE;
+    if ($promotion->applies($order)) {
+      $promotion_applies = TRUE;
+    }
+    else {
+      foreach ($order->getItems() as $orderItem) {
```
bojanz: $order_item instead of $orderItem

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+        if ($promotion->applies($orderItem)) {
+          $promotion_applies = TRUE;
+          break;
+        }
+      }
+    }
+
+    if ($promotion_applies === FALSE) {
```
bojanz: Why the strict comparison, when you're the one setting the boolean above? You should be able to trust it, no?

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+    /** @var \Drupal\commerce_promotion\PromotionStorageInterface $promotion_storage */
+    $promotion_storage = $entity_type_manager->getStorage('commerce_promotion');
+
+    $coupon = static::loadCouponByCode($coupon_code);
+    /** @var \Drupal\commerce_order\Entity\OrderTypeInterface $order_type */
+    $order_type = $order_type_storage->load($order->bundle());
+    $promotion = $promotion_storage->loadByCoupon($order_type, $order->getStore(), $coupon);
+
+    if ($promotion->applies($order)) {
+      $order->get('coupons')->appendItem($coupon);
```
bojanz: Looks like we're missing a helper, since we're repeating the applies logic in two places. We could at least have one on the form element (then think of where to move it in a followup)

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+    // Validate that the coupon code exists.
+    $coupon = static::loadCouponByCode($coupon_code);
+    if (empty($coupon)) {
+      $form_state->setErrorByName($code_path, t('Coupon is invalid'));
+      return;
+    }
+
+    foreach ($order->get('coupons') as $item) {
+      if ($item->target_id == $coupon->id()) {
+        $form_state->setErrorByName($code_path, t('Coupon is already redeemed'));
```
bojanz: s/is already/has already been?

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+   *   The coupon code.
+   *
+   * @return \Drupal\commerce_promotion\Entity\CouponInterface
+   *   The coupon.
+   *
+   * @todo Finish https://www.drupal.org/node/2855815, use that.
+   * @link https://www.drupal.org/node/2855815
+   */
+  protected static function loadCouponByCode($coupon_code) {
+    $couponStorage = \Drupal::entityTypeManager()->getStorage('commerce_promotion_coupon');
```
bojanz: $coupon_storage instead of $couponStorage.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+      }
+    }
+
+    \Drupal::service('commerce_order.order_refresh')->refresh($order);
+    $order->save();
+    drupal_set_message(t('Coupon applied'));
+  }
+
+  /**
+   * Load a coupon by its code.
```
bojanz: s/Load/Loads

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+    }
+    else {
+      foreach ($order->getItems() as $orderItem) {
+        if ($promotion->applies($orderItem)) {
+          $order->get('coupons')->appendItem($coupon);
+        }
+      }
+    }
+
+    \Drupal::service('commerce_order.order_refresh')->refresh($order);
```
bojanz: $order->save() should be refreshing the order. This sounds like we're refreshing an order right before it's refreshed again.

bojanz (changes requested):

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -0,0 +1,225 @@
...
+   * @param string $coupon_code
+   *   The coupon code.
+   *
+   * @return \Drupal\commerce_promotion\Entity\CouponInterface
+   *   The coupon.
+   *
+   * @todo Finish https://www.drupal.org/node/2855815, use that.
+   * @link https://www.drupal.org/node/2855815
+   */
+  protected static function loadCouponByCode($coupon_code) {
```
mglaman: Remove now that we have #653

PR #652, opened 2017-02-24
Issue 2855587: Investigate why DrupalCI fails on checkout based Funct…

drugan:
…ional and FunctionalJavascript tests

BrowserTestBase->drupalGet() takes as the first argument relative path or absolute path or Url object.

But seems that relative path fails in some environments to be resolved to proper url.

### DIFF
--- a/.travis-script-simpletest-js.sh
+++ b/.travis-script-simpletest-js.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+# @file
+# Simple script to run the tests via travis-ci.
+
+set -e $DRUPAL_TI_DEBUG
+
+export ARGS=( $DRUPAL_TI_SIMPLETEST_JS_ARGS )
+
+if [ -n "$DRUPAL_TI_SIMPLETEST_GROUP" ]
+then
+        ARGS=( "${ARGS[@]}" "$DRUPAL_TI_SIMPLETEST_GROUP" )
+fi
+
+
+cd "$DRUPAL_TI_DRUPAL_DIR"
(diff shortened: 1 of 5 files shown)


mglaman:
Hm. Does core have to do this? I'll try to run this patch on Drupal.org later and see what happens.

drugan:
I have local server and my dev site's product url looks like this:

http://mysites/acommerce/web/product/1

But with $this->drupalGet($this->product->toUrl()->toString()); it was like this in a test site:

http://mysites/acommerce/web/acommerce/product/1

Now I've run tests localy with $this->drupalGet($this->product->toUrl()) and they all green.

mglaman:
Oh! Our tests aren't using URL generator? Shoot! Updating to toUrl will fix it. We shouldn't be manually constructing at all

drugan:
Every time the Travis checks the PR it gives up on the PHP 7 version. No errors, no warnings or something indicating a trouble. Just this:

> The job exceeded the maximum time limit for jobs, and has been terminated.

Can we to increase the maximum time limit?

mglaman:
We have 50 minutes for each build. We can't bump it.

mglaman:
The URL fixes are 👍  however, it's not fixing the issues after all :(

mglaman:
See https://www.drupal.org/node/2855587#comment-11957449

drugan:
See https://www.drupal.org/node/2855587#comment-11958049

bojanz:
This needs a rebase. The run_tests.sh improvements were merged previously, and CheckoutOrderTest no longer has the access tests.

mglaman:
Let's just close it. It's only reproducible on Drupal.org and the DrupalCI vagrant.

PR #653, opened 2017-02-27
Issue #2855815: Add custom coupon storage, provide loadByCode method.

mitrpaka:


modules/promotion/src/CouponStorage.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Defines the coupon storage.
+ */
+class CouponStorage extends CommerceContentEntityStorage implements CouponStorageInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function loadByCode($coupon_code) {
```
mglaman: We can just change the parameter to code

mglaman (changes requested):
Thanks! We also need a test

mglaman (approved):

mglaman:
I made the change and added test.

PR #654, opened 2017-02-27
Use run-test.sh runner to add concurrency.

mglaman:


mglaman:
build__3311_-_drupalcommerce_commerce_-_travis_ci

We can be 66% faster

bojanz:
I won't pretend to understand what the shell script does, but speed is speed.

PR #655, opened 2017-02-28
Issue #2790533 by bojanz, vasike: Add the ability to set a default payment method

vasike:


modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentInformation.php
```diff
@@ -309,7 +309,18 @@ protected function getDefaultPaymentMethodOption(array $options) {
     // The order doesn't have a payment method/gateway specified, or it has, but it is no longer available.
     if (!$default_option || !isset($options[$default_option])) {
       $option_ids = array_keys($options);
-      $default_option = reset($option_ids);
+      // Use customer default payment method if available for the current pane.
+      if ($customer = $this->order->getCustomer()) {
+        /** @var \Drupal\commerce_payment\PaymentMethodStorageInterface $payment_method_storage */
+        $payment_method_storage = $this->entityTypeManager->getStorage('commerce_payment_method');
+        $default_payment_method = $payment_method_storage->loadDefaultByUser($this->order->getCustomer());
+        if (in_array($default_payment_method->id(), $option_ids)) {
```
nikathone, line 318: This might cause ` Call to a member function id() on boolean in Drupal\commerce_payment\Plugin\Commerce\CheckoutPane\PaymentInformation->getDefaultPaymentMethodOption()` if the current user doesn't have any payment method. Note when testing my user did have a completed order though. The fix would be something like `if ($default_payment_method->id() && in_array($default_payment_method->id(), $option_ids))`

nikathone: oops! Correction the fix should be `if ($default_payment_method && in_array($default_payment_method->id(), $option_ids))`

bojanz:
There is a newer patch on the issue, let's continue there.

PR #656, opened 2017-03-01
Issue #2856930 by klagler, vasike: Allow to change label/name of product variation title field

vasike:


bojanz:
See https://www.drupal.org/node/2856930#comment-11964893

PR #657, opened 2017-03-02
Issue #2857157: Implement registration after guest checkout

ReplicaObscura:
This relates to https://www.drupal.org/node/2857157

It adds a checkout pane to the last step which allows guests to register an account.

There's a twig template controlling the pane's layout and text.


bojanz:
A newer version of this code is available on the issue, and close to landing. Closing the PR.

PR #658, opened 2017-03-02
Issue #2855838 by mglaman, vasike: Deleting a promotion does not delete its coupons

vasike:


vasike:
new PR that includes this
https://github.com/drupalcommerce/commerce/pull/660

so close it

PR #659, opened 2017-03-02
Issue #2855836 by mglaman, vasike: Coupon codes need unique value validation constraint

vasike:


modules/promotion/tests/src/Kernel/CouponValidationTest.php
```diff
@@ -0,0 +1,60 @@
...
+    // Add other coupon with the same code.
+    $coupon = Coupon::create([
+      'code' => $coupon_code,
+      'status' => TRUE,
+    ]);
+
+    $violations = $coupon->validate();
+    $this->assertEquals(count($violations), 1);
+    $this->assertEquals($violations[0]->getPropertyPath(), 'code');
+    $this->assertEquals($violations[0]->getMessage(), 'A coupon with code %value already exists. Enter a unique code.');
```
mglaman, line 57: Can we update the test to change the code value and run validations again? I'd like the test to prove we can still save the coupon with the validation now added. We have tests that should prove this, but I'd rather be explicit in our test coverage.

mglaman (changes requested):
Code looks good to me. It's pretty much `\Drupal\Core\Validation\Plugin\Validation\Constraint\UniqueFieldValueValidator` but with specific message.

modules/promotion/tests/src/Kernel/CouponValidationTest.php
vasike: done

mglaman: 💃 yay

mglaman (approved):

mglaman:
@bojanz this looks good to me, now. My main Q is just should we use custom validator or reuse core one. But seems better to be verbose.

bojanz:
Committed in 0a7d06cb05855d41f9af595d380841d25a21c695b.

bojanz:
Committed in 0a7d06cb05855d41f9af595d380841d25a21c695b.

PR #661, opened 2017-03-04
Issue #2857357 by vasike, bojanz, Berdir: Replace entity.query usage

vasike:


modules/order/src/Form/OrderTypeDeleteForm.php
```diff
@@ -3,46 +3,18 @@
...
-      $container->get('entity.query')
-    );
-  }
-
-  /**
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
-    $order_count = $this->queryFactory->get('commerce_order')
+    $order_count = $order_item_count = $this->entityTypeManager->getStorage('commerce_order')->getQuery()
```
bojanz: This line is messed up :)

modules/checkout/commerce_checkout.module
```diff
@@ -26,6 +26,9 @@ function commerce_checkout_theme() {
     'commerce_checkout_form__with_summary' => [
       'base hook' => 'commerce_checkout_form',
     ],
+    'commerce_checkout_registration' => [
```
bojanz: Why do we need this?

modules/order/src/Form/OrderTypeDeleteForm.php
vasike: copied too much ;)

modules/checkout/commerce_checkout.module
vasike: no need. from other issue, sorry

PR #662, opened 2017-03-04
Issue #2792885 by plopesc, mglaman: Reference to variations is not re…

mglaman:
…moved when the variation is removed


bojanz:
https://www.drupal.org/project/commerce/issues/2792885#comment-12470139

PR #664, opened 2017-03-04
Issue #2741295 by mglaman, citlacom: ProductVariationStorage::loadEna…

mglaman:
…bled, loadFromContext do not respect product variations access rules


mglaman:
@bojanz I finally updated this, and added tests.

PR #665, opened 2017-03-04
Issue #2857837 by mglaman: Products and variations are bypassing acce…

mglaman:
…ss control

### DIFF
--- a/modules/product/commerce_product.install
+++ b/modules/product/commerce_product.install
@@ -1,0 +1,17 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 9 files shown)


bojanz:
This caused many test explosions. Perhaps because we don't yet have the view published/unpublished permissions & handling?

bojanz:
I'd love to solve https://www.drupal.org/node/2906923 first, then rebase this PR on top of latest Entity API.

That way we avoid bikeshedding twice.

bojanz:
See parent issue.

PR #669, opened 2017-03-14
Allow order types to have no carts

Berdir:
PR for https://www.drupal.org/node/2810723


mglaman:
Ah yeah way simpler to have a skip cart option here than a whole new formatter. Especially for the whole add to cart form, we'd have to completely abstract that otherwise.

bojanz:
Let's continue on the issue, which has the more recent work.

PR #673, opened 2017-03-16
Allow specifying form element  labels and descriptions in coupon redemption element

mglaman:
…ion element

### DIFF
--- a/modules/promotion/src/Element/CouponRedemptionForm.php
+++ b/modules/promotion/src/Element/CouponRedemptionForm.php
@@ -32,6 +32,9 @@ class CouponRedemptionForm extends CommerceElementBase {
     return [
       // The order to which the coupon will be applied to.
       '#order' => NULL,
+      '#title' => t('Coupon code'),
+      '#description' => t('Enter your coupon code here.'),
+      '#submit_title' => t('Apply coupon'),
       '#process' => [
         [$class, 'attachElementSubmit'],
         [$class, 'processForm'],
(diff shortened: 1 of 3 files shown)


modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -32,6 +32,9 @@ public function getInfo() {
     return [
       // The order to which the coupon will be applied to.
       '#order' => NULL,
+      '#code_label' => t('Coupon code'),
+      '#code_description' => t('Enter your coupon code here.'),
+      '#apply_label' => t('Apply coupon'),
```
mglaman: @bojanz bikeshed these labels.

bojanz: I think people would expect the coupon element to inherit the main title and description. Which leaves only the apply button as needing a custom title (#apply_title or #submit_title I guess)

mglaman: >   expect the coupon element to inherit the main title and description.

What do you mean? I imagined the checkout pane would allow for people to customize this default text, such as remove or sub out "coupon" for just "promotion".

When implementing the element I immediately needed to extend the class to make my own element in order to meet client requirements. Which if I had to on first go, I'm assuming 80% of all of our contributors would have to. And most will use a form_alter instead of extending element. I see this as a way to make cleaner code for end users (because checkout pane can provide a UI.)

bojanz: I'm not arguing against the use case at all. I just figure that if I pass a #title to the coupon redemption form element, I'd expect it to be shown above the textarea (Coupon code).

mglaman:
Oh, ha. Duh. Existing #title and #description keys. Went way over my head.
👍 I like. Thanks

PR #674, opened 2017-03-16
Issue #2861002 by mglaman: Coupon redemption should always load a fre…

mglaman:
…sh version of the order

### DIFF
--- a/modules/promotion/src/Element/CouponRedemptionForm.php
+++ b/modules/promotion/src/Element/CouponRedemptionForm.php
@@ -77,7 +77,6 @@ class CouponRedemptionForm extends CommerceElementBase {
     // $wrapper_id = Html::getUniqueId($id_prefix . '-ajax-wrapper');
     $wrapper_id = $id_prefix . '-ajax-wrapper';
 
-    $form_state->set('order', $element['#order']);
     $element = [
       '#tree' => TRUE,
       '#prefix' => '<div id="' . $wrapper_id . '">',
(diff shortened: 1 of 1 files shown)


modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -116,10 +115,10 @@ public static function validateForm(array &$element, FormStateInterface $form_st
     if (empty($coupon_code)) {
       return;
     }
+    $entity_type_manager = \Drupal::entityTypeManager();
     $code_path = implode('][', $coupon_parents);
     /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
-    $order = $form_state->get('order');
-    $entity_type_manager = \Drupal::entityTypeManager();
+    $order = $entity_type_manager->getStorage('commerce_order')->load($element['#order']->id());
```
mglaman, line 121: @bojanz do you want me to bust the order storage into its own var, or is this fine.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -165,9 +164,9 @@ public static function validateForm(array &$element, FormStateInterface $form_st
   public static function submitForm(array &$element, FormStateInterface $form_state) {
     $coupon_parents = array_merge($element['#parents'], ['code']);
     $coupon_code = $form_state->getValue($coupon_parents);
-    /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
-    $order = $form_state->get('order');
     $entity_type_manager = \Drupal::entityTypeManager();
+    /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
+    $order = $entity_type_manager->getStorage('commerce_order')->load($element['#order']->id());
```
mglaman, line 169: See previous.

mglaman (commented):
So this follows like the `commerce_payment_gateway_form` element and takes a full object to stuff into the element properties. We then use that to get the ID and reload it.

Should we change the element to take #order_id as a property.. then validate it by trying to load it? This feels cleaner,and follows existing pattern. Butttttt.. the whole stale object in AJAX thing which uncovered the issue.

modules/promotion/src/Element/CouponRedemptionForm.php
bojanz: I find it more readable to get the storage on its own line.

mglaman:
Closing. There should just be one PR for overall improvements.

PR #675, opened 2017-03-16
Remove PaymentProcess::isVisible(), do not assume that payment_information is the only way to set the payment gateway

Berdir:
https://www.drupal.org/node/2861039


bojanz:
See https://www.drupal.org/node/2861039#comment-11992729

PR #676, opened 2017-03-16
Issue #2861181 : Better handling of customer remote IDs for payments.

vasike:


modules/payment/commerce_payment.post_update.php
```diff
@@ -0,0 +1,77 @@
...
+ * Post update functions for Payment.
+ */
+
+/**
+ * Update user Remote ID field data using payments gateways instead of module.
+ * @see https://www.drupal.org/node/2861181
+ */
+function commerce_payment_post_update_1(&$sandbox) {
+  // Get the user ids that have 'commerce_remote_id' field not emtpy.
+  $result = \Drupal::entityQuery('user')->exists('commerce_remote_id')->execute();
```
mglaman: Wouldn't we only want to run this the first time, and I think we could just do a count query

modules/payment/commerce_payment.post_update.php
```diff
@@ -0,0 +1,77 @@
...
+    $payment_gateway_data = $payment_gateway->toArray();
+    $module = reset($payment_gateway_data['dependencies']['module']);
+    if (!isset($payment_gateways_by_module[$module])) {
+      $payment_gateways_by_module[$module] = $payment_gateway;
+    }
+  }
+
+  // Process entities by groups of 20.
+  $limit = 20;
+  $result = array_slice($result, $sandbox['current'], $limit);
```
mglaman: Why not put a range on the query instead

mglaman (commented):
Just some concerns around the post update function.

modules/payment/commerce_payment.post_update.php
vasike: is it necessary?

modules/payment/commerce_payment.post_update.php
vasike: is it necessary?

modules/payment/commerce_payment.post_update.php
bojanz: A count query would probably be more efficient, let's do that.

vasike: it should be done now

modules/payment/commerce_payment.post_update.php
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @see https://www.drupal.org/node/2861181
+ */
+function commerce_payment_post_update_1(&$sandbox) {
+  // Get the user ids that have 'commerce_remote_id' field not empty.
+  // First check is any user has this field set.
+  if (\Drupal::entityQuery('user')->exists('commerce_remote_id')->count()->execute() == 0) {
+    return t('No user Remote ID field data to update.');
+  }
+  $result = \Drupal::entityQuery('user')->exists('commerce_remote_id')->execute();
```
bojanz: We should run the count query only under if (!isset($sandbox['current'])) {
After that it's cached in $sandbox.

Then run the real query with a range each time.

modules/payment/commerce_payment.post_update.php
```diff
@@ -0,0 +1,82 @@
...
+    $sandbox['max'] = count($result);
+    // A place to store messages during the run.
+  }
+
+  // Get the first payment gateway for every payment module.
+  $payment_gateways_by_module = [];
+  /** @var \Drupal\commerce_payment\PaymentGatewayStorageInterface $payment_gateway_storage */
+  $payment_gateway_storage = \Drupal::entityTypeManager()->getStorage('commerce_payment_gateway');
+  $payment_gateways = $payment_gateway_storage->loadMultiple();
+  uasort($payment_gateways, [$payment_gateway_storage->getEntityType()->getClass(), 'sort']);
```
bojanz: You should be able to use $payment_gateway_storage->getEntityType()->getProvider() to get the module.

modules/payment/src/Entity/PaymentMethod.php
```diff
@@ -224,6 +224,29 @@ public function setCreatedTime($timestamp) {
   /**
    * {@inheritdoc}
    */
+  public function getOwnerRemoteId() {
```
bojanz: Is there a point in having these methods? We usually need the remote customer ID before creating a payment method, so we need them on PaymentGatewayBase instead.

bojanz (changes requested):

modules/payment/src/Entity/PaymentMethod.php
vasike: i also thought the same initially. but working on it looks like PaymentMethod entity methods

modules/payment/commerce_payment.post_update.php
vasike: should be done

modules/payment/commerce_payment.post_update.php
vasike: no idea, what you mean

bojanz:
Committed in 14786dd4afa8e94e5246199759e10ff19ac0f348.

PR #679, opened 2017-03-17
Issue #2861663 by mglaman: Activity view only shows 5 logs, and can b…

mglaman:
…e out of order

### DIFF
--- a/modules/log/config/install/views.view.commerce_activity.yml
+++ b/modules/log/config/install/views.view.commerce_activity.yml
@@ -1,3 +1,4 @@
+uuid: 146e4223-7414-428a-860d-0d5ffc5da7f8
 langcode: en
 status: true
 dependencies:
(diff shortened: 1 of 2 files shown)


modules/log/config/install/views.view.commerce_activity.yml
```diff
@@ -275,29 +275,14 @@ display:
...
-          plugin_id: date
         log_id:
           id: log_id
           table: commerce_log
           field: log_id
           relationship: none
           group_type: group
           admin_label: ''
-          order: DESC
+          order: ASC
```
bojanz, line 285: This needs to be DESC to keep the reverse-chronological ordering.

bojanz (changes requested):

bojanz:
One thing I noticed now: we currently sort by created DESC, log_id DESC.
So if two entries have the same created timestamp, the log_id sort takes over.
Why didn't that work on the client project?

I agree that the created DESC is unneeded, just commenting.

bojanz:
Committed final version in e2f9909d4bf14e96e0fbe0f8938c3a16865b9b81.

PR #680, opened 2017-03-17
Issue #2861664 by mglaman: Order assignment should generate a log to …

mglaman:
…mark the customer conversion

### DIFF
--- a/modules/log/src/EventSubscriber/OrderEventSubscriber.php
+++ b/modules/log/src/EventSubscriber/OrderEventSubscriber.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\commerce_log\EventSubscriber;
 
+use Drupal\commerce_order\Event\OrderAssignEvent;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\state_machine\Event\WorkflowTransitionEvent;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
@@ -34,6 +35,7 @@ class OrderEventSubscriber implements EventSubscriberInterface {
       'commerce_order.validate.pre_transition' => ['onValidateTransition', -100],
       'commerce_order.fulfill.pre_transition' => ['onFulfillTransition', -100],
       'commerce_order.cancel.pre_transition' => ['onCancelTransition', -100],
+      'commerce_order.order.assign' => ['onOrderAssign', -100],
     ];
     return $events;
   }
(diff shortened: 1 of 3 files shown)


bojanz:
The build failed because of phpcs:
```
FILE: ...e/commerce/modules/log/tests/src/Kernel/OrderIntegrationTest.php
FOUND 1 ERROR AND 1 WARNING AFFECTING 2 LINES
   8 | WARNING | [x] Unused use statement
     |         |     (Drupal.Classes.UnusedUseStatement.UnusedUse)
 200 | ERROR   | [x] Inline comments must end in full-stops, exclamation
     |         |     marks, colons, question marks, or closing
     |         |     parentheses
     |         |     (Drupal.Commenting.InlineComment.InvalidEndChar)
```

modules/log/src/EventSubscriber/OrderEventSubscriber.php
```diff
@@ -34,6 +36,7 @@ public static function getSubscribedEvents() {
       'commerce_order.validate.pre_transition' => ['onValidateTransition', -100],
       'commerce_order.fulfill.pre_transition' => ['onFulfillTransition', -100],
       'commerce_order.cancel.pre_transition' => ['onCancelTransition', -100],
+      OrderEvents::ORDER_ASSIGN => ['onOrderAssign', -100],
```
bojanz: We don't use constants above, so let's not use it here either, to stay consistent. 
Core generally discourages constants and recommends the real IDs to be used.

modules/log/tests/src/Kernel/OrderIntegrationTest.php
```diff
@@ -190,4 +193,25 @@ public function testPlaceValidateFulfillLogs() {
     $this->assertText('The order was fulfilled.');
   }
 
+  /**
+   * Tests that an order assignment log is generated.
+   */
+  public function testOrderAssignedLog() {
+    // Reassignment does not work on users
```
bojanz: The comment is confusing (doesn't work - why?). I suggest "Reassignment is currently only done on user login."

modules/log/src/EventSubscriber/OrderEventSubscriber.php
mglaman: We don't above because they don't exist. And that seems silly if core recommends real IDs yet constants provided. I'll change to keep it consistent here.

PR #682, opened 2017-03-20
Issue #2860840 : Product variantion add to cart not translated.

vasike:


src/Entity/CommerceContentEntityBase.php
```diff
@@ -0,0 +1,48 @@
...
+ * Provides the base class for Commerce content entities.
+ */
+class CommerceContentEntityBase extends ContentEntityBase implements CommerceContentEntityInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getTranslatedReferencedEntities($field_name) {
+    $refereced_entities = $this->get($field_name)->referencedEntities();
+    return $this->ensureTranslations($refereced_entities);
```
GoZOo: typo error on $refereced_entities. I guess you miss a 'n' in $referenced_entities

src/Entity/CommerceContentEntityBase.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Drupal\commerce\Entity;
+
+use Drupal\Core\Entity\ContentEntityBase;
+use Drupal\Core\Language\LanguageInterface;
+
+/**
+ * Provides the base class for Commerce content entities.
+ */
+class CommerceContentEntityBase extends ContentEntityBase implements CommerceContentEntityInterface {
```
mglaman: If this is a base class, it must be abstract

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -239,7 +239,7 @@ protected function getAttributeInfo(ProductVariationInterface $selected_variatio
       $field = $field_definitions[$field_name];
       $attributes[$field_name] = [
         'field_name' => $field_name,
-        'title' => $field->getLabel(),
+        'title' => $this->t($field->getLabel()),
```
mglaman: There's no need to wrap this in translatable markup, as the field is config and config is already translatable

mglaman (dismissed):
There's a lot of changes here. And we don't have any i18n testing. There's some initial adjustments to be made.

src/Entity/CommerceContentEntityBase.php
mglaman: Also, this only provides a public method for trnslated references. So really only relates to Product.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
vasike: i know, but there the field label won't get translated
i will update with some getTranslationFromContext somehow

src/Entity/CommerceContentEntityBase.php
vasike: i think the usage could be extented on other place - for example any getStore(s) needs
and thanks

src/Entity/CommerceContentEntityBase.php
vasike: indeed. merci

src/Entity/CommerceContentEntityBase.php
mglaman: How would stores need this? It doesn't have an architecture which has a concept of "child references"

mglaman:
Committed in 7af1dd81d0c4da9faebd86c941b4b0472bfa03ee

PR #684, opened 2017-03-21
2862339 - Add event listener to payments

haringsrob:
@bojanz I did actually provide a test for the following reason:
Even if we "assume" that the default behavior of the dispatcher would work, we will never have coverage for this functionality at all, leaving that file uncovered.

When I created the basic test for the Create event, everything went fine.

When I added the load event, it broke, I checked if ORDER_LOAD was ever called somewhere in the code and it is not, leaving me with 2 questions:

1. I think it is called dynamically, but I am not sure in what way
2. I really think it deserves full coverage for all the events, because now we can never say for sure that it actually works.

### DIFF
--- a/modules/payment/src/Event/PaymentEvent.php
+++ b/modules/payment/src/Event/PaymentEvent.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Drupal\commerce_payment\Event;
+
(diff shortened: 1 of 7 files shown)


haringsrob:
Closing in favour of #791

PR #686, opened 2017-03-22
issue #2859834 implementation of links in checkout progress block

finnef:


finnef:
DO: https://www.drupal.org/node/2859834

bojanz:
We also need a setting for enabling this, like the D7 module had.

modules/checkout/src/Plugin/Block/CheckoutProgressBlock.php
```diff
@@ -103,9 +104,20 @@ public function build() {
         $position = 'next';
       }
 
+      // Create breadcrumb style links for active checkout steps.
+      if ($index <= $current_step_index) {
+        $label = Link::createFromRoute($step_definition['label'], 'commerce_checkout.form', [
+          'commerce_order' => $order->id(),
+          'step' => $step_id,
+        ])->toString();
```
onatalushko: maybe ->to Renderable() will be better here ?

mglaman: Should always be renderable, unless that does fail

modules/checkout/templates/commerce-checkout-progress.html.twig
```diff
@@ -14,6 +14,6 @@
 #}
 <ol class="checkout-progress clearfix">
   {% for step in steps %}
-      <li class="checkout-progress--step checkout-progress--step__{{ step.position }}">{{ step.label }}</li>
+      <li class="checkout-progress--step checkout-progress--step__{{ step.position }}">{{ step.label | raw }}</li>
```
mglaman: If we remove `raw` toRenderable should be fine

mglaman (changes requested):

finnef:
changed to Renderable() and removed | raw.
Also added config checkbox

mglaman (changes requested):
```
    1)
    Drupal\Tests\commerce_checkout\Kernel\ChainCheckoutFlowResolverTest::testCheckoutFlowResolution
    Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for
    commerce_checkout.commerce_checkout_flow.default with the following errors:
    commerce_checkout.commerce_checkout_flow.default:configuration.display_checkout_progress_breadcrumb_links
    missing schema
```

the new option needs to be added to the module's schema.

can we also update a test where it clicks a link and goes back and continues?

finnef:
Added tests and schema.

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
```diff
@@ -155,13 +166,14 @@ public function testRegisterOrderCheckout() {
       'login[register][password][pass2]' => 'pass',
     ], 'Create account and continue');
     $this->assertSession()->pageTextContains('Billing information');
+    // Check breadcrumbs are not links. (the default setting)
+    $this->assertSession()->elementNotExists('css', '.block-commerce-checkout-progress li.checkout-progress--step > a');
```
mglaman, line 170: So we assert it's not a link, however we do not prove the feature works in any test

mglaman (changes requested):
Almost thumbs up here. However our tests do not prove this works or document the interaction.

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
finnef: The test to use the link to move back and fwd through the checkout steps is in testGuestOrderCheckout. The test for no link (by default) is in testRegisterOrderCheckout.

mglaman (approved):

modules/checkout/tests/src/Functional/CheckoutOrderTest.php
mglaman: Oh duh, wow i totally missed that. All good then!

bojanz:
Continued (rebased) in #732.

PR #689, opened 2017-03-22
Improve and expand Promotion plugins

mglaman:
…otion plugins


modules/promotion/src/Plugin/Commerce/PromotionOffer/FixedOff.php
```diff
@@ -0,0 +1,90 @@
...
+   */
+  public function execute() {
+    /** @var \Drupal\commerce_order\EntityAdjustableInterface $entity */
+    $entity = $this->getTargetEntity();
+
+    // @todo is there a sane way to add a getCurrencyCode to EntityAdjustableInterface.
+    // both order item and orders have ::getTotalPrice. Bug in the trenches.
+    $currency_code = $entity->getTotalPrice()->getCurrencyCode();
+    $adjustment_amount = new Price($this->getAmount(), $currency_code);
+    $this->applyAdjustment($entity, $adjustment_amount);
```
onatalushko, line 86: need check if entity total not more then adjustment total

mglaman: good catch

onatalushko: and I think we must do this check somethere in "base" class

mglaman: @bojanz this seems like a job inside of the entity, maybe? or should promotions check this

modules/promotion/src/Plugin/Commerce/PromotionOffer/FixedOff.php
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function execute() {
+    /** @var \Drupal\commerce_order\EntityAdjustableInterface $entity */
+    $entity = $this->getTargetEntity();
+
+    // @todo is there a sane way to add a getCurrencyCode to EntityAdjustableInterface.
+    // both order item and orders have ::getTotalPrice. Bug in the trenches.
+    $currency_code = $entity->getTotalPrice()->getCurrencyCode();
```
mglaman, line 84: :( @bojanz we have no way to get the currency code. This would break shipments.

src/Element/PluginSelect.php
```diff
@@ -127,6 +127,10 @@ public static function processPluginSelect(&$element, FormStateInterface $form_s
       $plugin = $plugin_manager->createInstance($target_plugin_id, $values['target_plugin_configuration']);
       if ($plugin instanceof PluginFormInterface) {
         $element['target_plugin_configuration'] = $plugin->buildConfigurationForm($element['target_plugin_configuration'], $form_state);
```
onatalushko, line 129: this should move after #parents set to make ajax works on target plugin form (my typos)

mglaman: okay, there's other issues about fixing the element and other items, which this might be part of

PR #690, opened 2017-03-23
Plugin select fixes

mglaman:


mglaman:
@bojanz can you review? I think this bumps on critical list for promotions. We'll need to test it against shipping, too.

modules/promotion/src/Plugin/Commerce/PromotionCondition/PromotionConditionBase.php
```diff
@@ -31,4 +33,12 @@ public function execute() {
     return $this->isNegated() ? !$result : $result;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
+    // @todo Update plugin_select to pass subform state automatically.
```
bojanz: Commerce plugins all rely on the complete form state. It would be weird to make conditions an exception to that, so the current workaround sounds better. We can evaluate converting all of our plugins to subform state in another issue.

bojanz:
Initializing the plugin for validation/submission with an empty $configuration sounds like a break from core behavior, and a potential source of bugs/confusion.

modules/promotion/src/Plugin/Commerce/PromotionCondition/PromotionConditionBase.php
mglaman: Okay +1

mglaman:
@bojanz the plugins are now initialized with their default values whenever a submit or AJAX rebuild triggers the form.

mglaman:
Changing a plugin target ID needs to reset the plugin config, like Address does for values.

mglaman:
Rebased into single commit so we can merge this 🔥

PR #692, opened 2017-03-24
Promotion form improvements

mglaman:
Opening for rolling patch

### DIFF
--- a/modules/promotion/css/promotion.form.css
+++ b/modules/promotion/css/promotion.form.css
@@ -1,0 +1,74 @@
+/**
+ * @file
+ * Styles for the product add/edit form.
+ *
+ * Copied from node-module.css
+ */
+
+/* Narrow screens */
+.layout-region {
+  box-sizing: border-box;
+}
+
+/* Wide screens */
+@media
+  screen and (min-width: 780px),
+  (orientation: landscape) and (min-device-height: 780px) {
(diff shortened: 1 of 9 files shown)


modules/promotion/css/promotion.form.css
```diff
@@ -0,0 +1,74 @@
+/**
+ * @file
+ * Styles for the product add/edit form.
```
bojanz, line 3: Untrue.

modules/promotion/src/Entity/Promotion.php
```diff
@@ -565,30 +565,23 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+      ->setLabel(t('Status'))
       ->setDescription(t('Whether the promotion is enabled.'))
       ->setDefaultValue(TRUE)
+      ->setRequired(TRUE)
+      ->setSettings([
+        'on_label' => t('Active'),
+        'off_label' => t('Disabled'),
+      ])
       ->setDisplayOptions('form', [
-        'type' => 'boolean_checkbox',
```
bojanz, line 572: Don't we want to revert back to the checkbox?

mglaman:
Committed in @4351967c03f3fae1bba0e2095b90701f606b59a8

PR #693, opened 2017-03-24
Plugin select widgets show weight field exposed when AJAX, and broken.

mglaman:
…posed when AJAX, and broken


mglaman:
🔥  @bojanz eyeballs O:)

PR #695, opened 2017-03-24
Issue #2763705 by mglaman: Implement a usage API

mglaman:
Here's a TDD approach to it. Uncovered a lot of flaws with our processing.

### DIFF
--- a/modules/promotion/commerce_promotion.install
+++ b/modules/promotion/commerce_promotion.install
@@ -1,0 +1,64 @@
+<?php
+
+/**
+ * @file
+ * Install, update and uninstall functions for the commerce_promotion module.
+ */
+
+/**
+ * Implements hook_schema().
+ */
+function commerce_promotion_schema() {
+  $schema['commerce_promotion_usage'] = [
+    'description' => 'Stores module data as key/value pairs per user.',
+    'fields' => [
+      'promotion_id' => [
+        'description' => 'Primary key: {commerce_promotion}.promotion_id for promotion.',
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'not null' => TRUE,
+        'default' => 0,
+      ],
+      'coupon_id' => [
+        'description' => 'Primary key: {commerce_promotion_coupon}.id for coupon.',
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'default' => 0,
+      ],
+      'order_id' => [
+        'description' => 'Primary key: {commerce_order}.order_id for order.',
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'not null' => TRUE,
+        'default' => 0,
+      ],
+      'uid' => [
+        'description' => 'Primary key: {users}.uid for user.',
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'not null' => FALSE,
+        'default' => 0,
+      ],
+    ],
+    'primary key' => ['promotion_id', 'coupon_id', 'order_id'],
+    'indexes' => [
+      'promotion_id' => ['promotion_id'],
+      'coupon_id' => ['coupon_id'],
+    ],
+    'foreign keys' => [
+      'promotion_id' => ['commerce_promotion' => 'promotion_id'],
+      'coupon_id' => ['commerce_promotion_coupon' => 'id'],
+      'order_id' => ['commerce_order' => 'order_id'],
+      'uid' => ['users' => 'uid'],
+    ],
+  ];
+
+  return $schema;
+}
+
+/**
+ * Install the `commerce_promotion_usage` table schema.
+ */
+function commerce_promotion_update_8001() {
+  drupal_install_schema('commerce_promotion');
+}
--- a/modules/promotion/commerce_promotion.post_update.php
+++ b/modules/promotion/commerce_promotion.post_update.php
@@ -84,3 +84,17 @@ function commerce_promotion_post_update_4() {
     $promotion->save();
   }
 }
+
+/**
+ * Remove promotion current_usage field.
+ */
+function commerce_promotion_post_update_5() {
+  $entity_definition_update = \Drupal::entityDefinitionUpdateManager();
+  $storage_definition = BaseFieldDefinition::create('integer')
+    ->setName('current_usage')
+    ->setTargetEntityTypeId('commerce_promotion')
+    ->setLabel(t('Current usage'))
+    ->setDescription(t('The number of times the promotion was used.'))
+    ->setDefaultValue(0);
+  $entity_definition_update->uninstallFieldStorageDefinition($storage_definition);
+}
--- a/modules/promotion/src/PromotionStorage.php
+++ b/modules/promotion/src/PromotionStorage.php
@@ -24,7 +24,26 @@ class PromotionStorage extends CommerceContentEntityStorage implements Promotion
     if (empty($result)) {
       return [];
     }
+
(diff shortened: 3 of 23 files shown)


modules/promotion/tests/src/Kernel/UsageTest.php
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * Manually test the add/get of usages.
+   *
+   * @covers ::getPromotionUsage
+   * @covers ::addPromotionUsage
+   * @covers ::getCouponUsage
+   * @covers ::addCouponUsage
+   */
+  public function testMockUsage() {
```
mglaman: @bojanz this method at leasts tests the logic and storage. Order refresh and legitimate implementation will take some time. If you'd like to review the table name, columns, and basic storage logic.

modules/promotion/src/PromotionUsageInterface.php
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Add a promotion usage record.
+   *
+   * @param \Drupal\commerce_promotion\Entity\PromotionInterface $promotion
+   *   The promotion.
+   * @param \Drupal\commerce_order\Entity\OrderInterface $order
+   *   The order.
+   */
+  public function addPromotionUsage(PromotionInterface $promotion, OrderInterface $order);
```
bojanz: why not a single addUsage($order, $promotion, $coupon = NULL)?

modules/promotion/src/PromotionUsageInterface.php
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Gets the usage for a coupon.
+   *
+   * @param \Drupal\commerce_promotion\Entity\CouponInterface $coupon
+   *   The coupon.
+   *
+   * @return int
+   *   The usage.
+   */
+  public function getCouponUsage(CouponInterface $coupon);
```
bojanz: Same here, getUsage($promotion, $coupon = NULL), and some documentation to clarify?

modules/promotion/src/PromotionStorage.php
```diff
@@ -24,8 +24,27 @@ public function loadValid(OrderTypeInterface $order_type, StoreInterface $store)
     if (empty($result)) {
       return [];
     }
+
+    $usages_query = $this->database->select('commerce_promotion_usage', 'cpu');
```
bojanz, line 28: Feels wrong to query the DB table outside of the service. We have an API, let's use it.

bojanz:
We need a post_update to drop the current_usage field.

modules/promotion/src/PromotionUsageInterface.php
mglaman: I guess. Just felt weird since we can get the promotion from the coupon, but a single method for adding works for me

modules/promotion/src/PromotionStorage.php
mglaman: The API doesn't allow for a bulk look up, and core does this as well within storage to provide a more performant query. Otherwise we'd have to do a SELECT for each promotion, and that's wasteful

bojanz: I agree that it's wasteful, but that means we need a new method on the service then. If we keep it this way people won't be able to easily swap the service to use a different storage, which is something that core encourages with its own examples.

bojanz: Which reminds me, we should look into the backend_overridable tag for the service.

modules/promotion/commerce_promotion.install
```diff
@@ -0,0 +1,64 @@
...
+ * @file
+ * Install, update and uninstall functions for the commerce_promotion module.
+ */
+
+/**
+ * Implements hook_schema().
+ */
+function commerce_promotion_schema() {
+  $schema['commerce_promotion_usage'] = [
+    'description' => 'Stores module data as key/value pairs per user.',
```
bojanz, line 13: Untrue.

modules/promotion/commerce_promotion.install
```diff
@@ -0,0 +1,64 @@
...
+    'fields' => [
+      'promotion_id' => [
+        'description' => 'Primary key: {commerce_promotion}.promotion_id for promotion.',
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'not null' => TRUE,
+        'default' => 0,
+      ],
+      'coupon_id' => [
+        'description' => 'Primary key: {commerce_promotion_coupon}.id for coupon.',
```
bojanz, line 23: None of these are primary keys, right? We might as well drop the phrase.

modules/promotion/commerce_promotion.post_update.php
```diff
@@ -84,3 +84,17 @@ function commerce_promotion_post_update_4() {
     $promotion->save();
   }
 }
+
+/**
+ * Remove promotion current_usage field.
```
bojanz, line 89: "Remove the promotion current_usage field."
People will see the lack of an article and think a Slavic person wrote this :P

modules/promotion/src/PromotionUsageInterface.php
```diff
@@ -0,0 +1,39 @@
...
+   * @param \Drupal\commerce_promotion\Entity\CouponInterface|null $coupon
+   *   The promotion's coupon, optional.
+   *
+   * @return int
+   *   The usage.
+   */
+  public function getUsage(PromotionInterface $promotion, CouponInterface $coupon = NULL);
+
+  /**
+   * Add a promotion usage record.
```
bojanz, line 28: "Adds a usage record for the given promotion." ?

modules/promotion/src/PromotionUsageInterface.php
```diff
@@ -0,0 +1,39 @@
...
+use Drupal\commerce_promotion\Entity\CouponInterface;
+use Drupal\commerce_promotion\Entity\PromotionInterface;
+
+/**
+ * Provides a way to check the usage of a promotion or coupon.
+ */
+interface PromotionUsageInterface {
+
+  /**
+   * Gets the usage for a promotion.
```
bojanz, line 15: "Gets the usage for the given promotion."?

modules/promotion/src/PromotionUsageInterface.php
```diff
@@ -0,0 +1,39 @@
...
+   *
+   * @param \Drupal\commerce_promotion\Entity\PromotionInterface $promotion
+   *   The promotion.
+   * @param \Drupal\commerce_promotion\Entity\CouponInterface|null $coupon
+   *   The promotion's coupon, optional.
```
bojanz, line 20: (Optional) The used coupon.

modules/promotion/src/PromotionStorage.php
```diff
@@ -24,8 +24,27 @@ public function loadValid(OrderTypeInterface $order_type, StoreInterface $store)
...
+    $usages_query->condition('promotion_id', array_keys($result), 'IN');
+    $usages_query->addField('cpu', 'promotion_id');
+    $usages_query->addExpression("COUNT(promotion_id)", 'count');
```
bojanz, line 31: Let's call the PromotionUsage method: getUsageMultiple(array $promotions, array $coupons)

bojanz (changes requested):

bojanz:
Final version committed in 5ab9e1477719aa943b79dfb35b909bbd523337b1.
I'll explain what changed in the issue.

PR #697, opened 2017-03-25
Log deps

mglaman:
…lling without commerce module enabled


bojanz:
Committed in f488a12a85159d0e177b1b050aa3747315e80fbf.

PR #698, opened 2017-03-27
Add a new promotion condition to filter by product

mglaman:


bojanz:
Now that 6736c38237a83339a7e09054e96c910bd1319afb this can now be rebased. 
Personally I'd repurpose it to only focus on the limit by product value functionality.

PR #701, opened 2017-03-31
Issue #2865705 by bojanz, mglaman: Promotions should operate on orders only

mglaman:
…s only


bojanz:
Committed in 6ea3160ece594f37b66caa4abcbbb97ca0610b55 with tweaks.

PR #702, opened 2017-03-31
Issue #2792653 by niko-, padma28, mglaman, chishah92, a.dmitriiev, an…

mglaman:
…dypost, steveoliver: Checkout coupon pane implementation

### DIFF
--- a/modules/promotion/config/schema/commerce_promotion.schema.yml
+++ b/modules/promotion/config/schema/commerce_promotion.schema.yml
@@ -1,0 +1,6 @@
+commerce_checkout.commerce_checkout_pane.coupon_redemption:
+  type: commerce_checkout_pane_configuration
+  mapping:
+    multiple_coupons:
+      type: boolean
+      label: 'Allow multiple coupons'
--- a/modules/promotion/src/Element/CouponRedemptionForm.php
+++ b/modules/promotion/src/Element/CouponRedemptionForm.php
@@ -5,8 +5,13 @@ namespace Drupal\commerce_promotion\Element;
 use Drupal\commerce\Element\CommerceElementTrait;
 use Drupal\commerce_order\Entity\OrderInterface;
 use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Ajax\AjaxResponse;
+use Drupal\Core\Ajax\CommandInterface;
+use Drupal\Core\Ajax\InsertCommand;
+use Drupal\Core\Ajax\PrependCommand;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element\FormElement;
+use Drupal\commerce_order\Adjustment;
 
 /**
  * Provides a form element for redeeming a coupon.
@@ -37,20 +42,17 @@ class CouponRedemptionForm extends FormElement {
   public function getInfo() {
     $class = get_class($this);
     return [
-      '#title' => t('Coupon code'),
-      '#description' => t('Enter your coupon code here.'),
-      '#submit_title' => t('Apply coupon'),
-      '#submit_message' => t('Coupon applied'),
-      '#remove_title' => t('Remove coupon'),
       // The coupon ID.
       '#default_value' => NULL,
       '#order_id' => NULL,
+      '#multiple_coupons' => FALSE,
       '#process' => [
         [$class, 'processForm'],
       ],
       '#element_validate' => [
         [$class, 'validateForm'],
       ],
+      '#element_ajax' => [],
       '#theme_wrappers' => ['container'],
     ];
   }
@@ -81,7 +83,6 @@ class CouponRedemptionForm extends FormElement {
       throw new \InvalidArgumentException('The commerce_coupon_redemption #order_id must be a valid order ID.');
     }
 
-    $has_coupons = !$order->get('coupons')->isEmpty();
     $id_prefix = implode('-', $element['#parents']);
     // @todo We cannot use unique IDs, or multiple elements on a page currently.
     // @see https://www.drupal.org/node/2675688
@@ -95,15 +96,18 @@ class CouponRedemptionForm extends FormElement {
       // Pass the id along to other methods.
       '#wrapper_id' => $wrapper_id,
     ] + $element;
+
+    $has_coupons = !$order->get('coupons')->isEmpty();
+    $element['coupons'] = CouponRedemptionForm::buildCouponsTable($element, $order);
     $element['code'] = [
       '#type' => 'textfield',
-      '#title' => $element['#title'],
-      '#description' => $element['#description'],
-      '#access' => !$has_coupons,
+      '#title' => t('Coupon code'),
+      '#description' => t('Enter your coupon code to redeem a promotion.'),
+      '#access' => !$has_coupons || $element['#multiple_coupons'],
     ];
     $element['apply'] = [
       '#type' => 'submit',
-      '#value' => $element['#submit_title'],
+      '#value' => t('Redeem'),
       '#name' => 'apply_coupon',
       '#limit_validation_errors' => [
         $element['#parents'],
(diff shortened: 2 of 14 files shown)


modules/promotion/config/schema/commerce_promotion.schema.yml
```diff
@@ -0,0 +1,21 @@
...
+    submit_title:
+      type: string
+      label: 'Submit button title'
+    submit_message:
+      type: string
+      label: 'Submit message'
+    remove_title:
+      type: string
+      label: 'Remove button title'
+    single_coupon_mode:
```
mglaman: I don't like this wording. "use_multiple"? It's the same wording used in the element type.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -45,6 +50,8 @@ public function getInfo() {
       // The coupon ID.
       '#default_value' => NULL,
       '#order_id' => NULL,
+      '#display_actions' => TRUE,
+      '#single_coupon_mode' => TRUE,
```
mglaman: See first comment.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -45,6 +50,8 @@ public function getInfo() {
       // The coupon ID.
       '#default_value' => NULL,
       '#order_id' => NULL,
+      '#display_actions' => TRUE,
```
mglaman: I don't know why this was put there. Need to review issue comments. Not sure why you would embed the element but not display actions? To list coupons?

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -81,7 +88,26 @@ public static function processForm(array $element, FormStateInterface $form_stat
...
+
+    // Determine if 'add coupon' buttons should be displayed. It should be
+    // on display_actions mode, if there is no coupons, or if at least one of
+    // the coupons is compatible with others.
+    $display_actions = !$has_coupons;
+
+    if (!$element['#single_coupon_mode']) {
+      /** @var \Drupal\commerce_promotion\Entity\CouponInterface $coupon */
+      foreach ((array) $coupons as $coupon) {
+        if ($coupon->getPromotion()->getCompatibility() == PromotionInterface::COMPATIBLE_ANY) {
```
mglaman: We should remove this. Not this element's job to respect compatibility. That's on redemption validation.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -95,11 +121,13 @@ public static function processForm(array $element, FormStateInterface $form_stat
       // Pass the id along to other methods.
       '#wrapper_id' => $wrapper_id,
     ] + $element;
+    $element['coupons'] = CouponRedemptionForm::buildAdjustmentsTable($element, $order, $display_actions);
```
mglaman: `self::`

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -95,11 +121,13 @@ public static function processForm(array $element, FormStateInterface $form_stat
...
+    $element['coupons']['#access'] = $has_coupons && !$element['#single_coupon_mode'];
```
mglaman: Should just wrap in if/else.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -144,7 +171,20 @@ public static function processForm(array $element, FormStateInterface $form_stat
...
     array_pop($parents);
-    return NestedArray::getValue($form, $parents);
+    $coupon_element = NestedArray::getValue($form, $parents);
+
+    $response = new AjaxResponse();
+    // To refresh the coupon.
+    $response->addCommand(new InsertCommand(NULL, $coupon_element));
+    // To refresh the order summary.
+    // @todo the checkout pane should handle this, not the element.
+    if (isset($form['sidebar']['order_summary'])) {
```
mglaman: I want to find an elegant way to attach extra callbacks here. Maybe `#element_ajax`?

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -224,12 +289,98 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+    $table = [
+      '#type' => 'table',
+      '#header' => [t('Label'), t('Amount')],
+      '#empty' => t('There are no special offers applied.'),
+    ];
+    if ($display_actions) {
+      $table['#header'][] = t('Remove');
+    }
+
+    $adjustments = $order->getAdjustments();
```
mglaman: This and following needs to be `collectAdjustments`

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -224,12 +289,98 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+        '#template' => '{{ label }}',
+        '#context' => [
+          'label' => $label,
+        ],
+      ];
+      $table[$index]['amount'] = [
+        '#type' => 'inline_template',
+        '#template' => '{{ price|commerce_price_format }}',
+        '#context' => [
+          'price' => $adjustment->getAmount(),
```
mglaman, line 321: The above logic seems insanely complex to just get the adjustment amount. Should review. Or open for a follow up.

modules/promotion/src/Plugin/Commerce/CheckoutPane/CouponRedemption.php
```diff
@@ -0,0 +1,105 @@
...
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return [
+      'title' => $this->t('Coupon code'),
+      'description' => $this->t('Enter your coupon code to redeem a promotion.'),
+      'submit_title' => $this->t('Redeem'),
+      'submit_message' => $this->t('Coupon applied'),
+      'remove_title' => $this->t('Remove coupon'),
+      'single_coupon_mode' => TRUE,
```
mglaman: See first comment.

mglaman (commented):
Initial nitpicks I found after finally reviewing the code.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -224,12 +289,98 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+   *   The element.
+   * @param \Drupal\commerce_order\Entity\OrderInterface $order
+   *   The order.
+   * @param bool $display_actions
+   *   TRUE if actions displayed.
+   *
+   * @return array
+   *   The element array.
+   */
+  public static function buildAdjustmentsTable(array $element, OrderInterface $order, $display_actions = TRUE) {
```
ReplicaObscura: @mglaman It might be useful if there were a separate render element for listing coupons, so that it could easily be rendered outside of the coupon redemption form as well.

As an example use case, I'm trying to get almost the exact same table (but without action buttons, giving one use case for that parameter) into the order receipt template so that it can be included in order emails.

modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -224,12 +271,95 @@ public static function validateForm(array &$element, FormStateInterface $form_st
       $form_state->setErrorByName($code_path, t('Coupon is invalid'));
       return;
     }
-    if (!$coupon->getPromotion()->applies($order)) {
-      $form_state->setErrorByName($code_path, t('Coupon is invalid'));
-      return;
-    }
```
ReplicaObscura: Howcome this was removed? At least in multiple coupon mode, this makes it appear to apply a coupon (show a success message) even when the promotion's conditions don't match. Haven't tested single coupon mode since this change yet.

mglaman: Because I thought Coupon available triggered that. but I guess not.

Coupon::available checks: status, promotion available
Promotion::available checks: bundle, stores, start/end time, usage limit.

So I guess we still need that.

mglaman:
shopping_cart___site-install

shopping_cart___site-install

ReplicaObscura:
Currently this is working pretty well for me. Only issue I'm running into now is that redeeming a coupon only seems to "stick" if it's done on the Review page. If it's done on the Order Information page, the coupon gets lost when getting to the Review step. I tried #703 to fix this issue, and it technically worked, but broke other things in even bigger ways. Strangely, this problem didn't seem to exist with the patch from a few days ago, and I'm not sure exactly when it started happening.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentInformation.php
```diff
@@ -317,6 +317,7 @@ public function validatePaneForm(array &$pane_form, FormStateInterface $form_sta
    * {@inheritdoc}
    */
   public function submitPaneForm(array &$pane_form, FormStateInterface $form_state, array &$complete_form) {
+    $this->order = $this->checkoutFlow->getOrder();
```
mglaman: @bojanz this fixes it. Along with rebuild refresh of order. So.. uh. I'm not sure what to do. Do we have the base form always refresh the order object? I can't think of how we can work around this.

bojanz:
Committed in ed36542d273c35c53122c49f2dd9ee0597a72434.

PR #703, opened 2017-04-02
CheckoutFlowBase should reload order on AJAX rebuild

mglaman:
Initial commit should cause test failures based on issue summary here https://www.drupal.org/node/2861529

### DIFF
--- a/modules/payment/tests/src/FunctionalJavascript/PaymentCheckoutTest.php
+++ b/modules/payment/tests/src/FunctionalJavascript/PaymentCheckoutTest.php
@@ -57,7 +57,7 @@ class PaymentCheckoutTest extends CommerceBrowserTestBase {
    */
   protected function getAdministratorPermissions() {
     return array_merge([
-      'administer profiles',
+      'administer profile',
     ], parent::getAdministratorPermissions());
   }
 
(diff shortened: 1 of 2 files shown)


modules/payment/tests/src/FunctionalJavascript/PaymentCheckoutTest.php
```diff
@@ -57,7 +57,7 @@ class PaymentCheckoutTest extends CommerceBrowserTestBase {
    */
   protected function getAdministratorPermissions() {
     return array_merge([
-      'administer profiles',
+      'administer profile',
```
ReplicaObscura, line 60: I think that this change might already be included in the latest commerce dev, on a different line, causing it not to apply cleanly. The change to CheckoutFlowBase applies just fine though.

ReplicaObscura:
In testing, it seems this is causing a problem at least in my specific case: I'm sometimes ending up with no billing or shipping information being added to the order.

I've got billing and shipping info on the same pane, and this is what happened:

1. Enter shipping information and then select a shipping method
2. Enter billing information
3. Enter and redeem a coupon code
4. Try to continue to next step of checkout, page reloads with "Select a shipping method" error
5. Try to continue again after re-selecting a shipping method
6. Review page comes up with no billing or shipping information visible

mglaman:
Proper fix is in #702

PR #704, opened 2017-04-03
Issue #2853280 : Cart should not depend on Product.

vasike:


mglaman:
Closing, let's just continue this in the issue queue via patch https://www.drupal.org/project/commerce/issues/2853280#comment-12504216

PR #705, opened 2017-04-03
Issue #2763255 by mglaman: Implement a nicer discount listing

mglaman:


mglaman:
promotions___site-install

mglaman:
promotions___site-install

mglaman:
promotions___site-install

mglaman:
promotions___site-install

bojanz:
Initial thoughts:

1) While the infinity sign is cute, it probably makes more sense to show "0 / Unlimited" for unlimited promotions. That way the useful info of "how often was this used" is still there even if there's no limit.

2) The smushing of start date / end date into status is confusing. "Inactive" is a synonym of "Disabled", and it's unclear that you need to activate a promotion manually even if you've set a start date in the future. I suggest having separate Start date and End date columns that can be sorted on, after the usage. The actual status can be appended to the label, and it's probably safe to do it for disabled promotions only. For example "10% off T-shirts (Disabled)"

3) I would expect the promotion label to be detailed enough to make the offer not needed.
Plus, if we'd include the offer, we'd want to allow the plugin to output a summary based on configuration ("10% off the order total", not "Percentage off the order total"). Shopify does something like that, though I still feel that it duplicates the label.

BigCommerce UI: 
https://support.bigcommerce.com/servlet/rtaImage?eid=ka613000000CiP6&feoid=00N1300000BR3CT&refid=0EM130000000ywY
Shopify UI: https://d13yacurqjgara.cloudfront.net/users/11118/screenshots/1512612/shopify-admin-discounts.png

mglaman:
My initial​ PR was more like BigCommerce and also mimicked most ad
dashboards. Showed if actually enabled or not. Then if running (based on
start,end, and usage.) This was based off of design work from Kelowna in
July.

Yeah. Bringing infinity symbol from D7 was a poor choice

bojanz:
Wrapped up in a02d6d7f5cc1cdaf2eb665ad33b1261c01ce96f0.

PR #707, opened 2017-04-04
Variation title generate event

ReplicaObscura:
Sorry for the stray merge commits, there are no changes from those merges represented here.

This is an attempt to resolve https://www.drupal.org/node/2846895 by adding a new event that's dispatched after generating the variation title, before saving it to the variation.


mglaman (changes requested):
Can we add a test module, like `commerce_product_test` even, which subscribes to this event. If the product's SKU is `GENERATE_TEST` (or something) it'll react and change the title. So we can prove it works and document the feature via tests.

ReplicaObscura:
I'll work on these tests ASAP btw, sorry for leaving this out there for so long.

PR #708, opened 2017-04-04
#2863864 - Disable promotions via cron if end date or usage limit hit

swickham78:


modules/promotion/commerce_promotion.module
```diff
@@ -79,3 +79,30 @@ function commerce_promotion_entity_base_field_info(EntityTypeInterface $entity_t
...
+ * Implements hook_cron()
+ */
+function commerce_promotion_cron() {
+  commerce_promotion_disabled_expired();
+}
+
+/**
+ * Get all expired promotions that are still active and disable them.
+ */
+function commerce_promotion_disabled_expired() {
```
mglaman: If the storage provides a method to load expired, we can just keep this within the cron, no need for its own function. If people want to run this manually then they can.

modules/promotion/commerce_promotion.module
```diff
@@ -79,3 +79,30 @@ function commerce_promotion_entity_base_field_info(EntityTypeInterface $entity_t
...
+  /** @var \Drupal\commerce_promotion\PromotionStorageInterface $promotion_storage */
+  $promotion_storage = \Drupal::service('entity_type.manager')->getStorage('commerce_promotion');
+
+  // Get query for all expired promotions.
+  $promotions = $promotion_storage->loadExpired();
+
+  if ($promotions !== FALSE) {
+    // Disable all expired promotions.
+    foreach ($promotions as $promotion) {
```
mglaman, line 98: @bojanz I think this is fine versus a queue. There shouldn't be that much activity, generally.

modules/promotion/src/PromotionStorage.php
```diff
@@ -107,6 +107,46 @@ public function loadByCoupon(OrderTypeInterface $order_type, StoreInterface $sto
...
+   *   Only include currently enabled promotions that have expired.
+   * @return array|bool|\Drupal\Core\Entity\EntityInterface[]
+   *   The expired promotion entities. Returns FALSE if none found.
+   */
+  public function loadExpired($only_enabled = TRUE) {
+    // Subquery to determine the amount of times a coupon has been used.
+    $usage_query = $this->database->select('commerce_promotion_usage', 'cpu');
+    $usage_query->addExpression('COUNT(cpu.promotion_id)', 'count');
+    $usage_query->where('cpu.promotion_id = cpd.promotion_id');
+    $usage_query->groupBy('cpu.promotion_id');
```
mglaman: We already have this query in the usage service `:: getUsageMultiple`. It feels like we should rework this method a bit so it can be used.

https://github.com/drupalcommerce/commerce/blob/8.x-2.x/modules/promotion/src/PromotionUsage.php#L65

modules/promotion/src/PromotionStorage.php
```diff
@@ -107,6 +107,46 @@ public function loadByCoupon(OrderTypeInterface $order_type, StoreInterface $sto
...
+
+    $query = $this->database->select($this->getDataTable(), 'cpd');
+
+    // We want to get results of queries that have passed their final date or
+    // have met their max usage.
+    $or_condition = $query->orConditionGroup()
```
mglaman: This is using a regular select, when we can run `$this->getQuery()`.

I'd actually rather do a load of `->condition('end_date', gmdate('Y-m-d'), '<')` and `->condition('status', TRUE)`

We need `<` because it's valid if `>=`. This would cancel promotions on their last day.

modules/promotion/src/PromotionStorage.php
```diff
@@ -107,6 +107,46 @@ public function loadByCoupon(OrderTypeInterface $order_type, StoreInterface $sto
...
+      ->condition('cpd.usage_limit', $usage_query, '<=');
+
+    $query->addField('cpd', 'promotion_id');
+    $query->condition($or_condition);
+
+    if ($only_enabled) {
+      $query->condition('cpd.status', 1);
+    }
+
+    $result = $query->execute()->fetchCol(0);
```
mglaman: We return the invalid and set them as disabled. Then I'd say run the usage checks afterwards for anything not yet expired

mglaman (changes requested):
There are a few changes and no tests. We definitely need to test this.

I don't think we need to use a queue worker. But I'd like to use more of the API for querying the promotions. Even if this means running an entity query to load invalid via end date. Then load valid and check their usages and add to list of promotions to exired/disable.

I'm not too worried on the performance since this should only be invoked during cron.

modules/promotion/src/PromotionStorage.php
swickham78: Would it work if I add another method to PromotionStorage that will load and check promotion uses by calling getUsageMultiple? That way I can run a quick query to only pass promotions to it which are active and have a usage set.

I think the only way I can mess around with getUsageMultiple to handle return usage on its own would be to be just make it load all promotions if none are passed which could be excessive even if we're not too concerned w/ performance.

Thoughts?

mglaman:
@swickham78 at quick glance this looks good, better change. Bojan made some changes which are causing conflict.

swickham78:
@mglaman Conflicts have been addressed.

mglaman:
@swickham78 okay, cool. Errors are just PHPCS. I'll get around to a proper review

modules/promotion/src/PromotionStorage.php
```diff
@@ -104,4 +104,61 @@ public function loadAvailable(OrderTypeInterface $order_type, StoreInterface $st
     return $promotions;
   }
 
+  /**
+   * Return active promotions that have passed their end date.
```
mglaman: Document blocks should only be on the interface, implementers just do `{@inheritdoc}`

modules/promotion/src/PromotionStorage.php
```diff
@@ -104,4 +104,61 @@ public function loadAvailable(OrderTypeInterface $order_type, StoreInterface $st
...
+
+    if (empty($result)) {
+      return [];
+    }
+
+    return $this->loadMultiple($result);
+  }
+
+  /**
+   * Returns active promotions which have a met their maximum usage.
```
mglaman: Document blocks should only be on the interface, implementers just do `{@inheritdoc}`

modules/promotion/src/PromotionStorageInterface.php
```diff
@@ -24,4 +24,20 @@
...
+   */
+  public function loadExpired();
+
+  /**
+   * Returns active promotions which have a met their maximum usage.
+   *
+   * @return \Drupal\commerce_promotion\Entity\PromotionInterface[]
+   *   Promotions with maxed usage.
+   */
+  public function loadMaxedUsage();
```
mglaman, line 41: @bojanz name bikeshedding.

modules/promotion/tests/src/Kernel/PromotionStorageTest.php
```diff
@@ -215,4 +250,64 @@ public function testWeight() {
...
+    $expired_promotion = Promotion::create([
+      'name' => 'Expired Promotion',
+      'status' => TRUE,
+      'start_date' => gmdate('Y-m-d', time() - 172800),
+      'end_date' => gmdate('Y-m-d', time() - 86400),
+    ]);
+    $this->assertEquals(SAVED_NEW, $expired_promotion->save());
+
+    $promotions = $this->promotionStorage->loadExpired();
+    $this->assertEquals(count($promotions), 1);
```
mglaman: Can do assertCount

mglaman (changes requested):
The storage methods are tested, but the cron op is still not tested.

swickham78:
@mglaman Completely forgot this was left unresolved. Just pushed a new commit with adjustments based on your last notes.

swickham78:
@mglaman Looks like this got lost in the shuffle. Nothing changed since the last push, just did a rebase and fixed some conflicts that came form it.

mglaman (approved):

PR #709, opened 2017-04-04
Issue 2842961 by drugan, smartparty, b_sharpe: Administer attribute labels

drugan:
Read more: https://www.drupal.org/node/2842961


bojanz:
Continued on the issue (https://www.drupal.org/node/2842961)

Adding to the list of issues to wrap up.

PR #710, opened 2017-04-05
Issue #2867046 by vasike, bojanz: Don't run order processors if the order has no items

vasike:


modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
```diff
@@ -18,9 +18,11 @@ class OrderPercentageOff extends PercentageOffBase {
    */
   public function execute() {
     $order = $this->getOrder();
-    $adjustment_amount = $order->getTotalPrice()->multiply($this->getAmount());
-    $adjustment_amount = $this->rounder->round($adjustment_amount);
-    $this->applyAdjustment($order, $adjustment_amount);
+    if ($order->getTotalPrice()) {
```
mglaman: This sounds like a bug where the order should _always_ return a total price. I say we remove this line, prove test fails and fix order entity for a proper TDD approach.

mglaman (changes requested):
This doesn't feel like a proper fix and allows for bugs to still happen outside of this scope.

modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderPercentageOff.php
vasike: updated with other approach, you suggested

mglaman (approved):

mglaman:
@vasike wonderful :) Thank you for update. It's a much more hardened approach that will prevent others in contrib / custom implementation from hitting error.

bojanz:
Committed an improved version in 005e9046af683bed84087d8c9f7fa0d196432657.

PR #713, opened 2017-04-13
Issue #2869333 by agoradesign: Move store context from PriceDefaultFo…

agoradesign:
…rmatter to PriceCalculatedFormatter

### DIFF
--- a/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
+++ b/modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
@@ -51,6 +51,13 @@ class PriceCalculatedFormatter extends PriceDefaultFormatter implements Containe
   protected $currentUser;
 
   /**
+   * The store context.
+   *
+   * @var \Drupal\commerce_store\StoreContextInterface
+   */
+  protected $storeContext;
+
+  /**
    * Constructs a new PriceCalculatedFormatter object.
    *
    * @param string $plugin_id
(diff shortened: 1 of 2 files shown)


modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
```diff
@@ -79,8 +86,10 @@ class PriceCalculatedFormatter extends PriceDefaultFormatter implements Containe
    *   The current user.
    */
   public function __construct($plugin_id, $plugin_definition, FieldDefinitionInterface $field_definition, array $settings, $label, $view_mode, array $third_party_settings, EntityTypeManagerInterface $entity_type_manager, NumberFormatterFactoryInterface $number_formatter_factory, ChainPriceResolverInterface $chain_price_resolver, StoreContextInterface $store_context, AccountInterface $current_user) {
-    parent::__construct($plugin_id, $plugin_definition, $field_definition, $settings, $label, $view_mode, $third_party_settings, $entity_type_manager, $number_formatter_factory, $store_context);
+    parent::__construct($plugin_id, $plugin_definition, $field_definition, $settings, $label, $view_mode, $third_party_settings, $entity_type_manager, $number_formatter_factory);
+
     $this->chainPriceResolver = $chain_price_resolver;
+    $this->storeContext = $store_context;
```
mglaman, line 92: This isn't being injected anymore. The constructor lists it but you did not override the parent `create` method.

mglaman (dismissed):

modules/price/src/Plugin/Field/FieldFormatter/PriceCalculatedFormatter.php
agoradesign: The parent create() method is already overridden anyway (and of course already containing the store context)

PR #714, opened 2017-04-16
Commerce profile select reuse

ReplicaObscura:
Here is my first crack at creating generic checkbox reuse functionality for the commerce_profile_select element. In this iteration, I could not figure out how to accomplish the logic in a generic way without the use of a callback function, though I'm sure it would be possible to make this more declarative with some additional thought.

The 3 options added are:
```
 *   '#reuse_profile_label' => $this->t('My billing address is the same as my shipping address.'),
 *   '#reuse_profile_source' => 'commerce_shipping_get_shipping_profile',
 *   '#reuse_profile_default' => FALSE,
```

I don't know if these are the best names for the options, I wanted to keep them as generic as possible. reuse_profile_source can be either a profile ID, or a callback function (since the profiles during checkout likely won't have IDs yet). reuse_profile_default simply determines if you want the checkbox initially checked or not. The first two options must be specified in order for the checkbox to show up at all, and the third is optional.

I'm testing this by modifying the PaymentAddForm billing_information element with the same options as specified above. And in commerce_shipping, to support it in a generic way that also works with the other profile select patch, here is the callback function I'm using:

```
function commerce_shipping_get_shipping_profile(array $element, FormStateInterface $form_state, array $complete_form) {
  $profile = NULL;

  $storage = $form_state->getStorage();
  if (isset($storage['pane_shipping_information[shipping_profile]']['profile'])) {
    $profile = $storage['pane_shipping_information[shipping_profile]']['profile'];
  } elseif (isset($complete_form['shipping_information']['shipping_profile']['#profile'])) {
    $profile = $complete_form['shipping_information']['shipping_profile']['#profile'];
  }

  return $profile;
}
```
I did not modify the PaymentAddForm in this patch, I figured it might make sense for there to be a separate patch that both adds these options to all the various billing profile forms along with pane options in the UI for controlling it. Additionally, a similar patch for commerce_shipping is needed.

With the callback function approach, the patches will be somewhat inter-dependent if we want to keep the logic encapsulated within the module that defines it. I figure commerce_shipping can offer a callback to return the default shipping profile from the form, and likewise commerce_order or commerce_checkout might be appropriate for a callback that can return the billing profile from the form, and then within the shipping pane it can reference the order callback, and vice versa. Does that make sense, or should the logic be reversed so shipping actually knows how to look up the billing profile from the form state itself?

ps: Sorry about the two pull requests in the commit log, mistakenly pulled upstream updates in the main branch that way and can't seem to get rid of them now without screwing up my other open pull requests.


bojanz:
Closing this PR, since it is outdated. "Billing same as shipping" is being implemented on the drupal.org issue.

PR #715, opened 2017-04-16
Commerce profile select reuse combined

ReplicaObscura:
This is related to the other PR I just submitted, #714, as well as to #631.

This is a combined PR containing both the profile reuse functionality from #714 as well as the profile select functionality from #631, since this requires a number of changes for them to both be able to apply together.

Obviously, one could be merged first, and then the other patch modified to support it, but it seemed simpler to tackle them together because they share some of the same functionality.


ReplicaObscura:
Latest updates work with the latest version of commerce. Also revamps much of the ProfileSelect code from the previous patches into something that I think is a little bit more manageable. Feedback needed.

bojanz:
Closing this. Matt is making big changes to the profile reuse PR, which will require the profile copy PR to be rebased. We couldn't do both at the same time because just the reuse PR has a huge scope.

PR #716, opened 2017-04-16
#2852207: Reuse shipping address

ReplicaObscura:
As a followup to #714 and #715 (depending which one is used), here's a PR that adds options to the UI for controlling the checkbox.

These will probably need to be modified/expanded to match what was in 1.x, I just needed to whip something up pretty quickly for now.

Since I was making the exact same changes to both BillingInformation and PaymentInformation, I decided to make a base class for them both to extend. Hopefully that's alright, but I can just copy the code between them if that would be better.

Note that this doesn't require #714 to apply and have the options function, but it does in order for the checkbox functionality to work, since that's all provided by the other patch.


ransomweaver:
This PR no longer applies as a patch. Can it be rebased? I am using it with 715.

bojanz:
Closing, since this work is outdated and isn't mergeable.

Thank you for your efforts, bmcclure

PR #717, opened 2017-04-17
Issue #2867789 by mglaman: Cart form and block strip trailing zeros i…

mglaman:
…n currency format

@bojanz I didn't add an update hook because of #712, which does. We maybe we merge these two together.. or commit this then that right away. Two fixes, one update hook.


bojanz:
I'll merge #712 next then.

PR #718, opened 2017-04-17
Issue #2831945 by mglaman: The payment step is always present in the …

mglaman:
…checkout progress block


bojanz:
Committed the real fix in 74ad0a4f13172ac30783ec578eeafa972d5019db.

PR #719, opened 2017-04-17
Issue #2870040 by mglaman: Remove payment method from order if there …

mglaman:
…is a decline exception


mglaman:
The test failures have to do with orders which have have  payment gateway reference but no payment method.

modules/payment/src/Plugin/Commerce/CheckoutPane/PaymentProcess.php
```diff
@@ -118,6 +118,8 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
       catch (DeclineException $e) {
         $message = $this->t('We encountered an error processing your payment method. Please verify your details and try again.');
         drupal_set_message($message, 'error');
+        $this->order->get('payment_gateway')->setValue(NULL);
```
mglaman, line 180: @bojanz If we do not remove this.. then the chain of logic here fails, which is another bug. Not sure if I should fix here or not.

```
    $default_option = NULL;
    if ($order_payment_method) {
      $default_option = $order_payment_method->id();
    }
    elseif ($order_payment_gateway && !($order_payment_gateway instanceof SupportsStoredPaymentMethodsInterface)) {
      $default_option = $order_payment_gateway->id();
    }
```

Returns payment gateway... which is not a method.

```
    $selected_option = $pane_form['payment_method'][$default_option];
    $payment_gateway = $payment_gateways[$selected_option['#payment_gateway']];
    if ($payment_gateway->getPlugin() instanceof SupportsStoredPaymentMethodsInterface) {
```

This supports no reconciliation for the fact we have a gateway ID. So I suppose the thing to do is remove returning gateway ID?

PR #721, opened 2017-04-18
The ProfileSelect form element breaks value callbacks of contained form elements

Berdir:
PR for https://www.drupal.org/node/2857051

Seems to fix my reported problem with date fields, did not really test anything else yet.


bojanz:
This looks good to me.

mglaman:
Looks good to me as well, and makes sense. Can't say I 100% understand it, but there is a linked failing test in core. I think we're fine to make a breaking chance like this.

PR #722, opened 2017-04-18
Split validation into multiple methods to simplify login and registration form customizations

Berdir:
PR for https://www.drupal.org/node/2870783

In my project, I have this now, combined with an alter hook to switch out the class:

```
/**
 * Customized Login pane that allows registration by e-mail.
 */
class EmailLoginPane extends Login {

  /**
   * {@inheritdoc}
   */
  public function buildPaneForm(array $pane_form, FormStateInterface $form_state, array &$complete_form) {
    $pane_form = parent::buildPaneForm($pane_form, $form_state, $complete_form);
    $pane_form['returning_customer']['name']['#title'] = $this->t('E-mail or username');
    $pane_form['register']['name']['#access'] = FALSE;
    return $pane_form;
  }

  /**
   * {@inheritdoc}
   */
  protected function getAndValidateRegisterUserValues(array &$pane_form, FormStateInterface $form_state) {
    if ($user_values = parent::getAndValidateRegisterUserValues($pane_form, $form_state)) {
      $user_values['name'] = $user_values['mail'];
      return $user_values;
    }
  }

}
```

Not sure yet about spliting up the login method further.

### DIFF
--- a/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
+++ b/modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
@@ -11,6 +11,7 @@ use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Url;
 use Drupal\Core\Link;
 use Drupal\user\UserAuthInterface;
+use Drupal\user\UserInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
 
@@ -195,7 +196,7 @@ class Login extends CheckoutPaneBase implements CheckoutPaneInterface, Container
       '#type' => 'textfield',
       '#title' => $this->t('Username'),
       '#size' => 60,
-      '#maxlength' => USERNAME_MAX_LENGTH,
+      '#maxlength' => UserInterface::USERNAME_MAX_LENGTH,
       '#attributes' => [
         'autocorrect' => 'none',
         'autocapitalize' => 'none',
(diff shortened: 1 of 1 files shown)


modules/checkout/src/Plugin/Commerce/CheckoutPane/Login.php
```diff
@@ -289,88 +290,171 @@ public function buildPaneForm(array $pane_form, FormStateInterface $form_state,
...
+   *   The pane form, containing the following basic properties:
+   *   - #parents: Identifies the position of the pane form in the overall
+   *     parent form, and identifies the location where the field values are
+   *     placed within $form_state->getValues().
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state of the parent form.
+   * @param array $complete_form
+   *   The complete form structure.
+   */
+  protected function validateRegister(array &$pane_form, FormStateInterface $form_state, array &$complete_form) {
```
bojanz, line 372: It's odd that we have a validateRegister() method that does no validation, the entire logic is in the oddly named getAndValidateRegisterUserValues, which is only used in one place. What's the reasoning behind that split?

Berdir: Yeah, the names are bad, suggestions welcome. The reason for the split is being able to change the user values, see my example. For that to work there needs to be a way to do something between bulding the values array and creating the entity. Another approach would be an "alter" method between or/and an event.

PR #723, opened 2017-04-21
Issue #2871569 by bucefal91: Fixing the order field formatters for views

bucefal91:
Issue: https://www.drupal.org/node/2871569

### DIFF
--- a/modules/order/src/Plugin/Field/FieldFormatter/OrderItemTable.php
+++ b/modules/order/src/Plugin/Field/FieldFormatter/OrderItemTable.php
@@ -23,14 +23,19 @@ class OrderItemTable extends FormatterBase {
    * {@inheritdoc}
    */
   public function viewElements(FieldItemListInterface $items, $langcode) {
-    $order = $items->getEntity();
-    return [
-      '#type' => 'view',
-      // @todo Allow the view to be configurable.
-      '#name' => 'commerce_order_item_table',
-      '#arguments' => [$order->id()],
-      '#embed' => TRUE,
-    ];
+    $element = [];
+    if ($items->count() > 0) {
+      $order = $items->getEntity();
+
+      $element[] = [
+        '#type' => 'view',
+        // @todo Allow the view to be configurable.
+        '#name' => 'commerce_order_item_table',
+        '#arguments' => [$order->id()],
+        '#embed' => TRUE,
+      ];
+    }
+    return $element;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/order/src/Plugin/Field/FieldFormatter/OrderTotalSummary.php
```diff
@@ -76,10 +76,15 @@ public static function create(ContainerInterface $container, array $configuratio
    */
   public function viewElements(FieldItemListInterface $items, $langcode) {
     $order = $items->getEntity();
-    return [
-      '#theme' => 'commerce_order_total_summary',
-      '#totals' => $this->orderTotalSummary->buildTotals($order),
-    ];
+
+    $element = [];
+    foreach ($items as $delta => $item) {
```
bojanz, line 81: This doesn't make sense, we'd just repeat the same summary for every delta. Let's just hardcode the delta 0, since we know that a price field can't have more than 1. We want $elements[0] = [...

modules/order/src/Plugin/Field/FieldFormatter/OrderItemTable.php
```diff
@@ -23,14 +23,19 @@ class OrderItemTable extends FormatterBase {
...
-    $order = $items->getEntity();
-    return [
-      '#type' => 'view',
-      // @todo Allow the view to be configurable.
-      '#name' => 'commerce_order_item_table',
-      '#arguments' => [$order->id()],
-      '#embed' => TRUE,
-    ];
+    $element = [];
+    if ($items->count() > 0) {
```
bojanz, line 27: The count change is a behavior change unrelated to the addressed bug, no?

modules/order/src/Plugin/Field/FieldFormatter/OrderTotalSummary.php
bucefal91: Hello! Yes, I see your logic. Mine was: normally it should just 1 item, but if somebody "up there went crazy" and made cardinality of this field more than 1, then it's what I am expected to do: print total X times. However, it's such an edge case that i have no preference between my logic and yours.

modules/order/src/Plugin/Field/FieldFormatter/OrderItemTable.php
bucefal91: I see it as related. If the items are empty (count() == 0), then do not even bother inserting a view. Otherwise insert 1 single view at 0 delta.

Basically, if some $items are empty then we shouldn't return anything at all in $element[0]. Otherwise it may appear misleading - empty items do get formatted into some renderable array.

I am thinking right now.. Since I do not know well inner kitchen of commerce 8.x maybe the view "commerce_order_item_table" has some "empty" logic?

Well, again, I made what I deemed the best given my limited knowledge about the context and have no preference here on whether to keep that count(). :)

bojanz:
Committed in 90979994cf3cdf3f587b9916299897f1734cb8f2.

PR #725, opened 2017-04-27
#2873445 CheckoutFlowBase actions should use #default_value instead of #value

ReplicaObscura:
Resolves https://www.drupal.org/node/2873445 (I think)


bojanz:
There is no example in core of a submit button using #default_value, it's always #value.
I think our fix needs to be different.

mglaman:
Per comments in the Drupal.org issue, all seems fine.

PR #732, opened 2017-05-05
issue #2859834 implementation of links in checkout progress block

luksak:


mglaman:
```
    1)
    Drupal\Tests\commerce_checkout\Functional\CheckoutOrderTest::testGuestOrderCheckout
    Behat\Mink\Exception\ExpectationException: 0 elements matching css
    &quot;.block-commerce-checkout-progress li.checkout-progress--step &gt;
    a&quot; found on the page, but should be 1.
```

We need to come up with a better selector. Because links enabled this breaks. Disabled it won't.

bojanz:
Continued on the issue (https://www.drupal.org/project/commerce/issues/2859834)

PR #734, opened 2017-05-14
Don't compare years greater than as string

Kingdutch:
> Since january 1st 2015 all digital goods sold to EU customers must use the customer zone. For example, an ebook sold to Germany needs to have German VAT applied.

The date returned by `format('Y')` should be an integer which would make php change 2015 into an integer but it's a bit weird not to just do that ourselves.


mglaman:
Re-running test. Failure was not due to change. Gremlin in JS payment checkout test.

mglaman:
Weird, this causes `Drupal\Tests\commerce_payment\FunctionalJavascript\PaymentCheckoutTest::testCheckoutWithNewPaymentMethod` to explode. And we have no way to export the browser output artifacts, yet.

bojanz:
Committed in d13fd1b26a845b7f8ec85437e07d1b79d60471fb, tests seem to be fine now.

PR #737, opened 2017-05-15
Issue #2878586: Fixed country code for Slovakia in European Union VAT tax type.

mitrpaka:


bojanz:
D'oh. Thank you for catching my sloppiness :) I need to re-validate that entire dataset before RC1.

PR #739, opened 2017-05-23
Avoid fatal error in OrderForm when customer was deleted

Berdir:
https://www.drupal.org/node/2880526

### DIFF
--- a/modules/order/src/Form/OrderForm.php
+++ b/modules/order/src/Form/OrderForm.php
@@ -121,8 +121,13 @@ class OrderForm extends ContentEntityForm {
       $form['uid']['#group'] = 'customer';
     }
     else {
-      $user_link = $order->getCustomer()->toLink()->toString();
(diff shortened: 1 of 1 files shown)


modules/order/src/Form/OrderForm.php
```diff
@@ -121,8 +121,13 @@ public function form(array $form, FormStateInterface $form_state) {
       $form['uid']['#group'] = 'customer';
     }
     else {
-      $user_link = $order->getCustomer()->toLink()->toString();
-      $form['customer']['uid'] = $this->fieldAsReadOnly($this->t('Customer'), $user_link);
+      if ($user = $order->getCustomer()) {
```
onatalushko, line 124: This kind of fix is wrong way to fix this bug.
Propper way described in https://www.drupal.org/node/2898218#comment-12231013

bojanz:
Continued on the issue.

PR #740, opened 2017-05-23
Issue #2880927 by mglaman: Add to cart form should load attribute by …

mglaman:
…the field's target bundle not based on its field name


bojanz:
Committed in da0c182fe926b9be8c45b3e7425044a80625b11b.

PR #742, opened 2017-05-31
Issue #2882861 by mglaman: Fire an event to filter gateways when load…

mglaman:
…ing them for an order


bojanz:
Committed in 5aa9dd3196d5a4e5b5a820216a22024788de53bc.

PR #743, opened 2017-06-02
Prefix classes needed for JS functionality with 'js-'

gauravjeet:
https://www.drupal.org/node/2877358

### DIFF
--- a/modules/cart/commerce_cart.module
+++ b/modules/cart/commerce_cart.module
@@ -238,5 +238,5 @@ function commerce_cart_views_data_alter(array &$data) {
  * Prepares variables for the cart block element template.
  */
 function template_preprocess_commerce_cart_block(&$variables) {
-  $variables['attributes']['class'][] = 'cart--cart-block';
+  $variables['attributes']['class'][] = 'js-cart--cart-block';
 }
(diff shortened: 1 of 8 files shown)


modules/cart/commerce_cart.module
```diff
@@ -238,5 +238,5 @@ function commerce_cart_views_data_alter(array &$data) {
  * Prepares variables for the cart block element template.
  */
 function template_preprocess_commerce_cart_block(&$variables) {
-  $variables['attributes']['class'][] = 'cart--cart-block';
+  $variables['attributes']['class'][] = 'js-cart--cart-block';
```
mglaman, line 241: Shouldn't we also keep the non `js-` variant? for theming

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowWithPanesBase.php
```diff
@@ -237,7 +237,7 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
         ['data' => $this->t('Settings'), 'colspan' => 2],
       ],
       '#attributes' => [
-        'class' => ['checkout-pane-overview'],
+        'class' => ['js-checkout-pane-overview'],
```
mglaman, line 240: See previous comment

modules/payment/src/PluginForm/PaymentOffsiteForm.php
```diff
@@ -106,7 +106,7 @@ protected function buildRedirectForm(array $form, FormStateInterface $form_state
    */
   public static function processRedirectForm(array $element, FormStateInterface $form_state, array &$complete_form) {
     $complete_form['#action'] = $element['#action'];
-    $complete_form['#attributes']['class'][] = 'payment-redirect-form';
+    $complete_form['#attributes']['class'][] = 'js-payment-redirect-form';
```
mglaman, line 109: See previous

mglaman (changes requested):
We should add the `js-` class name along with the existing. Just because people will be theming against the existing CSS classes. @bojanz thoughts?

PR #745, opened 2017-06-06
Issue #2875370: Add 'Order total items' promotion condition plugin

steveoliver:
See https://www.drupal.org/node/2875370.


bojanz:
https://www.drupal.org/node/2875370 has a reroll.

PR #746, opened 2017-06-09
create coupon widget fails of no promotion entity

finnef:
Fix for broken widget when creating new coupons
DO: https://www.drupal.org/node/2884849


bojanz:
That would still crash in the non-coupon case. Pushed an alternative fix in 9897285814cd2b412858204d9aeb534f0954431e.

PR #748, opened 2017-06-16
Issue #2886832 by quex: Missing implementation of hook_user_cancel br…

quexpl:
PR for https://www.drupal.org/node/2886832

### DIFF
--- a/modules/order/commerce_order.module
+++ b/modules/order/commerce_order.module
@@ -191,6 +191,7 @@ function commerce_order_user_predelete(EntityInterface $entity) {
     ->getStorage('commerce_order')
     ->loadByProperties([
       'uid' => $entity->id(),
+      'state' => 'draft',
     ]);
 
   foreach ($list as $commerce_order) {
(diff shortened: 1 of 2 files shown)


bojanz:
We can delete draft orders, but all others must remain.

quexpl:
Yeah, that make sense, I added condition.

modules/order/commerce_order.module
```diff
@@ -181,3 +181,41 @@ function commerce_order_mail($key, &$message, $params) {
...
+    $commerce_order->delete();
+  }
+}
+
+/**
+ * Implements hook_user_cancel().
+ */
+function commerce_order_user_cancel($edit, $account, $method) {
+  switch ($method) {
+    case 'user_cancel_reassign':
```
mglaman, line 214: Wouldn't we want to also run this at `user_cancel_delete` method, too? So that orders don't have a broken reference.

mglaman:
Also, there's no tests

modules/order/commerce_order.module
quexpl: hook_user_cancel is not invoked for user_cancel_delete method, https://github.com/drupal/drupal/blob/8.4.x/core/modules/user/user.module#L698 that's why i skipped it.

quexpl:
I will write tests, thanks for feedback

bojanz:
Let's continue on the issue.

PR #749, opened 2017-06-21
Checkout complete page improvements

gauravjeet:
https://www.drupal.org/node/2869452


mglaman:
The path needs to use `.id`, as the order number may not be the ID. Tests pass because by default it is the same.

haringsrob:
I think we can combine the texts of this PR with #788?

PR #750, opened 2017-06-22
Attribute widget should respect if order item has a pre-selected purchased entity

mglaman:
…tem has a pre-selected purchased entity


mglaman:
Committed in c93123ffda8566abe68abe7bc95aa568e5595361

PR #755, opened 2017-07-03
Issue #2794909 by drugan, mglaman, Frank HH-germany, agoradesign, iam…

mglaman:
…puma: Make the quantity field configurable to include HTML properties such as 'Step' property

### DIFF
--- a/modules/cart/src/Plugin/views/field/EditQuantity.php
+++ b/modules/cart/src/Plugin/views/field/EditQuantity.php
@@ -88,18 +88,27 @@ class EditQuantity extends FieldPluginBase {
 
     $form[$this->options['id']]['#tree'] = TRUE;
     foreach ($this->view->result as $row_index => $row) {
+      /** @var \Drupal\commerce_order\Entity\OrderItemInterface $order_item */
       $order_item = $this->getEntity($row);
+
+      /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
+      $form_display = commerce_get_entity_display('commerce_order_item', $order_item->bundle(), 'form');
+      $quantity_component = $form_display->getComponent('quantity');
+
       $quantity = $order_item->getQuantity();
+      if ($quantity_component['settings']['step'] == 1) {
+        $quantity = (int) $quantity;
+      }
 
       $form[$this->options['id']][$row_index] = [
         '#type' => 'number',
         '#title' => $this->t('Quantity'),
         '#title_display' => 'invisible',
-        '#default_value' => round($quantity),
+        '#default_value' => $quantity,
         '#size' => 4,
-        '#min' => 1,
+        '#min' => 0,
         '#max' => 9999,
-        '#step' => 1,
+        '#step' => $quantity_component['settings']['step'],
       ];
     }
     // Replace the form submit button label.
--- a/modules/order/src/Entity/OrderItem.php
+++ b/modules/order/src/Entity/OrderItem.php
@@ -261,6 +261,7 @@ class OrderItem extends ContentEntityBase implements OrderItemInterface {
       ->setDescription(t('The number of purchased units.'))
       ->setReadOnly(TRUE)
       ->setSetting('unsigned', TRUE)
+      ->setSetting('min', 0)
       ->setDefaultValue(1)
       ->setDisplayOptions('form', [
         'type' => 'number',
--- a/modules/product/config/optional/core.entity_form_display.commerce_order_item.default.default.yml
+++ b/modules/product/config/optional/core.entity_form_display.commerce_order_item.default.default.yml
@@ -6,6 +6,9 @@ dependencies:
   enforced:
     module:
       - commerce_product
+  module:
+    - commerce
+    - commerce_price
 id: commerce_order_item.default.default
 targetEntityType: commerce_order_item
 bundle: default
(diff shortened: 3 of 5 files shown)


modules/cart/src/Plugin/views/field/EditQuantity.php
```diff
@@ -88,18 +88,27 @@ public function viewsForm(array &$form, FormStateInterface $form_state) {
 
     $form[$this->options['id']]['#tree'] = TRUE;
     foreach ($this->view->result as $row_index => $row) {
+      /** @var \Drupal\commerce_order\Entity\OrderItemInterface $order_item */
       $order_item = $this->getEntity($row);
+
+      /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
+      $form_display = commerce_get_entity_display('commerce_order_item', $order_item->bundle(), 'form');
```
mglaman, line 95: @bojanz in general, I believe we need to have the cart form mimic the order item form display as much as possible. This kind of hackishly does that. Would rather save for a follow up to limit scope.

modules/order/src/Entity/OrderItem.php
```diff
@@ -261,6 +261,7 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setDescription(t('The number of purchased units.'))
       ->setReadOnly(TRUE)
       ->setSetting('unsigned', TRUE)
+      ->setSetting('min', 0)
```
mglaman, line 264: Setting this prevents negative quantity. Kind of out of scope, but part of bigger problem.

modules/product/config/optional/core.entity_form_display.commerce_order_item.default.default.yml
```diff
@@ -19,17 +22,21 @@ content:
       size: 60
       placeholder: ''
     third_party_settings: {  }
+    region: content
   quantity:
-    type: number
+    type: commerce_number
```
mglaman, line 27: Setting default will cause cart form to respect step.

finnef:
Code reviewed and OK. Lots of things you could add, but let's keep the scope as small as possible and get this committed.
Test on my Commerce installation worked fine. Like the improvement in UI.

modules/order/src/Entity/OrderItem.php
```diff
@@ -261,6 +261,7 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
       ->setDefaultValue(1)
       ->setDisplayOptions('form', [
         'type' => 'number',
```
bojanz, line 267: The display options were not updated to use commerce_number instead of number.

src/Plugin/Field/FieldWidget/CommerceNumberWidget.php
```diff
@@ -0,0 +1,88 @@
...
+ *   id = "commerce_number",
+ *   label = @Translation("Commerce number field"),
+ *   field_types = {
+ *     "integer",
+ *     "decimal",
+ *     "float"
+ *   }
+ * )
+ */
+class CommerceNumberWidget extends NumberWidget {
```
bojanz, line 22: We never prefix classes with Commerce.

src/Plugin/Field/FieldWidget/CommerceNumberWidget.php
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
+    return [
+      'placeholder' => '',
```
bojanz, line 29: Placeholder is already defined by the parent method.

src/Plugin/Field/FieldWidget/CommerceNumberWidget.php
```diff
@@ -0,0 +1,88 @@
...
+use Drupal\Core\Field\Plugin\Field\FieldWidget\NumberWidget;
+
+/**
+ * Plugin implementation of the 'commerce_number' widget.
+ *
+ * @FieldWidget(
+ *   id = "commerce_number",
+ *   label = @Translation("Commerce number field"),
+ *   field_types = {
+ *     "integer",
```
bojanz, line 16: Doesn't make sense to allow this widget on integer fields.

src/Plugin/Field/FieldWidget/CommerceNumberWidget.php
```diff
@@ -0,0 +1,88 @@
...
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Field\Plugin\Field\FieldWidget\NumberWidget;
+
+/**
+ * Plugin implementation of the 'commerce_number' widget.
+ *
+ * @FieldWidget(
+ *   id = "commerce_number",
+ *   label = @Translation("Commerce number field"),
```
bojanz, line 14: Label is a bit ugly. Let's go with "Number field (Commerce)"

src/Plugin/Field/FieldWidget/CommerceNumberWidget.php
```diff
@@ -0,0 +1,88 @@
...
+      case 'decimal':
+        $step = (string) pow(0.1, $field_settings['scale']);
+        break;
+
+      case 'float':
+        $step = 'any';
+        break;
+    }
+
+    $element['step'] = [
```
bojanz, line 52: We should add a comment explaining why we're not using the commerce_number field (can't use both the HTML5 input element and have locale-specific inputs with commas like commerce_number allows)

bojanz (changes requested):

src/Plugin/Field/FieldWidget/CommerceNumberWidget.php
```diff
@@ -0,0 +1,88 @@
...
+      '#type' => 'number',
+      '#min' => 0,
```
bojanz: We can take #min and #max from field settings, like the parent does.

bojanz:
Commited in 8becee97bdb7f1f0ec86215b789dea9f4b0ae8e1.

PR #758, opened 2017-07-08
Multiple add to cart forms with different variation types throws exception

mglaman:


modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php
```diff
@@ -182,38 +172,40 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
...
         }
       }
+      if ($match) {
+        $current_variation = $variation;
+        break;
+      }
     }
-
-    return $current_variation;
+    $form_state->setValueForElement($element['variation'], $current_variation->id());
```
mglaman: This does not persist. It only sets the value during the validation cycle, which causes test fails.

I think `commerce_product_variation_select` element is what we need. Handle loadEnabled, filtering access. Displaying attributes and resolving final variant ID. So we do not need to attempt to piece the attributes to a variation within the widget.

PR #760, opened 2017-07-12
Expand the commerce_profile_select form element with the ability to reuse profiles

bojanz:
Continuation of #631, done by czigor and mglaman.

### DIFF
--- a/modules/order/src/Element/ProfileSelect.php
+++ b/modules/order/src/Element/ProfileSelect.php
@@ -3,9 +3,11 @@
 namespace Drupal\commerce_order\Element;
 
 use Drupal\commerce\Element\CommerceElementTrait;
+use Drupal\Component\Utility\Html;
+use Drupal\Component\Utility\NestedArray;
 use Drupal\Core\Entity\Entity\EntityFormDisplay;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\Render\Element\RenderElement;
+use Drupal\Core\Render\Element\FormElement;
 use Drupal\profile\Entity\ProfileInterface;
 
 /**
@@ -18,15 +20,14 @@ use Drupal\profile\Entity\ProfileInterface;
  *   '#default_value' => $profile,
  *   '#default_country' => 'FR',
  *   '#available_countries' => ['US', 'FR'],
+ *   '#profile_type' => 'customer',
+ *   '#owner_uid' => \Drupal::currentUser()->id(),
  * ];
  * @endcode
- * To access the profile in validation or submission callbacks, use
- * $form['billing_profile']['#profile']. Due to Drupal core limitations the
- * profile can't be accessed via $form_state->getValue('billing_profile').
  *
- * @RenderElement("commerce_profile_select")
+ * @FormElement("commerce_profile_select")
  */
-class ProfileSelect extends RenderElement {
+class ProfileSelect extends FormElement {
 
   use CommerceElementTrait;
 
@@ -40,9 +41,14 @@ class ProfileSelect extends RenderElement {
       '#default_country' => NULL,
       // A list of country codes. If empty, all countries will be available.
       '#available_countries' => [],
+      '#title' => t('Select a profile'),
+      '#create_title' => t('+ Enter a new profile'),
 
       // The profile entity operated on. Required.
-      '#default_value' => NULL,
+      '#default_value' => '_new',
+      '#owner_uid' => 0,
+      // Provide default to not break contrib which have outdated elements.
+      '#profile_type' => 'customer',
       '#process' => [
         [$class, 'attachElementSubmit'],
         [$class, 'processForm'],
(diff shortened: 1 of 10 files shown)


modules/order/src/Element/ProfileSelect.php
```diff
@@ -18,15 +20,14 @@
...
  * @endcode
- * To access the profile in validation or submission callbacks, use
- * $form['billing_profile']['#profile']. Due to Drupal core limitations the
- * profile can't be accessed via $form_state->getValue('billing_profile').
  *
- * @RenderElement("commerce_profile_select")
+ * @FormElement("commerce_profile_select")
  */
-class ProfileSelect extends RenderElement {
+class ProfileSelect extends FormElement {
```
bojanz, line 30: This is reverting the fix from #2857051.

Might have started by accident, since this PR is older than that fix.

modules/order/src/Element/ProfileSelect.php
```diff
@@ -76,30 +102,115 @@ public function getInfo() {
...
+      $element['edit_button'] = [
+        '#type' => 'submit',
+        '#value' => t('Edit'),
+        '#limit_validation_errors' => [],
+        '#ajax' => [
+          'callback' => [get_called_class(), 'ajaxRefresh'],
+          'wrapper' => $wrapper_id,
+        ],
+        '#submit' => [[get_called_class(), 'ajaxSubmit']],
+        '#name' => 'edit_profile',
```
ymhuang0808, line 193: I'm testing the patch when enabling commerce_shipping module. 

Because, the commerce_shipping also uses the `ProfileSelect` element, in checkout page, there are multiple submit buttons with same `#name`. If the `edit_button` element is clicked, the trigger element will be wrong. 

In my testing, when I clicked the "Edit" button, the trigger element was the "Continue to review" button. However, the expected and correct trigger element should be "Edit" button. The problem may result from the same value of `#name` and more detail is in issue [#2546700](https://www.drupal.org/node/2546700). 

In order to prevent the situation, we should have a workaround for the issue to make the value of `#name` unique. 
My rough implementation:

```
$form_name = (isset($element['#parents'])) ? implode('_', $element['#parents']) . '_' : '';
$form_name .= 'edit_profile';
```

And the value of the `#name` is `$form_name`.

modules/order/src/Element/ProfileSelect.php
```diff
@@ -76,30 +102,115 @@ public function getInfo() {
...
+
+    $id_prefix = implode('-', $element['#parents']);
+    $wrapper_id = Html::getUniqueId($id_prefix . '-ajax-wrapper');
+    $element = [
+      '#tree' => TRUE,
+      '#prefix' => '<div id="' . $wrapper_id . '">',
+      '#suffix' => '</div>',
+      // Pass the id along to other methods.
+      '#wrapper_id' => $wrapper_id,
+      '#element_mode' => $form_state->get('element_mode') ?: 'view',
```
ymhuang0808, line 142: If the multiple `ProfileSelect` form elements exist in the same form (eg. in checkout page, the shipping info and billing info), the `element_mode` in `$form_state` should be separated by the form parents. 

For example, commerce_shipping module is installed, and there are two form elements in the same checkout form page. Once a `edit_button` within shipping info (or payment info) is clicked, the same `element_mode` value in `$form_state` is set into _edit_ by `ajaxSubmit()`.  Therefore, `#element_mode` in shipping info and payment info are _edit_. It caused another form element in payment info doesn't contain `edit_button`. 

Perhaps, the element mode name should be with the form parents.

```
$form_state->get('element_mode-' . $id_prefix . '-edit_button')
```

modules/order/src/Element/ProfileSelect.php
```diff
@@ -133,9 +264,33 @@ public static function validateForm(array &$element, FormStateInterface $form_st
...
+    $element = NestedArray::getValue($form, array_slice($triggering_element['#array_parents'], 0, -1));
+    return $element;
+  }
+
+  /**
+   * Ajax submit callback.
+   */
+  public static function ajaxSubmit(array &$form, FormStateInterface $form_state) {
+    $triggering_element = $form_state->getTriggeringElement();
+    $form_state->set('element_mode', $triggering_element['#element_mode']);
```
ymhuang0808, line 292: As previous mention, the `element_mode` key in form state should be changed into:

```
    $parents = implode('-', $triggering_element['#parents']);
    $element_mode_name = 'element_mode-' . $parents;

    $form_state->set($element_mode_name, $triggering_element['#element_mode']);
```

ymhuang0808 (changes requested):
Support multiple form elements in the same form for commerce_shipping module

ymhuang0808:
Hi @bojanz I would like to enhance the PR, should I create a PR into 8.x-2.x or profile_reuse branch?

mglaman:
@ymhuang0808 hm, yeah that's the problem with Github. I'll wait to see what @bojanz says.

But to follow up: one of the main issues is that the element does not work if there are multiple elements on the same page. Correct?

ymhuang0808:
@mglaman Yes. I'm tring to fix that.

ymhuang0808:
@bojanz and @mglaman The enhancement for the PR is ready to review on https://github.com/fliegen/commerce/commit/57c11175d279fe55db9454b4274608195211aeb8. Please take a look. I think I should create a new PR into profile_reuse branch.

The commit has a few changes:

- Make the "Edit" button available in multiple form elements
- If the value of address field is changed, it will create a new customer profile entity
- Fix in edit mode, the address field is not changed with profile_selection
- Add functional JavaScript tests for multiple form elements

BTW, in order to support the ProfileSelect, I also revised the commerce_shipping. 

Thanks.

bojanz:
This PR is outdated, the issue has many updated patches.

PR #761, opened 2017-07-17
Issue #2891257 by vasike: Add the fixed amount order item offer

vasike:


bojanz:
Committed in 5a69573425464f574a82b9a273c052b7b7d1e0c3 with a simplified test.

PR #762, opened 2017-07-17
Issue #2895457: Null currency on product page

heddn:


modules/price/src/Plugin/Field/FieldFormatter/PriceDefaultFormatter.php
```diff
@@ -155,16 +155,18 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
     foreach ($items as $delta => $item) {
-      $currency = $currencies[$item->currency_code];
-      $elements[$delta] = [
-        '#markup' => $this->numberFormatter->formatCurrency($item->number, $currency),
-        '#cache' => [
-          'contexts' => [
-            'languages:' . LanguageInterface::TYPE_INTERFACE,
-            'country',
+      $currency = isset($currencies[$item->currency_code]) ? $currencies[$item->currency_code] : NULL;
+      if ($currency) {
```
mglaman, line 159: How could a price field have a currency value which has yet not been imported?

heddn:
The currency was deleted but existing product entered data still referenced the currency code. I deleted CAN before switching it all over to USD. We should probably default the value to 'n/a' or something like that.

This was on the product edit form, because IEF loads the variations.

bojanz:
We can't babysit broken data, one formatter doesn't even scratch the surface.

Instead, we need to prevent the data from being broken (by forbidding deletion of anything that's in use).
There are several open issues about this. My initial thoughts are at https://www.drupal.org/node/2841471#comment-11971257
That's what we need to solve.

PR #763, opened 2017-07-29
Use new EntityPublished Interface and Trait, remove Save and .. buttons from ProductForm

Berdir:
This requires 8.4

### DIFF
--- a/modules/product/src/Entity/ProductInterface.php
+++ b/modules/product/src/Entity/ProductInterface.php
@@ -5,12 +5,13 @@ namespace Drupal\commerce_product\Entity;
 use Drupal\commerce_store\Entity\EntityStoresInterface;
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityChangedInterface;
+use Drupal\Core\Entity\EntityPublishedInterface;
 use Drupal\user\EntityOwnerInterface;
 
 /**
  * Defines the interface for products.
  */
-interface ProductInterface extends ContentEntityInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoresInterface {
+interface ProductInterface extends ContentEntityInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoresInterface, EntityPublishedInterface {
 
   /**
    * Gets the product title.
--- a/modules/product/src/Entity/ProductInterface.php
+++ b/modules/product/src/Entity/ProductInterface.php
@@ -32,26 +32,6 @@ interface ProductInterface extends ContentEntityInterface, EntityChangedInterfac
   public function setTitle($title);
 
   /**
-   * Get whether the product is published.
-   *
-   * Unpublished products are only visible to their authors and administrators.
-   *
-   * @return bool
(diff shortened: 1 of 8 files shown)


bojanz:
Which part requires 8.4? EntityPublishedInterface and trait are in 8.3, no?

Also, do we need an update function like this one:
```
/**
 * Set the 'published' entity key.
 */
function comment_update_8301() {
  $definition_update_manager = \Drupal::entityDefinitionUpdateManager();
  $entity_type = $definition_update_manager->getEntityType('comment');
  $keys = $entity_type->getKeys();
  $keys['published'] = 'status';
  $entity_type->set('entity_keys', $keys);
  $definition_update_manager->updateEntityType($entity_type);
}
```
I was confused about that the last time I looked into it.
Looks the same in node_update_8301(), but I thought the entity type would be automatically updated on cache clear.

modules/product/src/Entity/ProductInterface.php
```diff
@@ -5,12 +5,13 @@
...
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityChangedInterface;
+use Drupal\Core\Entity\EntityPublishedInterface;
 use Drupal\user\EntityOwnerInterface;
 
 /**
  * Defines the interface for products.
  */
-interface ProductInterface extends ContentEntityInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoresInterface {
+interface ProductInterface extends ContentEntityInterface, EntityChangedInterface, EntityOwnerInterface, EntityStoresInterface, EntityPublishedInterface {
```
bojanz, line 14: We also need to remove the original isPublished / setPublished methods that are still on the interface.

Berdir:
Oh right, thought the Trait was new in 8.4.x but that's only EditorialContentEntityBase, which we don't need as products are (currently) not revisionable.

Right, an update function is needed, otherwise you'd need to run entity-updates. It doesn't actually change anything as status is already a status key (I don't 100% understand why the separate status key was added, to be honest)

What's your plan with 8.3/8.4? I can work more on this if you want to merge it, otherwise I'll probably only update it when 8.4 is out.

bojanz:
Let's merge it soon. I figure it's better to sync with (future) core before 2.0, rather than change the UX in the first next release. Of course, that will only work if we can retain 8.3 compatibility, I won't require 8.4 until it gets a stable release (around Commerce 2.1 probably).

Berdir:
Pushed an update and the footer region.

Still completely untested, don't actually have a local installation here right now :)

You might want to keep an eye on https://www.drupal.org/node/2892304.

And on https://www.drupal.org/node/2893740, feedback on that is welcome ;)

bojanz:
Thanks for the links! Following.

I'll test the PR manually as soon as it goes green on Travis.

bojanz:
Fixed the test fail, committed the EntityPublishedInterface part in https://www.drupal.org/node/2870672.
Will do the ProductForm part next.

bojanz:
Continued on the issue https://www.drupal.org/node/2898258

PR #764, opened 2017-07-29
Issue #2896128: Don't show an error if the coupon is already applied.

mitrpaka:


modules/promotion/src/Element/CouponRedemptionForm.php
```diff
@@ -211,7 +211,7 @@ public static function validateForm(array &$element, FormStateInterface $form_st
     $order = $order_storage->load($element['#order_id']);
     foreach ($order->get('coupons') as $item) {
       if ($item->target_id == $coupon->id()) {
-        $form_state->setErrorByName($coupon_code_path, t('The provided coupon code is invalid.'));
+        drupal_set_message(t('Coupon %code already applied.', ['%code' => $coupon->getCode()]));
```
bojanz, line 214: Let's not set a message at all. We don't have a confirmation message on initial success either.

Also, this code now makes validation pass, and the applyCoupon() submit function get called. That function needs to make sure it's re-entrant, by only appending the item if it doesn't exist. Otherwise we'll get duplicate IDs in the field.

bojanz:
Let's continue on the issue https://www.drupal.org/project/commerce/issues/2896128.
When there's an issue, there's no need for a PR anymore.

PR #765, opened 2017-07-30
Travis CI: Define distribution for builds to Ubuntu Precise.

mitrpaka:
As Travis have changed the default distribution for builds to Ubuntu Trusty 14.04 (see https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming?utm_source=web&utm_medium=banner&&utm_campaign=trusty-default), CI will fail on forks unless distribution for builds set to Precise.

Error on Trusty (comes with Phantom JS 2.1.1 by default):
phantomjs: error while loading shared libraries: libicui18n.so.48: cannot open shared object file: No such file or directory

The command "phantomjs --version" failed and exited with 127 during .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ language: php
 
 sudo: false
 
+dist: precise
+
 php:
   - 7
   - 5.6


bojanz:
```
  - wget https://s3.amazonaws.com/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -O $PWD/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2
  - tar -xvf $PWD/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -C $PWD/travis-phantomjs
  - export PATH=$PWD/travis-phantomjs:$PATH
```
We are explicitly using the PhantomJS for precise here, did you remove that part before testing with the phantomjs that comes with Trusty?

mitrpaka:
Did not remove those in the first round which explains error messages shown. If PhantomJS install part for precise removed, Travis using Trusty runs successfully with PHP 7 setup but has timeout (The job exceeded the maximum time limit for jobs, and has been terminated.) and stability (MySQL has gone away) issues with PHP 5.x setups. Therefore continued use of Precise recommended for time being.

Initial point was that at time of Travis transition from Precise to Trusty, in order to have consistency between commerce and fork repositories (old and new), distribution used for Travis builds would be explicitly set to Precise (as also guided in Travis docs).

Will leave this now for others to comment further as they have more experience with Travis and its setup than I have currently (Travis is "new kid on the block" still for me ;-)

bojanz:
Okay, let's go with this for now, then investigate a move to 14.04 in a different commit. Thanks!

subhojit777:
Re https://github.com/drupalcommerce/commerce/pull/765#issuecomment-318927781:

> PHP 7 setup but has timeout (The job exceeded the maximum time limit for jobs, and has been terminated.)

For me too the tests were timing out for Ubuntu 14.04 and PhantomJS 2.1.1. Restricting the OS version to 12.04 and PhantomJS 2.0.0 helped me. Thanks @mitrpaka

PR #767, opened 2017-08-02
Issue #2894792: Add Adjustment:getPercentage()

mitrpaka:
Kept $percentage as float (not integer) as percentage may not always be integer.

### DIFF
--- a/modules/order/src/Adjustment.php
+++ b/modules/order/src/Adjustment.php
@@ -50,7 +50,7 @@ final class Adjustment {
   /**
    * The adjustment percentage.
    *
-   * @var float
+   * @var string
    */
   protected $percentage;
 
@@ -130,7 +130,7 @@ final class Adjustment {
   /**
    * Gets the adjustment percentage.
    *
-   * @return float|null
+   * @return string|null
    *   The adjustment percentage or NULL if not available.
    */
   public function getPercentage() {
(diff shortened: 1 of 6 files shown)


modules/order/src/Adjustment.php
```diff
@@ -48,6 +48,13 @@
   protected $included = FALSE;
 
   /**
+   * The adjustment percentage.
+   *
+   * @var float
```
bojanz: decimals are always strings in Commerce, due to our bcmath-based arithmetic.

modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderItemPercentageOff.php
```diff
@@ -32,6 +32,7 @@ public function apply(EntityInterface $entity, PromotionInterface $promotion) {
       // @todo Change to label from UI when added in #2770731.
       'label' => t('Discount'),
       'amount' => $adjustment_amount->multiply('-1'),
+      'percentage' => $this->getAmount() * 100,
```
bojanz: Commerce always stores percentages in decimal form, so the multiplication by 100 is not needed

bojanz:
Committed in 289a07ebcb6e90b6c4fc828c20a2664ec78608cd.

PR #768, opened 2017-08-09
Issue #2901000 by mglaman: Twig filter `commerce_price_format` does n…

mglaman:
…ot do rounding


bojanz:
We decided to fix the intl number formatter instead.

PR #769, opened 2017-08-10
Added empty cart button views plugin.

swickham78:
Adds the empty cart button requested here - https://www.drupal.org/node/2900917

### DIFF
--- a/modules/cart/src/Plugin/views/field/EmptyCartButton.php
+++ b/modules/cart/src/Plugin/views/field/EmptyCartButton.php
@@ -3,7 +3,6 @@
 namespace Drupal\commerce_cart\Plugin\views\field;
 
 use Drupal\commerce_cart\CartManagerInterface;
-use Drupal\commerce_order\Entity\Order;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\views\Plugin\views\field\FieldPluginBase;
--- a/modules/cart/src/Plugin/views/field/EmptyCartButton.php
+++ b/modules/cart/src/Plugin/views/field/EmptyCartButton.php
@@ -1,0 +1,125 @@
+<?php
+
+namespace Drupal\commerce_cart\Plugin\views\field;
+
+use Drupal\commerce_cart\CartManagerInterface;
+use Drupal\commerce_order\Entity\Order;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\views\Plugin\views\field\FieldPluginBase;
+use Drupal\views\Plugin\views\field\UncacheableFieldHandlerTrait;
+use Drupal\views\ResultRow;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines a form element for emptying the cart.
+ *
+ * @ViewsField("commerce_order_item_empty_cart_button")
+ */
+class EmptyCartButton extends FieldPluginBase {
+
+  use UncacheableFieldHandlerTrait;
+
+  /**
+   * The cart manager.
+   *
+   * @var \Drupal\commerce_cart\CartManagerInterface
+   */
+  protected $cartManager;
+
+  /**
+   * Constructs a new EmptyCartButton object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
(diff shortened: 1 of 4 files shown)


swickham78:
Adds the empty cart button requested here - https://www.drupal.org/node/2900917

modules/cart/src/Plugin/views/field/EmptyCartButton.php
```diff
@@ -0,0 +1,125 @@
...
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   */
+  public function viewsFormSubmit(array &$form, FormStateInterface $form_state) {
+    $triggering_element = $form_state->getTriggeringElement();
+    if (!empty($triggering_element['#empty_cart_button'])) {
+      /** @var \Drupal\commerce_order\Entity\OrderInterface $cart */
+      $cart = Order::load($this->view->argument['order_id']->getValue());
```
mglaman: We may as well inject the entity type manager and get storage from there

modules/cart/src/Plugin/views/field/EmptyCartButton.php
```diff
@@ -0,0 +1,125 @@
...
+use Drupal\views\Plugin\views\field\UncacheableFieldHandlerTrait;
+use Drupal\views\ResultRow;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines a form element for emptying the cart.
+ *
+ * @ViewsField("commerce_order_item_empty_cart_button")
+ */
+class EmptyCartButton extends FieldPluginBase {
```
mglaman, line 18: Needs a test

mglaman (changes requested):
Needs test, then little DI nit

modules/cart/src/Plugin/views/field/EmptyCartButton.php
swickham78: I assumed this would just be an optional item a user could add to their Cart Form View and wasn't sure if tests got done in those cases. 

So are you saying it should be changed to be an item that's enabled by default, or keep it disabled by default and the test should enable this item for the View, then test it?

swickham78: @mglaman Just making sure these questions didn't fly under your radar.

mglaman: Hm. So we're not providing it by default. But, I believe tests are our first line of defense when it comes to documentation. Every feature should have a test. That way we know it works and it exists. Otherwise we have features without coverage and without implementation documentation.

A test module should provide a view which implements this button.

swickham78:
@mglaman Automated tests have been added now.

modules/price/commerce_price.module
```diff
@@ -26,7 +26,9 @@ function commerce_price_theme() {
  * Implements hook_ENTITY_TYPE_insert() for 'configurable_language'.
  */
 function commerce_price_configurable_language_insert(ConfigurableLanguage $language) {
-  // Allow the importer to import any found translations for the new language.
-  $importer = \Drupal::service('commerce_price.currency_importer');
-  $importer->importTranslations([$language->getId()]);
+  if (!\Drupal::isConfigSyncing()) {
```
mglaman: Why did we have to add this in?

mglaman (changes requested):
WOOT! Awesome on test. But I have one question on this `commerce_price.module` change

modules/price/commerce_price.module
mglaman: (Just because it looks like a bug fix, which could go into its own commit and fix. Unless triggered here.)

swickham78: I'm really not sure what happened here, all that work is from a commit bradjones1 made for issue #2901821. It's the last commit made before my work after rebasing against branch 8.x-2.x. I may have screwed up the rebase but I'm not sure how this would come up as a result of something I did wrong.

swickham78:
@mglaman I think I have that issue sorted out, looks like the PR no longer contains the changes to the commerce_price module.

mglaman (approved):

mglaman:
I think this looks good, would just want to check if @bojanz wants eyeballs quick.

bojanz:
Committed in 1cea25a727948ed51683924507ebb6e2451a2a46 after addressing the feedback from the issue.

PR #770, opened 2017-08-11
Issue #2901442 by sumanthkumarc: Add a setOffer method to Promotions

sumanthkumarc:


modules/promotion/src/Entity/Promotion.php
```diff
@@ -176,6 +176,16 @@ public function getOffer() {
   /**
    * {@inheritdoc}
    */
+  public function setOffer($target_plugin_id, array $target_plugin_configuration) {
```
mglaman: We should pass an Offer plugin, and then the method knows how to extract the plugin ID and configuration

mglaman (changes requested):

modules/promotion/src/Entity/Promotion.php
bojanz: Agreed.

sumanthkumarc: Thanks for the suggestion.

```php
// Returns array with plugin id and target config value.
$offer_plugin_config_array =  $promotion->get('offer')->first()->getValue();

// Returns offer plugin object of type PromotionOfferBase. Can we get all info from this object???
$offer_plugin_object =  $promotion->getOffer();
```

@mglaman which one should be passed to setOffer method?? as per my knowledge, $offer_object can be enforced for typehint but not sure if we get all reqd. info. from this object. i'm talking about configuration like amount etc. getConfiguration() doesn't help here, gives different keys.

if we pass the $offer_plugin_config_array, i'm not sure how we can enforce if the reqd. values are present and if user knows to get this value from??

mglaman (approved):

mglaman:
@sumanthkumarc can we update the entity's Kernel test to document / test the method?

modules/promotion/src/Entity/PromotionInterface.php
```diff
@@ -96,6 +97,16 @@ public function setOrderTypeIds(array $order_type_ids);
   public function getOffer();
 
   /**
+   * Sets the offer.
+   *
+   * @param \Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\PromotionOfferInterface $offer_plugin
+   *    The offer plugin.
```
bojanz, line 103: Wrong indent of the param description (3 spaces instead of 2)

modules/promotion/src/Entity/PromotionInterface.php
```diff
@@ -96,6 +97,16 @@ public function setOrderTypeIds(array $order_type_ids);
   public function getOffer();
 
   /**
+   * Sets the offer.
+   *
+   * @param \Drupal\commerce_promotion\Plugin\Commerce\PromotionOffer\PromotionOfferInterface $offer_plugin
```
bojanz, line 102: $offer_plugin should be just $offer, we don't suffix like that.

bojanz:
Tweaked the code style, added a test, committed in b7f1cea4ea22b7f3dfae9c16096c73d98342294e.

PR #772, opened 2017-08-12
#2901609 - Provide Search API processor to exclude disabled variations/products.

swickham78:


src/Plugin/search_api/processor/ProductStatus.php
```diff
@@ -0,0 +1,59 @@
...
+ *   },
+ * )
+ */
+class ProductStatus extends ProcessorPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function supportsIndex(IndexInterface $index) {
+    $supported_entity_types = ['commerce_product', 'commerce_product_variation'];
```
mglaman, line 28: bojanz said we should sync with the Search API maintainers about part of this. We should see if their processor could support all entity types which implement EntityPublishedInterface. We'll be there after #763.

However, we'd still need specific for variations. Unless variations implemented EntityPublishedInterface (it has status key.) And it's published status was based on its own value and parent product. In latter case we wouldn't need this.

- [ ] Reach out to Search API maintainers about patching the EntityStatus processor
- [ ] Discuss with @bojanz about Variations implementing EntityPublishedInterface

mglaman (changes requested):
Adding some notes which bojanz and I discussed in Slack after I made the issue.

bojanz:
This plugin is supposed to be in the commerce_product module, not Commerce.

src/Plugin/search_api/processor/ProductStatus.php
bojanz: I'd love to see us refocus this plugin to variations (commerce_product_variation_status) while fixing Search API to automatically handle entity status.

bojanz: Search API issue: https://www.drupal.org/node/2907756

bojanz:
The search API issue has landed, we can repurpose this code to be for variations only.

bojanz:
https://www.drupal.org/project/commerce/issues/2901609#comment-12845988

PR #774, opened 2017-08-14
Issue #2896815: Replace the order summary view with a Twig template

mglaman:


modules/checkout/commerce_checkout.module
```diff
@@ -34,6 +34,13 @@ function commerce_checkout_theme() {
     'commerce_checkout_form__with_sidebar' => [
       'base hook' => 'commerce_checkout_form',
     ],
+    'commerce_checkout_order_summary' => [
+      'variables' => [
+        'order' => NULL,
+        'checkout_flow_plugin_id' => '',
```
mglaman: These are here so that people can do conditionals in the Twig template. For instance, we use this to check if on "order_information" and display "+ applicable taxes" if there are no adjustments. Should we document that somehow here, or in the Twig template?

modules/checkout/src/Plugin/Commerce/CheckoutPane/OrderSummary.php
```diff
@@ -29,10 +30,16 @@ public function defaultConfiguration() {
    * {@inheritdoc}
    */
   public function buildConfigurationSummary() {
-    $view_storage = $this->entityTypeManager->getStorage('view');
-    $view = $view_storage->load($this->configuration['view']);
-    if ($view) {
-      return $this->t('View: @view', ['@view' => $view->label()]);;
+    $use_views = $this->configuration['use_views'];
+    if (!$use_views) {
+      return $this->t('Using template to display order summary');
```
mglaman: Bikeshed summary text.

modules/checkout/src/Plugin/Commerce/CheckoutPane/OrderSummary.php
```diff
@@ -42,6 +49,13 @@ public function buildConfigurationSummary() {
   public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
     $form = parent::buildConfigurationForm($form, $form_state);
 
+    $form['use_views'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Use a View to display order summary'),
+      '#description' => $this->t('This setting allows you to use a View to override the template output of the order summary'),
```
mglaman, line 55: Bikeshed description.

mglaman (commented):
Ready for review!

modules/checkout/commerce_checkout.module
```diff
@@ -34,6 +34,13 @@ function commerce_checkout_theme() {
     'commerce_checkout_form__with_sidebar' => [
       'base hook' => 'commerce_checkout_form',
     ],
+    'commerce_checkout_order_summary' => [
+      'variables' => [
+        'order' => NULL,
```
bojanz: Other commerce_checkout and commerce_order templates use order_entity instead of order.

modules/checkout/config/schema/commerce_checkout.schema.yml
```diff
@@ -70,6 +70,9 @@ commerce_checkout.commerce_checkout_pane.login:
 commerce_checkout.commerce_checkout_pane.order_summary:
   type: commerce_checkout_pane_configuration
   mapping:
+    use_views:
```
bojanz, line 73: Can't we just infer use_views from an empty 'view' key? We usually do it that way.

modules/checkout/templates/commerce-checkout-order-summary.html.twig
```diff
@@ -0,0 +1,28 @@
+{#
+/**
+ * See commerce-order-receipt.html.twig for examples.
+ *
+ * Available variables:
+ * - order_entity: Order entity
+ * - items: The order items
+ * - totals: Result from commerce_order.order_total_summary service
+ * - checkout_flow_plugin_id: The current checkout flow ID.
```
bojanz: What's the point of checkout_flow_plugin_id and checkout_flow_step if we're not using them?

mglaman: See my comment above

> These are here so that people can do conditionals in the Twig template. For instance, we use this to check if on "order_information" and display "+ applicable taxes" if there are no adjustments. Should we document that somehow here, or in the Twig template?

I use them often; we injected them into our own templates.

mglaman: I guess we could add suggestions instead

modules/checkout/commerce_checkout.module
mglaman: I know they do in Twig; but they set it via preprocess, as we do here. But it takes an "argument" of order.

modules/checkout/config/schema/commerce_checkout.schema.yml
mglaman: How would the UI look, then? I added this because then the UI will use `#states` to toggle the dropdown. 

Should the select list be "Display option:" and "Use template" or underneath is a random list of views? UI feels weird that way.

modules/checkout/commerce_checkout.module
```diff
@@ -34,6 +34,12 @@ function commerce_checkout_theme() {
     'commerce_checkout_form__with_sidebar' => [
       'base hook' => 'commerce_checkout_form',
     ],
+    'commerce_checkout_order_summary' => [
+      'variables' => [
+        'order_entity' => NULL,
+        'checkout_flow_step' => '',
```
bojanz, line 40: We don't mention a "checkout flow step" anywhere else in the codebase. It's a checkout step, so let's use "checkout_step"

modules/checkout/commerce_checkout.module
```diff
@@ -55,6 +61,21 @@ function commerce_checkout_theme_suggestions_commerce_checkout_form(array $varia
...
 
 /**
+ * Implements hook_preprocess_HOOK().
+ */
+function commerce_checkout_preprocess_commerce_checkout_order_summary(&$variables) {
+  /** @var \Drupal\commerce_order\Entity\OrderInterface $order */
+  $order = $variables['order_entity'];
+  /** @var \Drupal\commerce_order\OrderTotalSummaryInterface $order_total_summary */
+  $order_total_summary = \Drupal::service('commerce_order.order_total_summary');
+  $variables['totals'] = [
```
bojanz, line 71: This confused me and I had to do a double check, because I expected totals to be the raw array like in the receipt template. Instead, I got the rendered version. For more power let's pass both totals and rendered_totals, we lose nothing by doing so.

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
```diff
@@ -248,11 +248,10 @@ public function buildForm(array $form, FormStateInterface $form_state, $step_id
...
     if ($this->hasSidebar($step_id)) {
       $form['sidebar']['order_summary'] = [
-        '#type' => 'view',
-        '#name' => 'commerce_checkout_order_summary',
-        '#display_id' => 'default',
-        '#arguments' => [$this->order->id()],
-        '#embed' => TRUE,
+        '#theme' => 'commerce_checkout_order_summary',
+        '#order' => $this->order,
+        '#checkout_flow_plugin_id' => $this->getPluginId(),
```
bojanz, line 253: #checkout_flow_plugin_id still passed here even though we removed it.

modules/checkout/src/Plugin/Commerce/CheckoutPane/OrderSummary.php
```diff
@@ -21,18 +21,25 @@ class OrderSummary extends CheckoutPaneBase implements CheckoutPaneInterface {
    */
   public function defaultConfiguration() {
     return [
-      'view' => 'commerce_checkout_order_summary',
+      // @todo remove this, if checkbox not yet, remove `view`
```
bojanz, line 24: Unresolved todo. use_views needs to be removed.

modules/checkout/src/Plugin/Commerce/CheckoutPane/OrderSummary.php
bojanz: "Overrides the checkout order summary template with the output of a View."
Unsure if View should be capitalized, but the wording makes sense to me.

modules/checkout/src/Plugin/Commerce/CheckoutPane/OrderSummary.php
```diff
@@ -70,21 +89,36 @@ public function submitConfigurationForm(array &$form, FormStateInterface $form_s
...
    */
   public function buildPaneForm(array $pane_form, FormStateInterface $form_state, array &$complete_form) {
-    $pane_form['view'] = [
-      '#type' => 'view',
-      '#name' => $this->configuration['view'],
-      '#display_id' => 'default',
-      '#arguments' => [$this->order->id()],
-      '#embed' => TRUE,
-    ];
+    if (!$this->configuration['view']) {
```
bojanz, line 105: Stylistic nitpick: Feels weird for the first condition to be a negation, it requires more energy to mentally parse. I'd always start with if ($this->configuration['view']) then handle the other case in the else.

modules/checkout/src/Plugin/Commerce/CheckoutPane/OrderSummary.php
```diff
@@ -70,21 +89,36 @@ public function submitConfigurationForm(array &$form, FormStateInterface $form_s
...
+      $pane_form['summary'] = [
+        '#theme' => 'commerce_checkout_order_summary',
+        '#order' => $this->order,
+        '#checkout_flow_plugin_id' => $this->getPluginId(),
```
bojanz, line 109: #checkout_flow_plugin_id still passed even though it was removed.

modules/checkout/src/Plugin/Commerce/CheckoutFlow/CheckoutFlowBase.php
```diff
@@ -248,11 +248,10 @@ public function buildForm(array $form, FormStateInterface $form_state, $step_id
     $form['#attached']['library'][] = 'commerce_checkout/form';
     if ($this->hasSidebar($step_id)) {
       $form['sidebar']['order_summary'] = [
-        '#type' => 'view',
-        '#name' => 'commerce_checkout_order_summary',
-        '#display_id' => 'default',
-        '#arguments' => [$this->order->id()],
-        '#embed' => TRUE,
+        '#theme' => 'commerce_checkout_order_summary',
+        '#order' => $this->order,
```
bojanz, line 252: #order was renamed to #order_entity, but neither usage was updated, so the current code crashes.

modules/checkout/templates/commerce-checkout-order-summary.html.twig
```diff
@@ -0,0 +1,27 @@
+{#
+/**
+ * See commerce-order-receipt.html.twig for examples.
+ *
+ * Available variables:
+ * - order_entity: Order entity
+ * - items: The order items
```
bojanz, line 7: "items" doesn't exist.

modules/checkout/templates/commerce-checkout-order-summary.html.twig
```diff
@@ -0,0 +1,27 @@
+{#
+/**
+ * See commerce-order-receipt.html.twig for examples.
```
bojanz, line 3: Missing @file

modules/checkout/templates/commerce-checkout-order-summary.html.twig
```diff
@@ -0,0 +1,27 @@
...
+ * - checkout_flow_step: The current checkout flow step.
+ */
+#}
+{% block order_items %}
+  <table>
+    <tbody>
+    {% for order_item in order_entity.getItems %}
+      <tr>
+        <td>{{ order_item.getQuantity|number_format }}&nbsp;x</td>
+        <td>{{- order_item.label -}}</td>
```
bojanz, line 18: We have a regression here. Previously we showed a rendered entity, now we just show the order item label.
We should be showing the rendered entity when it exists, and the label otherwise as a fallback.

modules/checkout/templates/commerce-checkout-order-summary.html.twig
```diff
@@ -0,0 +1,27 @@
...
+ * See commerce-order-receipt.html.twig for examples.
+ *
+ * Available variables:
+ * - order_entity: Order entity
+ * - items: The order items
+ * - totals: Result from commerce_order.order_total_summary service
+ * - checkout_flow_step: The current checkout flow step.
+ */
+#}
+{% block order_items %}
```
bojanz, line 12: This template is missing a wrapper div with {{ attributes }}, making it impossible to append IDs and classes.

modules/checkout/templates/commerce-checkout-order-summary.html.twig
```diff
@@ -0,0 +1,27 @@
...
+  <table>
```
bojanz, line 13: Should we be using a table for this, or something like flexbox? In any case, the element is missing a class.

bojanz (changes requested):

bojanz:
Addressing the feedback.

bojanz:
Committed in 8c05109cd8078791b9f09cf58669b25585f2b5fb.

PR #776, opened 2017-08-21
Issue #2903490 stop order item percentage promotions from creating negative order totals

swickham78:


modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderItemPercentageOff.php
```diff
@@ -24,17 +25,32 @@ public function apply(EntityInterface $entity, PromotionInterface $promotion) {
...
-    $adjustment_amount = $order_item->getUnitPrice()->multiply($this->getPercentage());
+
+    $item_adjustments = $order_item->getAdjustments();
+    $item_adjustment_total = 0;
+
+    // Previous adjustments on the order item need to be included in
+    // calculations.
+    if (count($item_adjustments)) {
+      foreach ($item_adjustments as $item_adjustment) {
+        $item_adjustment_total += (float) $item_adjustment->getAmount()->getNumber();
```
mglaman, line 36: Why not use the price objects `add` method here?

```
$adjustment_total = new Price('0', $order_item->getUnitPrice()->getCurrencyCode();
foreach ($order_item->getAdjustments() as $adjustment) {
  $adjustment_total = $adjustment_total->add($adjustment->getAmount());
}
```

modules/promotion/src/Plugin/Commerce/PromotionOffer/OrderItemPercentageOff.php
```diff
@@ -24,17 +25,32 @@ public function apply(EntityInterface $entity, PromotionInterface $promotion) {
...
 
-    $order_item->addAdjustment(new Adjustment([
-      'type' => 'promotion',
-      // @todo Change to label from UI when added in #2770731.
-      'label' => t('Discount'),
-      'amount' => $adjustment_amount->multiply('-1'),
-      'percentage' => $this->getPercentage(),
-      'source_id' => $promotion->id(),
-    ]));
+    if (!$adjustment_amount->isZero()) {
```
mglaman, line 44: This won't fix the issue

```
  public function isZero() {
    return Calculator::compare($this->number, '0') == 0;
  }
```

mglaman (changes requested):
I really think we need a test, first, which proves the error.

bojanz:
Let's continue on the issue. The patch will need a rewrite for the updated promotions in 2.8

PR #781, opened 2017-09-01
 [WIP] Expand the "Calculated price" formatter with the ability to show prices with promotions/taxes/fees

mglaman:
Still, things to do

https://www.drupal.org/node/2805549

* [ ] What to do about `PriceCalculatedFormatter` plugin ID?
* [x] Setting for what adjustment types to display in `PriceCalculatedFormatter`
* [ ] Currently no tax test.
* [x] Return something like order total summary so we can have fancy formatters
* [x] Have an `addProcessor` method, inject Promotion/Tax processor in compiler pass


edurenye:
The formatter options are missing in the list of task to do, this is just a reminder.
Good work by now, I'm using your pull request as a composer patch. Thank you for it!

bojanz:
The issue has newer code.

PR #782, opened 2017-09-02
2906360: Split Offsite interface onNotify method to a new interface

jackbravo:
This is from https://www.drupal.org/node/2906360

We are developing a payment gateway in Bolivia that uses cash payments in stores (https://www.drupal.org/project/commerce_pagos_net). The plugin is implementing both the OnsitePaymentInterface (because during checkout we call a web service to send the code that the user will have to present at the physical store to pay) and also implements the OffistePaymentInterface, because the payments provider sends a JSON back to our site once the user pays in any store. It works right now because in the PaymentProcess class, the `if` checks before if we are implementing the OnsitePaymentInterface, but if the if order changed then it would no longer work. I think it would be better to separate the onNotify method of the OffsitePaymentInterface to another interface so that any payment gateway can implement the onNotify method and use the OffsitePaymentController.

### DIFF
--- a/modules/payment/src/Plugin/Commerce/PaymentGateway/OffsitePaymentNotificationInterface.php
+++ b/modules/payment/src/Plugin/Commerce/PaymentGateway/OffsitePaymentNotificationInterface.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Drupal\commerce_payment\Plugin\Commerce\PaymentGateway;
+
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Defines the interface for gateways which receive payment notifications.
+ */
+interface OffsitePaymentNotificationInterface {
+
+  /**
+   * Processes the "notify" request.
+   *
+   * Note:
+   * This method can't throw exceptions on failure because some payment
+   * providers expect an error response to be returned in that case.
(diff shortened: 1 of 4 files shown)


modules/payment/src/Plugin/Commerce/PaymentGateway/OffsitePaymentNotificationInterface.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Drupal\commerce_payment\Plugin\Commerce\PaymentGateway;
+
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Defines the interface for gateways which receive payment notifications.
+ */
+interface OffsitePaymentNotificationInterface {
```
bojanz: Let's rename this to SupportsNotificationsInterface, that would fit with the others (refunds, authorizations, etc)

jackbravo: Sounds good. Done =)

bojanz:
Committed in 755924ef869f4f721ec6e28986607d6b773c3716. Thanks!

PR #784, opened 2017-09-03
Replace the cart block view with a Twig template

thomascys:
PR for https://www.drupal.org/node/2906024

### DIFF
--- a/modules/cart/commerce_cart.module
+++ b/modules/cart/commerce_cart.module
@@ -36,6 +36,11 @@ function commerce_cart_theme($existing, $type, $theme, $path) {
         'links' => [],
       ],
     ],
+    'commerce_cart_order_items_table' => [
+      'variables' => [
+        'order_entity' => NULL,
+      ],
+    ],
     'commerce_cart_empty_page' => [
       'render element' => 'element',
     ],
--- a/modules/cart/commerce_cart.module
+++ b/modules/cart/commerce_cart.module
@@ -31,6 +31,7 @@ function commerce_cart_theme($existing, $type, $theme, $path) {
         'count' => NULL,
         'count_text' => '',
         'content' => NULL,
+        'cart_items' => NULL,
         'url' => NULL,
         'links' => [],
       ],
--- a/modules/cart/commerce_cart.module
+++ b/modules/cart/commerce_cart.module
@@ -211,11 +211,22 @@ function commerce_cart_form_commerce_order_type_form_alter(array &$form, FormSta
     '#options' => $available_form_views,
     '#default_value' => $cart_form_view,
   ];
+  $form['commerce_cart']['cart_block_use_view'] = [
+    '#type' => 'checkbox',
+    '#title' => t('Use a View to display the cart block contents'),
+    '#description' => t('Overrides the cart block contents with the output of a View.'),
+    '#default_value' => !empty($cart_block_view),
+  ];
   $form['commerce_cart']['cart_block_view'] = [
     '#type' => 'select',
     '#title' => t('Shopping cart block view'),
     '#options' => $available_block_views,
     '#default_value' => $cart_block_view,
+    '#states' => [
+      'visible' => [
+        ':input[name="commerce_cart[cart_block_use_view]"]' => ['checked' => TRUE],
+      ],
+    ],
   ];
 
   $form['actions']['submit']['#submit'][] = 'commerce_cart_order_type_form_submit';
@@ -257,7 +268,10 @@ function commerce_cart_order_type_form_submit($form, FormStateInterface $form_st
   $order_type = $form_state->getFormObject()->getEntity();
   $settings = &$form_state->getValue('commerce_cart');
   $order_type->setThirdPartySetting('commerce_cart', 'cart_form_view', $settings['cart_form_view']);
-  $order_type->setThirdPartySetting('commerce_cart', 'cart_block_view', $settings['cart_block_view']);
+  $order_type->setThirdPartySetting('commerce_cart', 'cart_block_view', FALSE);
+  if ($settings['cart_block_use_view']) {
+    $order_type->setThirdPartySetting('commerce_cart', 'cart_block_view', $settings['cart_block_view']);
+  }
   $order_type->save();
 }
 
--- a/modules/cart/commerce_cart.module
+++ b/modules/cart/commerce_cart.module
@@ -31,7 +31,6 @@ function commerce_cart_theme($existing, $type, $theme, $path) {
         'count' => NULL,
         'count_text' => '',
         'content' => NULL,
-        'cart_items' => NULL,
         'url' => NULL,
         'links' => [],
       ],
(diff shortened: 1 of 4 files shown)


modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -102,6 +131,8 @@ public function blockForm($form, FormStateInterface $form_state) {
    */
   public function blockSubmit($form, FormStateInterface $form_state) {
     $this->configuration['dropdown'] = $form_state->getValue('commerce_cart_dropdown');
+    $this->configuration['use_view'] = $form_state->getValue('use_view');
```
bojanz: We should not be storing use_view in configuration. Follow the OrderSummary checkout pane example in how it deals with this (saving view as NULL when the checkbox is unchecked)

modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -157,6 +194,7 @@ public function build() {
       '#count_text' => $this->formatPlural($count, '@count item', '@count items'),
       '#url' => Url::fromRoute('commerce_cart.page')->toString(),
       '#content' => $cart_views,
+      '#cart_items' => $cart_items,
```
bojanz: This doesn't match the approach we took for the order summary twig template. Makes more sense to pass the full #order_entity.

thomascys: Thanks for the feedback. I'll need to investigate how this should work out.

thomascys: If i need to pass the full #order_entity then how will this work out when you have multiple order types?

mglaman: If we're using Twig, shouldn't we just skip passing View objects to the template?

One issue we have here is that we have one table for all items. We need to keep `cart_items`, but perhaps it contains an array of order items, keyed by the order ID.

mglaman: EDIT: I see how we fix this. Right now the existing cart block is being harnessed. We need a new theme hook. Maybe `commerce_cart_order_items_table`.

Which has

```
+        <table>
 +          <tbody>
 +          {% for cart_item in cart_items %}
 +            <tr>
 +              <td>{{ cart_item.getQuantity|number_format }} x</td>
 +              <td>{{ cart_item.label }}</td>
 +              <td>{{ cart_item.getTotalPrice|commerce_price_format }}</td>
 +            </tr>
 +          {% endfor %}
 +          </tbody>
 +        </table>
```

Then the `#content` value is an array of those.  The new hook received `#order_entity`

mglaman (changes requested):
The way forward, here, is to create `commerce_cart_order_items_table` theme hook which takes an `#order_entity` object. The block's `#content` property of the render array is an array of `commerce_cart_order_items_table` render arrays.

mglaman:
@thomascys let me know of any clarifications. We really just need the new theme hook. Feel free to ping me in Drupal Slack or IRC.

thomascys:
Thanks Matt for the feedback! I committed a first attempt for the suggested `commerce_cart_order_items_table` theme hook.

mglaman (approved):

mglaman:
Off hand that looks great! Awesome and thanks @thomascys. I'll look into the test fail and manually test later.

modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -125,9 +158,14 @@ public function build() {
     });
 
     $count = 0;
-    $cart_views = [];
+    $content = [];
     if (!empty($carts)) {
-      $cart_views = $this->getCartViews($carts);
+      if ($this->configuration['view']) {
```
mglaman: This is causing the error, because no `!empty()` check, which also means it needs a default value of `NULL`.

mglaman (changes requested):
So I just realized something.

Are we respecting `cart_block_view` from the commerce_cart's third_party_settings? I suppose we shouldn't be adding settings here, but reading settings from the order type. Similar concept, just need to move form stuff to `commerce_cart_form_commerce_order_type_form_alter`

thomascys:
Thanks Matt for pointing me in the right direction. 

I moved the configuration to `commerce_cart_form_commerce_order_type_form_alter`.

I also moved some things around in the build, getCartViews & getCartTables methods since now you can have a different configuration for every order type.
For example: 'default' order type can be rendered as a view and 'digital' order type can be rendered by Twig.

modules/cart/commerce_cart.module
```diff
@@ -30,10 +30,16 @@ function commerce_cart_theme($existing, $type, $theme, $path) {
         'count' => NULL,
         'count_text' => '',
         'content' => NULL,
+        'cart_items' => NULL,
```
mglaman: We can remove this line, correct? Since we now have the new theme hook.

modules/cart/src/Plugin/Block/CartBlock.php
```diff
@@ -125,10 +126,21 @@ public function build() {
...
     $count = 0;
-    $cart_views = [];
+    $content = [];
     if (!empty($carts)) {
-      $cart_views = $this->getCartViews($carts);
+      $order_type_storage = $this->entityTypeManager->getStorage('commerce_order_type');
       foreach ($carts as $cart_id => $cart) {
+        if ($this->configuration['dropdown']) {
+          $order_type = $cart->bundle();
+          $order_type_entity = $order_type_storage->load($order_type);
```
mglaman: We should follow the `$order_types = $order_type_storage->loadMultiple(array_unique($order_type_ids));` pattern. `loadMultiple` is really that much more performant.

modules/cart/templates/commerce-cart-block.html.twig
```diff
@@ -5,11 +5,11 @@
       <span class="cart-block--summary__count">{{ count_text }}</span>
     </a>
   </div>
-  {% if content %}
+  {% if (content) %}
```
mglaman: In twig you don't need to wrap with `(` and `)`

modules/cart/templates/commerce-cart-block.html.twig
```diff
@@ -5,11 +5,11 @@
...
   <div class="cart-block--contents">
     <div class="cart-block--contents__inner">
       <div class="cart-block--contents__items">
-        {{ content }}
+          {{ content }}
```
mglaman: Extra indent

mglaman (changes requested):
Few more nits! Looking good

mglaman:
Also: yay tests are passing! I just realized, however, we'll need to add tests for changing this setting :( We failed to do so for the checkout summary and a bug snuck in. I can help write that if you'd like.

thomascys:
Did some refactoring based on your feedback. I would like to leave the testing part for you since i really need to up my game on writing good tests.

mglaman (approved):

bojanz:
Let's continue on the issue: https://www.drupal.org/project/commerce/issues/2906024

Thanks for the work so far, Thomas!

PR #785, opened 2017-09-04
Be able to choose a checkout step to place the order

xuxucode:
PR for https://www.drupal.org/node/2906563


xuxucode:
Placing an order on payment step does not make sense.

PR #787, opened 2017-09-08
Issue #2905147 by mglaman: Commerce Log: Call to a member function la…

mglaman:
…bel() on null


bojanz:
Committed in 750759579fcd62eda5137624acf4154b405701e1.

PR #789, opened 2017-09-11
2885421 promotion description text format

finnef:
See https://www.drupal.org/node/2885421
Altered commerce_promotion entity description base field to allow for text formats. Added an update hook to alter existing data. Used low level DB API because Entity API is not aware yet of the changed base field storage definition when you update.


finnef:
Hmm the error in PHP 5.5 is 
Drupal\Tests\commerce_promotion\FunctionalJavascript\CouponRedemptionPaneTest::testCheckoutWithMainSubmit
    Behat\Mink\Exception\ResponseTextException: The text &quot;Visa ending in
    1111&quot; was not found anywhere in the text of the current page.

I don't see how this is related to the description field of the commerce_promotion entity, which is not used in this test at all.

Ah. Travis resource shortage was the cause. Rerun worked.

PR #790, opened 2017-09-12
2869209 promotion compatibility

finnef:
This patch makes the suggested promotion compatibility functionality in the current GUI (compatible with all or none) work.
See https://www.drupal.org/node/2869209


bojanz:
Work has continued on the issue https://www.drupal.org/node/2869209

PR #786, opened 2017-09-05
Issue #2804227 Add $order->getBalance()

jackbravo:
This is a rebase of [PR 508](https://github.com/drupalcommerce/commerce/pull/508). Adding a commerce_order_post_update_6() function.

PR #791, opened 2017-09-12
2862339 - Add event listener to payments

jackbravo:
This is a reroll of PR [684](https://github.com/drupalcommerce/commerce/pull/684) with some changes. I added the 'event' annotation to the Payment entity, so now all the events should work. There are tests right now for the create and load event and both pass (load event was not working). I also changed the getOrder method to getEntity, and added a test for that (as mentioned in comment [#6](https://www.drupal.org/node/2862339#comment-12218372). This issue is related to https://www.drupal.org/node/2856586 so I'm planning on adding the order_paid_in_full event too but maybe in another PR that depends on this one.


modules/payment/src/PaymentStorage.php
```diff
@@ -53,7 +55,13 @@ protected function doCreate(array $values) {
       $values['type'] = $payment_type->getPluginId();
     }
 
-    return parent::doCreate($values);
+    $payment = parent::doCreate($values);
+
+    // Notify other modules.
+    $event = new PaymentEvent($payment);
```
haringsrob: I think we can move these 2 lines into a new private method: notifyOfPaymentCreation() (or suggest another name that fits better)

That way the comment is no longer needed and it will be easier to follow the code.

haringsrob (commented):
Got a small comment that could improve code quality

modules/payment/src/PaymentStorage.php
jackbravo: Could be. I'm also trying to tie this to issue [2856586-add-paid-in-full-event](https://www.drupal.org/node/2856586), which depends on [2804227-add-order-getBalance](https://www.drupal.org/node/2804227) ([PR 786](https://github.com/drupalcommerce/commerce/pull/786)).

PR 786 sets the payment balance on Payment::preSave() method, which gets the $storage interface as an argument, so if we can notify general events from the EntityStorageInterface we can solve that issue too. Right now, inside the Payment object, we don't have access to the eventDispatcher...

heddn (changes requested):
There's already events available from state_machine. See https://docs.drupalcommerce.org/commerce2/developer-guide/orders/react-to-workflow-transitions.

Do we need this at all?

Examples:
'commerce_payment.authorize_capture.post_transition'
'commerce_payment.receive.post_transition'
'commerce_payment.partially_refund.post_transition'

jackbravo:
Hmmmm, this is indeed a very good question @heddn =p. I was planning on using this code as foundation to implement the 'order is paid in full' event (issue [2856586](https://www.drupal.org/node/2856586)).

Maybe we can skip that and people should just subscribe to those events.

But then again, that event sounds really useful, but maybe we don't need to implement these CRUD events first, and we can just add the 'order paid in full' without all the others.

PR #794, opened 2017-09-14
Issue #2907367 by mglaman: Add indexes to important fields

mglaman:
Need to figure out the update hook.

Looks like we need to run `\Drupal\Core\Entity\Sql\SqlContentEntityStorageSchema::onEntityTypeUpdate`. This is called from `\Drupal\Core\Entity\EntityDefinitionUpdateManager::doEntityUpdate`, the entity definition updater. That invokes `\Drupal\Core\Entity\EntityTypeListenerInterface::onEntityTypeUpdate`. 

Actually, we can get the entity type's storage and run: `\Drupal\Core\Entity\Sql\SqlContentEntityStorage::onEntityTypeUpdate` implemented in the storage.

```
  /**
   * {@inheritdoc}
   */
  public function onEntityTypeUpdate(EntityTypeInterface $entity_type, EntityTypeInterface $original) {
    // Ensure we have an updated entity type definition.
    $this->entityType = $entity_type;
    // The table layout may have changed depending on the new entity type
    // definition.
    $this->initTableLayout();
    // Let the schema handler adapt to possible table layout changes.
    $this->wrapSchemaException(function () use ($entity_type, $original) {
      $this->getStorageSchema()->onEntityTypeUpdate($entity_type, $original);
    });
  }
```

So in commerce.post_update.php we should add a hook which runs this for all entities implementing our handler (just our entities.)

### DIFF
--- a/commerce.install
+++ b/commerce.install
@@ -5,6 +5,8 @@
  * Contains install and update functions for Commerce.
  */
 
+use Drupal\commerce\CommerceContentEntityStorageSchema;
+
 /**
  * Implements hook_requirements().
  */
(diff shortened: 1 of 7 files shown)


mglaman:
Ran upgrade. Didn't see changes.

mglaman:
Fixed. Upgrade didn't work because I forgot `getHandlerClass` returned a string. Properly checking class and running updates. Now have indexes!

bojanz:
Tested, didn't work. Matt thinks we need to get the uncached definitions.

PR #795, opened 2017-09-14
patch for addressbook reuse

nikunj-addweb:


bojanz:
Hi, you appear to have rolled this against an earlier Commerce version, so it's reverting many of the commits that are in -dev. Our current addressbook PR is in #760, please focus on reviewing that. Thanks!

PR #797, opened 2017-09-14
Add events and logging for payments

heddn:
This takes pieces from #610 and #791 and combines them.

### DIFF
--- a/modules/log/commerce_log.commerce_log_categories.yml
+++ b/modules/log/commerce_log.commerce_log_categories.yml
@@ -5,3 +5,7 @@ commerce_cart:
 commerce_order:
   label: Order
   entity_type: commerce_order
+
+commerce_payment:
+  label: Payment
+  entity_type: commerce_payment
--- a/modules/log/commerce_log.commerce_log_templates.yml
+++ b/modules/log/commerce_log.commerce_log_templates.yml
@@ -28,15 +28,15 @@ order_assigned:
   label: 'Order assigned'
   template: '<p>The order was assigned to {{ user }}.</p>'
 
-payment_insert:
+payment_inserted:
   category: commerce_payment
   label: 'Payment created'
   template: '<p>Payment of {{ amount|commerce_price_format }} was created.</p>'
-payment_update:
+payment_price_updated:
   category: commerce_payment
-  label: 'Payment updated'
-  template: '<p>Payment of {{ amount|commerce_price_format }} was updated.</p>'
-payment_delete:
+  label: 'Payment price updated'
+  template: '<p>Payment was changed from {{ previous_amount|commerce_price_format }} to {{ current_amount|commerce_price_format }}.</p>'
+payment_deleted:
   category: commerce_payment
   label: 'Payment deleted'
   template: '<p>Payment of {{ amount|commerce_price_format }} was deleted.</p>'
--- a/modules/log/commerce_log.commerce_log_templates.yml
+++ b/modules/log/commerce_log.commerce_log_templates.yml
@@ -28,19 +28,11 @@ order_assigned:
   label: 'Order assigned'
   template: '<p>The order was assigned to {{ user }}.</p>'
 
-payment_inserted:
+payment_log:
   category: commerce_payment
-  label: 'Payment created'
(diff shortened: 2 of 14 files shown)


modules/log/commerce_log.module
```diff
@@ -20,3 +22,8 @@ function commerce_log_preprocess_commerce_order(&$variables) {
     '#title' => t('Order activity'),
   ];
 }
+
+function commerce_log_entity_type_alter(array &$entity_types) {
+  /** @var $entity_types \Drupal\Core\Entity\EntityTypeInterface[] */
+  $entity_types['commerce_payment']->setListBuilderClass(PaymentListBuilder::class);
```
heddn: I don't care if the module is enabled, if it isn't this will just get ignored.

modules/log/config/install/views.view.commerce_activity.yml
```diff
@@ -337,7 +337,7 @@ display:
             type: none
             fail: 'not found'
           validate_options: {  }
-          break_phrase: false
+          break_phrase: true
```
heddn: This let's us pass in multiple entity id arguments to the source id contextual argument.

modules/log/commerce_log.module
bojanz: If the module is disabled there won't be an $entity_types['commerce_payment'] object, causing a crash, no?

modules/log/src/PaymentListBuilder.php
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function render() {
+    $build = parent::render();
+    $entityIds = [];
+    foreach ($this->load() as $entity) {
+      $entityIds[] = $entity->id();
+    }
+    $build['log']['title'] = [
```
heddn: Maybe some views expert will figure out how to print the title. It didn't want to. So I punted and did this hack.

modules/log/commerce_log.commerce_log_templates.yml
```diff
@@ -27,3 +27,20 @@ order_assigned:
...
+  template: '<p>Payment of {{ amount|commerce_price_format }} was created.</p>'
+payment_update:
+  category: commerce_payment
+  label: 'Payment updated'
+  template: '<p>Payment of {{ amount|commerce_price_format }} was updated.</p>'
+payment_delete:
+  category: commerce_payment
+  label: 'Payment deleted'
+  template: '<p>Payment of {{ amount|commerce_price_format }} was deleted.</p>'
+payment_refunded:
```
heddn, line 35: I figured that insert, update, delete and refund are the only things of interest.

heddn: Sounds like we want more details in the messages. Let's add the current state as part of the log too.

modules/log/commerce_log.module
heddn: Ah, that's true. I ignored that important item. I was still in d7 array mode.

modules/log/commerce_log.commerce_log_categories.yml
```diff
@@ -5,3 +5,7 @@ commerce_cart:
 commerce_order:
   label: Order
   entity_type: commerce_order
+
+commerce_payment:
+  label: Payment
+  entity_type: commerce_payment
```
heddn, line 11: Should this read commerce_order?

modules/payment/src/Event/PaymentEvents.php
```diff
@@ -10,7 +10,74 @@
...
+  const PAYMENT_PRESAVE = 'commerce_payment.commerce_payment.presave';
+
+  /**
+   * Name of the event fired after saving a new payment.
+   *
+   * @Event
+   *
+   * @see \Drupal\commerce_payment\Event\PaymentEvent
+   */
+  const PAYMENT_INSERT = 'commerce_payment.commerce_payment.insert';
```
heddn, line 48: I see an INSERT and a CREATE in product variation events. Do we need to duplicate the same here? The docs between the two aren't very clear about what the difference is.

PR #799, opened 2017-09-15
Add theme suggestions for the checkout order summary and completion message templates

mglaman:
…er summary and completion message templates


mglaman:
@bojanz the checkout message currently only receives the order. Should we also pass it the checkout_flow entity as well, then, so we can provide suggestion?

mglaman:
I'm stupid.

```
  /** @var \Drupal\commerce_checkout\Entity\CheckoutFlowInterface $checkout_flow */
  $checkout_flow = $variables['order_entity']->get('checkout_flow')->entity;
```

Updated.

PR #801, opened 2017-09-19
Initial product variation logging PoC

heddn:


heddn:
It looks like the following. I'd like feedback on the text of the wording and placement of where to put the UI components of the log. Right now I have it placed on the product edit page.

screenshot_2017-09-19_15-26-28

heddn:
For the field diffs, I'm using a json serialized string of the original value compared to the current value. Where value is retrieved from the field using `getValue()`

heddn (commented):
The tests were failing because we now have more events and the hard coded ids in the tests were off. I'm using array_pop now instead.

PR #804, opened 2017-09-22
Issue #2856586: order paid in full event

jackbravo:
This fixes issue #2856586: order paid in full event. It depends on [PR 786](https://github.com/drupalcommerce/commerce/pull/786): add $order->getBalance().


jackbravo:
This latest push failed only on PHP 5.5 and in a javascript test, so I would think that the patch is clean.

bojanz:
Continued on the issue.

PR #805, opened 2017-09-22
Issue #2910937 by rgpublic, facine: Checkout shows payment methods which cannot be reused

facine:


bojanz:
The parent issue was closed and fixed elsewhere.

PR #807, opened 2017-09-29
Issue 2904629: fix CartOrderPlacedTest uses two different users for c…

p4trizio:
related to issue https://www.drupal.org/node/2904629


mglaman:
@bojanz did a review. Looks good. Our test was broken. This fixes it. Luckily tests still pass. 😥

PR #809, opened 2017-09-30
Issue #2912996 - Anonymous order payments

bradjones1:
See https://www.drupal.org/node/2912996

### DIFF
--- a/modules/payment/src/Form/PaymentAddForm.php
+++ b/modules/payment/src/Form/PaymentAddForm.php
@@ -3,6 +3,7 @@
 namespace Drupal\commerce_payment\Form;
 
 use Drupal\commerce\EntityHelper;
+use Drupal\commerce_payment\Entity\Payment;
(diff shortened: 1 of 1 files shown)


bradjones1:
Ah, coding standards. Will fix.

bradjones1:
@mglaman Any thoughts on why this fails only on PHP7, and on coupon redemption?

bojanz:
@bradjones1
We have a random failure in the coupon redemption test that we're still chasing. I think it's not always visible cause your Travis has different limits than the drupalcommerce/commerce travis.

bradjones1:
@bojanz Ok thanks. Any other thoughts on this PR though?

bradjones1:
Further work at https://github.com/bradjones1/commerce/pull/1

bojanz:
Let's continue on the issue (https://www.drupal.org/node/2912996). I commented about the code needing a reroll for the new PaymentOptions API.

PR #810, opened 2017-10-01
Issue 2897698: fix with a view catalog, only the first product variat…

p4trizio:
Issue https://www.drupal.org/node/2897698
The patch provided seems to solve a problem with ajax add to cart form

### DIFF
--- a/modules/product/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
+++ b/modules/product/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
@@ -113,7 +113,9 @@ class ProductVariationTitleWidget extends ProductVariationWidgetBase implements 
     }
 
     // Set the selected variation in the form state for our AJAX callback.
-    $form_state->set('selected_variation', $selected_variation->id());
+    if($selected_variation) {
+      $form_state->set('selected_variation', $selected_variation->id());
+    }
 
     $variation_options = [];
     foreach ($variations as $option) {
@@ -124,7 +126,8 @@ class ProductVariationTitleWidget extends ProductVariationWidgetBase implements 
       '#title' => $this->getSetting('label_text'),
       '#options' => $variation_options,
       '#required' => TRUE,
-      '#default_value' => $selected_variation->id(),
+      '#default_value' => $selected_variation ? $selected_variation->id() : NULL,
+      '#attributes' => ['data-disable-refocus' => 'true'],
       '#ajax' => [
         'callback' => [get_class($this), 'ajaxRefresh'],
         'wrapper' => $form['#wrapper_id'],
@@ -153,6 +156,9 @@ class ProductVariationTitleWidget extends ProductVariationWidgetBase implements 
    */
   protected function selectVariationFromUserInput(array $variations, array $user_input) {
     $current_variation = NULL;
+    if (!isset($variations[$user_input['variation']])) {
+      $variations[$user_input['variation']] = [];
+    }
     if (!empty($user_input['variation']) && $variations[$user_input['variation']]) {
       $current_variation = $variations[$user_input['variation']];
     }


mglaman:
Failed due to


```
The command "phpcs --standard=phpcs.xml src -s" exited with 0.
7.28s$ phpcs --standard=phpcs.xml modules -s
FILE: ...uct/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
 116 | ERROR | [x] Expected 1 space after IF keyword; 0 found
     |       |     (Drupal.ControlStructures.ControlSignature.SpaceAfterKeyword)
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
Time: 7.2 secs; Memory: 46Mb
```

modules/product/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
```diff
@@ -153,6 +156,9 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
    */
   protected function selectVariationFromUserInput(array $variations, array $user_input) {
     $current_variation = NULL;
+    if (!isset($variations[$user_input['variation']])) {
```
mglaman, line 159: The `!empty($user_input['variation'])` check below should cover this `isset`. `empty` handles not set as well.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
```diff
@@ -124,7 +126,8 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
       '#title' => $this->getSetting('label_text'),
       '#options' => $variation_options,
       '#required' => TRUE,
-      '#default_value' => $selected_variation->id(),
+      '#default_value' => $selected_variation ? $selected_variation->id() : NULL,
+      '#attributes' => ['data-disable-refocus' => 'true'],
```
mglaman, line 130: > That is, when you choose one of the options in the second, or third ,,,etc, select input, the focus jumps to the first select input.

This only fixes this issue for title widget, not attributes. Or is issue here?

mglaman (changes requested):
phpcs fixes. general questions.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
drugan: The issue of "jumping" also exists (not fixed yet) on the ProductVariationAttributesWidget. It could be fixed with rejecting refocusing using HTML attribute as the above example or explicitly assigning persistent ID for a widget:

``` 
// .....
$element['variation'] = [
  '#type' => 'select',
  '#id' => "edit-purchased-entity-0-product-{$product->id()}-variation-{$selected_variation->id()}",
  '#title' => $this->getSetting('label_text'),
  '#options' => $variation_options,
  '#required' => TRUE,
  '#default_value' => $selected_variation->id(),
  '#ajax' => [
    'callback' => [get_class($this), 'ajaxRefresh'],
    'wrapper' => $form['#wrapper_id'],
  ],
];
// ....
```

modules/product/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
drugan: May be instead of all the above checks for empty user input (and therefore empty selected variation) we just replicate logic from here:
https://github.com/drupalcommerce/commerce/blob/8.x-2.x/modules/product/src/Plugin/Field/FieldWidget/ProductVariationAttributesWidget.php#L124

```
if ($form_state->isRebuilding()) {
  $parents = array_merge($element['#field_parents'], [$items->getName(), $delta]);
  $user_input = (array) NestedArray::getValue($form_state->getUserInput(), $parents);
  $selected_variation = $this->selectVariationFromUserInput($variations, $user_input);
}
```

So, if the form is rebuilding, which means it is not initial page loading, then take user input which cannot be empty at this moment as the rebuilding itself is initiated by a user input. For me it works flawlessly.

modules/product/src/Plugin/Field/FieldWidget/ProductVariationTitleWidget.php
kaido24: As @drugan told this jumping can be fixed by adding unique id for variation form.
Just mixed his code from his project with this problem here and got working code.
There where issues with jumping and not selecting the variations as planned.

    $base_id = $form_state->getFormObject()->getBaseFormId();
    $id = strtr($form_state->getFormObject()->getFormId(), [$base_id => '']);
    foreach ($this->getAttributeInfo($selected_variation, $variations) as $field_name => $attribute) {
      $element['attributes'][$field_name] = [
        '#id' => 'edit-purchased-entity-0-attributes-' . Html::getClass($field_name . $id),
        '#type' => $attribute['element_type'],
        '#title' => $attribute['title'],
        '#options' => $attribute['values'],
        '#required' => $attribute['required'],
        '#default_value' => $selected_variation->getAttributeValueId($field_name),
        '#limit_validation_errors' => [],
        '#ajax' => [
          'callback' => [get_class($this), 'ajaxRefresh'],
          'wrapper' => $form['#wrapper_id'],
        ],
      ];

kaido24: [fix_jumping.patch.txt](https://github.com/drupalcommerce/commerce/files/1775935/fix_jumping.patch.txt)

bojanz:
The issue now has an updated patch, let's continue there (http://drupal.org/node/2897698)

PR #811, opened 2017-10-10
Issue #2915274 by mglaman: Add PanelizerEntity plugin to support Pane…

mglaman:
…lizer integration for products


modules/product/src/Plugin/PanelizerEntity/PanelizerProduct.php
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * @param array $configuration
```
bojanz: Constructs a new PanelizerProduct object.

PR #813, opened 2017-10-13
Issue #2915942 by steveoliver: Update the payment information pane if it exists with coupon pane

steveoliver:
See https://www.drupal.org/node/2915942.


bojanz:
This is being fixed generically in https://www.drupal.org/project/commerce/issues/2710999

PR #814, opened 2017-10-13
In the OrderStore condition plugin, use the correct storage for stores

erikerskine:
When creating a condition based on stores, the wrong UUIDs are saved. They are those of the product rather than the store.


bojanz:
Fail :( Thanks for catching this!

If you open a d.o issue, I can credit you there as well.

drugan:
I wonder how the whole system worked without this so far?!

... and all the tests are keeping silence in green.

@erikerskine Great catch!

erikerskine:
@bojanz see https://www.drupal.org/node/2916099

bojanz:
Unfortunately our condition tests don't cover their config forms, just the evaluate() part.

Committed in d2c9dce103c42de55675e1b6e7b5185c9e86c466.

PR #815, opened 2017-10-17
New test AjaxLoadedCartFormText.php.

chancegrissom:
Products loaded via views ajax-enabled pagination currently have broken Add to Cart forms. 

Attempting to modify the ajax-loaded products attributes will cause /views/ajax to throw a 404. 

Attempting to add the ajax-loaded product to ones cart (even without modifying the attributes) redirects to /views/ajax, which renders the 404 page.

This is only the failing test case. I tried to keep it as minimal as possible.


mglaman:
~~Let's make sure to open an Drupal.org issue~~

Didn't see https://www.drupal.org/node/2916671

bojanz:
Let's continue on the issue.

PR #816, opened 2017-10-19
Feature/usage limit multiple

gmorice:
Hi, I work on feature asked by https://www.drupal.org/node/2902495.
PR add feature to Promotion usage limit. We can define global usage limit and usage limit per user.
On click on radio button to enable usage limit, usage limit per user field is displayed below the global usage limit field. Both usage limits are checked to limit the promotion use.


bojanz:
Let's continue on the issue. Lisa Streeter will be rerolling the patch and doing further work: https://www.drupal.org/project/commerce/issues/2902495

PR #818, opened 2017-10-23
Issue #2898217 by tilenav: Delete payment methods when their payment gateway is deleted

tilenav:
I hope this is the kind of queue you were expecting. Keep in mind that until the payment methods are deleted (cron run), users will receive an error on /user/{user}/payment-methods.


tilenav:
The last build went through on PHP 5.5 and 7, failed on 5.6 due to 

> 0.02s$ phpenv global 5.6
>rbenv: version `5.6' not installed
>The command "phpenv global 5.6" failed and exited with 1 during .

PR #819, opened 2017-10-26
Issue #2898218 by tilenav: create soft deletes for payment methods

tilenav:
Current solution only for payment methods, could maybe do a broader solution for all/most commerce entities if necessary.


tilenav:
In progress: Needs to be hidden from display on payment information

mglaman:
Closing. Moved to D.o https://www.drupal.org/project/commerce/issues/2898218#comment-12529185

PR #820, opened 2017-10-26
Issue 2919078 by nikathone: Displaying wrong coupon usage number

nikathone:
Not sure if this is a bug with promotion but under `Drupal\commerce_promotion\CouponListBuilder::load()` line 76 we have `$this->usageCounts = $this->usage->loadMultiple($coupons);` shouldn't this be `$this->usageCounts = $this->usage->loadMultipleByCoupon($coupons);`


bojanz:
Committed in 06069b60284a82630af65ab45df4abcbfb76da1d. Thanks!

PR #821, opened 2017-10-30
WIP / test only -- 2804227 get order balance jack

steveoliver:
This is a rebase of @jackbravo's branch of the same name.  Let's see what Travis has to say.


bojanz:
Continued on the issue https://www.drupal.org/node/2804227

PR #828, opened 2017-11-12
2922811 - Deprecate Commerce's BundleFieldDefinition, BundlePluginInterface

bradjones1:
https://www.drupal.org/project/issues/commerce/2922811

### DIFF
--- a/src/BundleFieldDefinition.php
+++ b/src/BundleFieldDefinition.php
@@ -2,11 +2,15 @@
 
 namespace Drupal\commerce;
 
+@trigger_error('The ' . __NAMESPACE__ . '\BundleFieldDefinition is deprecated. Instead, use \Drupal\entity\BundleFieldDefinition', E_USER_DEPRECATED);
+
 use Drupal\entity\BundleFieldDefinition as BaseBundleFieldDefinition;
 
 /**
  * Provides a field definition class for bundle fields.
  *
  * Note: This code has moved to Entity API, see the parent class.
+ *
+ * @deprecated in Commerce 2.0
  */
 class BundleFieldDefinition extends BaseBundleFieldDefinition {}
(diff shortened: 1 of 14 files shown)


src/BundleFieldDefinition.php
```diff
@@ -2,11 +2,15 @@
 
 namespace Drupal\commerce;
 
+@trigger_error('The '. __NAMESPACE__ . '\BundleFieldDefinition is deprecated. Instead, use \Drupal\entity\BundleFieldDefinition', E_USER_DEPRECATED);
```
bojanz: ----------------------------------------------------------------------
 5 | ERROR | [x] Concat operator must be surrounded by a single space
   |       |     (Squiz.Strings.ConcatenationSpacing.PaddingFound)

bradjones1: Fixed. Also figured we should do the same with `BundlePluginInterface`.

bradjones1:
Rebased.

bojanz:
Unfortunately, this will immediately crash all of contrib that uses bundle plugins (recurring, license, gateways with custom payment method types), because ConfigurableFieldManager gets an unexpected class (\Drupal\commerce\BundleFieldDefinition and \Drupal\entity\BundleFieldDefinition are not interchangeable). And if we keep the old typehint in ConfigurableFieldManager, then there's no way to move off the deprecated classes.

So it looks like deprecating BundlePluginInterface is actually the only part we can do.

EDIT: Nevermind, looks like I'm wrong, the Commerce BundleFieldDefinition extends the Entity one. The crash I observed was without this PR, trying to change the commerce_recurring usage before Commerce is fixed.

bojanz:
Committed in 654bf3db4c1a879193ab4437e0865d2b69c74ccf.

PR #831, opened 2017-11-13
#2922381 - Unable to export the 'online' store entity form display configuration

robincee:
See https://www.drupal.org/node/2922381


robincee:
PHP 5.6 trips over a payment test? (rest is green). That's not something we change in this pr?

PR #835, opened 2017-11-23
Readme update

pazhyn:
Drupal Commerce 2.x is already stable. Let's update README.md accordingly.


pazhyn:
@bojanz can you make a quick review of this tiny update? It is the face of the project for developers and current README.md it can mislead them. Thanks.

pazhyn:
@bojanz, what do you think about adding Drupal Commerce logo into README.md? 
Repository homepage looks pretty nice with it - https://github.com/anyforsoft/commerce/tree/readme

drugan:
@pazhyn seems that you are the same as me and giving to *README* files particular attention :+1: 

The only notice is that you should avoid using HTML tags in the *README* files despite the *Github* "understands" the markup and successfully renders it. Remember these files in the first place are intended for reading in a text editor or using `cat` command before software is installed and could render anything at all. So, the HTML tags just pollute a file making it less readable.

Instead, I'd recommend to use the markdown (see **aA B i <>** controls above a comment textarea) for inserting the image into the file:

\!\[Drupal Commerce\]\(https://drupalcommerce.org/sites/default/files/dclogo_stacked_2c_on_white.png "On Hover Title")

`...will be converted into:`
Drupal Commerce


BTW, there is a *Drupal* module which renders *README* file on the **admin/help/my_module** page the same way as the *Github* does it:

https://www.drupal.org/project/readmehelp

pazhyn:
@drugan thank you for your suggestion! you are totally right, this markup is much better.
please review it.

p.s.: thank you for your module readmehelp!

README.md
```diff
@@ -1,13 +1,13 @@
...
 ===============
 [![Build Status](https://travis-ci.org/drupalcommerce/commerce.svg?branch=8.x-2.x)](https://travis-ci.org/drupalcommerce/commerce)
 
 Drupal Commerce is the leading flexible eCommerce solution for Drupal,
 powering over 60,000 online stores of all sizes.
 
-Note: 8.x-2.x is in beta. It is already usable for production websites 
-and upgrades between versions are supported. Bugs are still expected. 
-Please report bugs in the [issue queue](https://www.drupal.org/project/issues/commerce?version=8.x).
+Drupal Commerce 8.x-2.x is a stable release. It is usable for production websites.
```
drugan, line 10: Actually, this part was wrong before your changes and after that. Just because what you're requesting to change is the `8.x-2.x` dev branch (sometimes called *default* or *main* or *master* or *HEAD*) of the Drupal Commerce. At the same time a **stable** release on the *Drupal* git repository it is not any particular branch. It is just some commit in the history having a tag, for example `8.x-2.1` (the latest currently). You'll not find this tag on the *Github* repository because it exists only on *Drupal* repository.

So, what we have: the users of the `HEAD` seeing the text which states that `8.x-2.x is in beta` which looks like a bit weird because **.x** versions can not be in `beta` or in `alpha` or `stable`. Furthermore, [this text is seen](http://cgit.drupalcode.org/commerce/tree/README.md?h=8.x-2.1#n8) by people using the latest `stable` version, not `beta`. The same with your text. You're stating that `Drupal Commerce 8.x-2.x is a stable release` and `It is usable for production websites`. Again, it is wrong.

I am not so fine in English but the right phrase should be found which will satisfy both users of the `dev` version and `stable` release of the *Drupal Commerce*.

drugan (changes requested):

README.md
```diff
@@ -1,13 +1,13 @@
+![Drupal Commerce](https://drupalcommerce.org/sites/default/files/dclogo_stacked_2c_on_white.png "Drupal Commerce logo")
```
drugan, line 1: I'd recommend to download this image into the [icons](https://github.com/drupalcommerce/commerce/tree/8.x-2.x/icons) folder. So, the markdown will be the following:

\!\[Drupal Commerce\]\(icons/dclogo_stacked_2c_on_white.png "Drupal Commerce logo")

You may see the example how it's done on the [Commerce Extended Attributes](https://github.com/drugan/commerce_xattributes) module's [README.md](https://raw.githubusercontent.com/drugan/commerce_xattributes/8.x-1.x/README.md) file.

**UPDATE:** may be even better would be to download the image into the `commerce` root folder and rename it to `logo.png`, so the final markdown will be that:

\!\[Drupal Commerce\]\(logo.png "Drupal Commerce logo")

drugan (changes requested):

bojanz:
Let's skip downloading the logo for now. I am not sure how it meshes with GPL licensing.

I've gone ahead and committed the version that just removes the offending part: 7d71e14aff254f38b134bd3da1672d1044a1d00f

Thanks, everyone!

PR #837, opened 2017-11-24
2925946 - Pass entity type ID to buildFieldDefinitions()

bradjones1:
https://www.drupal.org/project/commerce/issues/2925946

### DIFF
--- a/src/EntityTraitManager.php
+++ b/src/EntityTraitManager.php
@@ -95,7 +95,7 @@ class EntityTraitManager extends DefaultPluginManager implements EntityTraitMana
   public function installTrait(EntityTraitInterface $trait, $entity_type_id, $bundle) {
     // The fields provided by an entity trait are maintained as locked
     // configurable fields, for simplicity.
-    foreach ($trait->buildFieldDefinitions() as $field_name => $field_definition) {
+    foreach ($trait->buildFieldDefinitions($entity_type_id) as $field_name => $field_definition) {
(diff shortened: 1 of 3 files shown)


bradjones1:
@bojanz Not sure exactly how to handle the change to the interface... adding the parameter to the existing interface would break BC for existing traits. Or is this something that would be eligible for a breaking change? Not too excited about the new interface but it seems like it's the only way to both document this and maintain compatibility.

bradjones1:
To answer what is probably a logical question, the use case here is a trait that is the same for both products and product variations, but for the target entity types/bundles for an entity reference field.

bradjones1:
Argh... actually this is totally the wrong approach. Any thoughts?

joachim-n:
I would add a property to the trait plugin base class, and then either set that in the constructor or add a setter.

IIRC, backwards compatibility states that methods may get added to interfaces, and that base classes will add them too. So if you inherit from base classes, your code is fine.

bojanz:
I think we can sidestep this pain by either introducing a setEntityTypeId() and having installTrait() call it, or extending EntityTraitBase to look for an _entity_type in the configuration (which is the pattern we use in TaxTypeBase, PaymentGatewayBase, others) then modifying CommerceBundleEntityBase to pass the entity type ID to createInstance()

bradjones1:
After talking to Bojan in Slack, it looks like this is too edge-case-y and can easily be done on the implementing side without having to change anything in Commerce.

PR #838, opened 2017-11-24
Allow entity bundle traits to specify display and form mode settings

bradjones1:


bradjones1:
For my own reference: a version of this that will apply after #828 

https://gist.github.com/bradjones1/04ec2e3b171640af4078b9071bdfcb00

bradjones1:
I think the code standards error is incorrect...?

PR #841, opened 2017-12-05
2926563: Label validation and test update

haringsrob:
https://www.drupal.org/project/commerce/issues/2926563

I think this is somewhat in the right direction. Let me know if any modifications are required.


bojanz:
https://www.drupal.org/project/commerce/issues/2926563 has a newer patch

PR #842, opened 2017-12-11
#2910190: Price getters should not be NULL

haringsrob:


bojanz:
The current code is correct. We documented that NULL is a possible value, and it is.

Do a git blame, a previous version of the code did what you're suggesting (falling back to an object with the store's default currency), but that caused problems in multilingual environments and had to be removed, hence the current assumption that we don't know the order's currency until it has 1 complete order item.

PR #844, opened 2017-12-13
Return NULL if referenced entity is not found

xuxucode:
Methods such as `ProductVariation::getProduct`, `OrderItem::getPurchasedEntity` expect a NULL value if the referenced entity is not found.


bojanz:
Let's please have matching issues in the future. Opened one: https://www.drupal.org/project/commerce/issues/2993347

bojanz:
Committed in 2f47afe5481a676bf15bb958a6114bacac1ccd1e.

PR #845, opened 2017-12-14
Issue #2927833, #2874149, #2930722: VAT Number, EU VAT Validation + exemption, EU Event tax handling

archnode:
This is an updated pull request which includes:
- #2874149: Tax number field type + basic validation.
- #2930722: Tax number validation and validation for tax number in the EU
- #2927833: EU event tax handling and VAT exemption

### DIFF
--- a/modules/tax/commerce_tax.install
+++ b/modules/tax/commerce_tax.install
@@ -1,0 +1,13 @@
+<?php
+
+/**
+ * @file
+ * Install, update and uninstall functions for the commerce_tax module.
+ */
+
+/**
+ * Ensure that the post_update function is triggered.
+ */
+function commerce_tax_update_8201() {
+  // Do nothing.
+}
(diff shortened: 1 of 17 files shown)


bojanz:
Keep in mind that merging multiple issues into one drastically increases the time it will take to review them. This will only postpone getting those changes into Commerce.

archnode:
@bojanz Thank you for the reply! I created a single pull request as #2927833 depends on #2930722 which in turn depends on #2874149, which is not ideal, but not really feasible otherwise (afaik).

I prepared individual Issues and patches on d.o. which only contain the relevant portions so that its possible to review starting with #2874149, but I can also provide individual pull requests, as the issues correspond with branches in my fork anyway. Is there any preferred way, or something else I can do to speed up the process?

bojanz:
This PR is now outdated, work continued in the issues.

PR #846, opened 2017-12-28
Add registration fields to stores for commerce tax zones.

NickDickinsonWilde:
## Problem:
The multiple tax zones in the Canadian taxes (likely other areas), may not all be applicable since there are varying levels of registration requirements (sales volumes) and you may (for example) be registered to collect GST (also automatically HST), but not BC PST or MB PST or QST etc. However, it is currently all or nothing as to zones.

## Proposed Solution:
* Create registration fields per tax zone/type - this has an added benefit of being an easy way to provide data for receipts/invoices for jurisdictions that require tax numbers being present on invoices/receipts
* If present in tax zone definition, check for value on store as an extra registration confirmation.

To test, entity updates must be applied (drush entup). Only Canadian zones updated to use this - other zones will continue to work as previously.

## Todo:
Testing, documentation, update function(?).

### DIFF
--- a/modules/tax/commerce_tax.module
+++ b/modules/tax/commerce_tax.module
@@ -40,6 +40,11 @@ function commerce_tax_entity_base_field_info(EntityTypeInterface $entity_type) {
       ->setDisplayConfigurable('view', TRUE)
       ->setDisplayConfigurable('form', TRUE);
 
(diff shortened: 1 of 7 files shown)


NickDickinsonWilde:
Tests failing is mostly whatever here - Mostly about feedback on the proposed solution right now.

PR #847, opened 2018-01-17
#2856583 Allow free orders

mglaman:


bojanz:
This was fixed in https://www.drupal.org/project/commerce/issues/2856583

PR #849, opened 2018-02-13
Issue #2892207 by drugan: Increase the order item quantity scale

drugan:
See: https://www.drupal.org/project/commerce/issues/2892207


bojanz:
There is a newer patch on the issue. Time to phase out pull requests.

PR #855, opened 2018-06-01
2804227 get order balance jack

sheanhoxie:
#821 continuing on with this original issue from d.o. 2804227


bojanz:
Please post your changes to https://www.drupal.org/node/2804227 if they aren't contained in the latest patch. Thanks!

PR #856, opened 2018-06-21
2980695 - i18n - Missing t() wrapper for string "Please Select

idiazroncero:
Quick fix for a small issue on i18n, missing t() wrapper for a label element. See also:
https://www.drupal.org/project/commerce/issues/2980695


bojanz:
Continued on the issue.

PR #859, opened 2018-08-15
Add 'checkout' route

aaronbauman:
Feature request:
add /checkout route as a convenient, paramater-free way to redirect users to appropriate step of checkout

Use case:
I want to send a user to checkout, but don't know their cart id

### DIFF
--- a/modules/checkout/commerce_checkout.routing.yml
+++ b/modules/checkout/commerce_checkout.routing.yml
@@ -10,3 +10,12 @@ commerce_checkout.form:
     parameters:
       commerce_order:
         type: entity:commerce_order
+
+commerce_checkout.checkout:
+  path: '/checkout'
+  defaults:
+    _controller: '\Drupal\commerce_checkout\Controller\CheckoutController::checkoutRedirect'
+  requirements:
+    _access: 'TRUE'
+  options:
+    no_cache: TRUE
--- a/modules/checkout/src/Controller/CheckoutController.php
+++ b/modules/checkout/src/Controller/CheckoutController.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\commerce_checkout\Controller;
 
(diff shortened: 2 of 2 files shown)


bojanz:
There is no guarantee that the user will have a single cart. There might be a cart per order type, per store, or simply two carts due to the user logging in mid-checkout.

aaronbauman:
Yes, and in that case, this will redirect them to the first one.
Or as an administrator of a multi-cart operation, don't use this link.

bojanz:
That's the thing, there is no single-cart and multi-cart operation. Commerce is always in multi-cart mode. Without the ability to restrict that, having such a route inside Commerce itself doesn't make sense.

andyg5000:
People want what they want, especially when the 1.x version does this and they're used to it. I had the same need and wrote a custom route to handle it. 

My fear was that core might add a route in the future that would conflict with mine, but clearly that won't happen ;) 

I'll write a blog post on this tomorrow and we can point others to it for this and similar customization's

aaronbauman:
Thanks Andy - I'm working on my first sites with Commerce 8.x, so I'm coming from the ubercart/commerce7 world.
The sites I'm building do not have need for multiple stores or multiple carts.
I could be wrong, but I don't think that's an edge case.

bojanz:
Okay, let's move this proposal into an issue (with a patch), and keep it open for feedback.

(We've pretty much transitioned away from PRs as the main source of proposed changes).

aaronbauman:
https://www.drupal.org/project/commerce/issues/2994356

andyg5000:
I agree with @aaronbauman. It's great that the default is to support multiple stores. However, when a customer ends up with 2+ active carts for the same store (when on /cart), it's going to cause some confusion for the customer. An option to handle this and the redirect of /cart could be part of the bigger proposal. See y'all on d.o !

PR #862, opened 2018-10-08
Issue #2990765: Limit available currencies in payment gateway forms.

zaporylie:
https://www.drupal.org/project/commerce/issues/2990765


bojanz:
If we have a patch, we don't need a PR :) DrupalCI is faster than Travis these days, so I've even started prefering patches

zaporylie:
Duly noted.

PR #863, opened 2018-10-24
Running tests against php7.2 throws errors attempting to count non-countable objects.

mdlutz24:
Testing the obvious fixes here. The tests appear to be trying to test that EXACTLY ONE thing exists, and while they passed prior to 7.2, they are actually testing the AT LEAST ONE thing exists, because find only returns the first found item, and count()ing a non-countable non-null object always returned 1.  Now testing a non-countable throws an error, so I switched the finds to findAll. The tests now test that EXACTLY ONE thing was found, so we will see if they need any updates due to the change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ sudo: false
 dist: precise
 
 php:
+  - 7.2
   - 7.1
   - 7.0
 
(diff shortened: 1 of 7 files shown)


mdlutz24:
The only test that needed adjustment was the addstore test.  It should expect to find 2 stores, the one created in CommerceBrowserTestBase::setUp() and the one created by the test.

bojanz:
Committed in d1738ab1fc81fc769353f529a7407c9dd617b34f.

PR #865, opened 2018-11-21
Fix php notice "Undefined index: title"

zviryatko:
Drupal's core expecting to have `title` property in hook requirements, see https://github.com/drupal/drupal/blob/8.6.3/core/modules/system/src/SystemManager.php#L116, currently, there's PHP notice in the admin pages.


bojanz:
Please open an issue on drupal.org so that we can give you the appropriate credit

zviryatko:
Here it is: https://www.drupal.org/project/commerce/issues/3015975

bojanz:
Continued on the issue.

PR #870, opened 2019-01-08
Product variation titles shall be updated when product title changes

geek-merlin:
[Product variation titles shall be updated when product title changes \[\#2986678\] \| Drupal\.org](https://www.drupal.org/project/commerce/issues/2986678#comment-12917845)


bojanz:
We don't pair PRs and issues anymore, there's no need since Drupal's testbot works fine nowadays.

So let's continue on https://www.drupal.org/project/commerce/issues/2986678

PR #872, opened 2019-07-17
remove dead code in EndDateWidget

lalop:
simply remove a useless set


bojanz:
Thanks! Please open an issue on drupal.org in the future so that you can get properly credited. We are phasing out PRs.

PR #876, opened 2019-10-21
3089178 make the build pass against 88 without deprecation suppression

mdlutz24:
Wanted to throw this against travis


mdlutz24:
Fixed in https://github.com/drupalcommerce/commerce/commit/b65dcc0114bd061d5a1a8842a4c055d123b41ee5

